PR #3, opened 2016-05-19
System requirement doc page created

ThomasRyd:
Michael, can you look at the text and formatting (should we have table borders)? Also, not sure why image is not showing


michaelatmender:
The image isn't showing because it's inside html rather than markdown, so the relative path isn't working. It would have to be written like: /user/pages/01.Getting-started/02.System-requirements/mender_client_partition_layout.png

However I would say a table is unnecessary here, it doesn't add anything. Best to keep to markdown only. Likewise with the links: use [link text]( link-url?target=_blank ) rather than <a> tags.

PR #10, opened 2016-06-24
MEN-341: Document uboot integration

kacf:


kacf:
@michaelatmender: Maybe just check that I got the structure right.

PR #15, opened 2016-07-13
First pass at defining the mender.conf file

csimmonds:
This is bare-bones stuff. I would like comments that fill in some of the question marks. For example
1. What is the significance of the DeviceID? What is the format? How is it generated? Is it unique for each device? If so, how is it programmed into each device?
2. What is the ServerCertificate for? How is it generated? How do I set it in the YP build?
3. How should I change values from defaults in my YP build. For example, if I want PollIntervalSeconds to be 3600, what is the best way to do that?

TIA, Chris


eysteins:
@kacf perhaps you could help clarify?

kacf:
This file is still a bit up in the air, I'm not completely certain what's the best way to fill it with data, since it takes information from many sources. A couple of points though:

* `DeviceID` is going away completely, you can remove it (even from meta-mender, if you want).
* Poll interval is not configurable with a YP variable right now.
* Neither is http/https.
* The server certificate location is fuzzy right now. I think the intention is to generate offline, specify a location of that certificate on the build host (so the path will be on the build host, not the device), and then this certificate will be put on the device under `/etc/mender`. But I'm uncertain how well this works right now, you'll have to try it out a bit.

csimmonds:
It looks like mender.conf is not really useful to the system integrator right now, so I am dropping this PR

PR #17, opened 2016-07-20
MEN-547: Write verification checklist for correct U-Boot integration.

kacf:

### DIFF
--- a/04.Devices/04.Configuration-checklist/docs.md
+++ b/04.Devices/04.Configuration-checklist/docs.md
@@ -1,0 +1,147 @@
+---
+title: Configuration checklist
+taxonomy:
+    category: docs
+---
+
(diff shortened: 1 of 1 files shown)


kacf:
@csimmonds: Perhaps you could verify that this looks sane to your eyes?

I plan to follow up with a PR with a few entries of the type "What if step X fails?"

csimmonds:
@kacf: It's quite an intricate series of steps. I feel that it needs a bit more context at the beginning ("we are going to create two different images and verify that mender can update from one to the other"...) and to explain what each step is for *before* giving the sequence of commands

kacf:
Good idea, I'll add that.

kacf:
@csimmonds: I have updated with more contextual info, check the last commit!

PR #18, opened 2016-07-22
MEN-523: Add section about disabling Mender as a system service.

kacf:


kacf:
@csimmonds: Feel free to change the location of the text, "Customizations" may not be the best section name...

PR #21, opened 2016-08-12
Change u-boot-fw-utils instructions after change in meta-mender.

kacf:


kacf:
Depends on https://github.com/mendersoftware/meta-mender/pull/115.

PR #41, opened 2016-10-26
Fix invalid formatting

lolsborn:
Inline comments in config is invalid and won't build.  Use double-quotes for consistency.


eysteins:
Thanks, yes I should have checked better. :)

PR #48, opened 2016-10-28
Troubleshooting updates

kacf:


kacf:
@eysteins: Merge if you agree (particularly with the removal of `UBOOT_SUFFIX` and moving of `CONFIG_FHANDLE`).

eysteins:
@kacf nice!

I would keep the Kernel support in system requirements (shorten it if you want), because people ask "what are the kernel requirements for Mender" and if we don't document that (even though there are none or only for systemd) people will keep wondering and asking. :)

kacf:
Thanks, I omitted that commit!

PR #54, opened 2016-11-15
Follow splitting of meta-mender into several layers.

kacf:


kacf:
@eysteins: If you can think of other places this needs to be mentioned, let me know.

I'll send a heads up to the mailing list.

eysteins:
Looks good, I can't think of other places to update!
The only thing is that we use "device" rather than "board" unless it is in an established setting like "board support package". I have updated that in a follow-up commit.

PR #56, opened 2016-11-18
Update all updates to refer to `.mender` instead of `.ext4`.

kacf:
The latter doesn't work anymore after we changed to the mender
artifact format.

### DIFF
--- a/01.Getting-started/05.Standalone-deployments/docs.md
+++ b/01.Getting-started/05.Standalone-deployments/docs.md
@@ -32,13 +32,12 @@ suffix, so they are easy to recognize. This file contains
 all the partitions of the given storage device, as
 described in [Partition layout](../../Devices/Partition-layout).
 In addition, you need a rootfs image to update to. The suffix
-of this file depends on the file system used for rootfs,
-for example `.ext4`.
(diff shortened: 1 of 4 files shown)


eysteins:
This one is a bit tricky as we document using the `stable` branch. So I think stable must be fast-forwarded to support .mender simultaneously with this getting merged. Perhaps @pasinskim has a grand plan.

kacf:
We should create a stable doc branch if we need that. There is already a ton of changes in master that are not relevant for stable, such as the breaking up into meta sub layers. I think @GregorioDiStefano has been holding back the updates from master to the actual docs site (correct me if I'm wrong).

GregorioDiStefano:
Yes, we need some sort of drop down, in order to pick which documentation we want to view

pasinskim:
Yup. I agree. We should probably have a stable documentation branch which
should be in sync with Mender stable. As it is right now it will be tricky
to move stable "tags" in code as we will end up with outdated
documentation.

kacf:
Sounds good. In the meantime, I'll merge this if no one has any objections before tomorrow.

PR #57, opened 2016-11-18
MEN-763: Document new artifact and device type related variables.

kacf:
* MENDER_ARTIFACT_NAME
* MENDER_DEVICE_TYPE
* MENDER_DEVICE_TYPES_COMPATIBLE

### DIFF
--- a/03.Artifacts/01.Building-Mender-Yocto-image/docs.md
+++ b/03.Artifacts/01.Building-Mender-Yocto-image/docs.md
@@ -120,14 +120,21 @@ part of your Yocto Project build environment.
 Add these lines to the start of your `conf/local.conf`:
 
 ```
+MENDER_ARTIFACT_NAME = "my-mender-image-1.0"
+
 INHERIT += "mender-full"
+
 MACHINE = "<YOUR-MACHINE>"
+
 DISTRO_FEATURES_append = " systemd"
 VIRTUAL-RUNTIME_init_manager = "systemd"
 DISTRO_FEATURES_BACKFILL_CONSIDERED = "sysvinit"
 VIRTUAL-RUNTIME_initscripts = ""
+
 IMAGE_FSTYPES = "ext4"
 ```
+
+`MENDER_ARTIFACT_NAME` is name of the image or update that will be built. This is what the device will report that it is running, and different updates must have different names because Mender will skip installation of an artifact if it is already installed.
 
 Please replace `<YOUR-MACHINE>` with the correct machine for your device.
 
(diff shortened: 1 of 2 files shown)


kacf:
Don't merge yet, the meta-mender part is not completely done. Review appreciated though.

03.Artifacts/01.Building-Mender-Yocto-image/docs.md
```diff
@@ -120,15 +120,22 @@ part of your Yocto Project build environment.
...
+
 DISTRO_FEATURES_append = " systemd"
 VIRTUAL-RUNTIME_init_manager = "systemd"
 DISTRO_FEATURES_BACKFILL_CONSIDERED = "sysvinit"
 VIRTUAL-RUNTIME_initscripts = ""
+
 IMAGE_FSTYPES = "ext4"
 ```
 
+`MENDER_ARTIFACT_NAME` is name of the image or update that will be built. This is what the device will report that it is running, and different updates must have different names.
```
eysteins: Maybe add "...different updates must have different names because Mender will skip installation of an artifact if it is already installed."

03.Artifacts/06.Variables/docs.md
```diff
@@ -17,49 +17,73 @@ Influences which file system type Mender will build for the rootfs partitions in
...
 
-The partition Mender uses as the first (A) rootfs partition. See [More detailed storage configuration](../../Devices/Partition-layout#more-detailed-storage-configuration) for more information.
+The size of the persistent data partition in the generated `.sdimg` file. See [Configuring the partition sizes](../../Devices/Partition-layout#configuring-the-partition-sizes) for more information.
 
 
-#### MENDER_ROOTFS_PART_B
+#### MENDER_DEVICE_TYPE
 
-The partition Mender uses as the second (B) rootfs partition. See [More detailed storage configuration](../../Devices/Partition-layout#more-detailed-storage-configuration) for more information.
+A string that defines the type of device this image will be installed on. This variable is only relevant when building a complete partitioned image (`.sdimg` suffix). Once a device is flashed with this, it will not change, even if the device is updated (though it can be changed manually by editing `/data/mender/device_type` on the device).
```
eysteins: Very nitpicky and not sure about this, but shouldn't it be in /var/lib/mender which is symlinked to /data/mender?

eysteins:
:+1:

eysteins:
Not directly related, but do you think the Artifact format reference should be moved to the docs as well: https://github.com/mendersoftware/artifacts/blob/master/Documentation/artifact-format.md ?

I think we should at least explain at a high level what this .mender file is somewhere, probably under "Architecture". I can think a bit about it.

03.Artifacts/06.Variables/docs.md
kacf: Yes, I was unsure about the last part as well. I'll remove it, since you also have doubts. :-)

And yes, you're right about `/var/lib/mender`, I need to remember that next time.

03.Artifacts/01.Building-Mender-Yocto-image/docs.md
kacf: Will fix!

kacf:
> Not directly related, but do you think the Artifact format reference should be moved to the docs as well: https://github.com/mendersoftware/artifacts/blob/master/Documentation/artifact-format.md ?

Hmm, tricky... I'm not sure. I think I'm leaning towards it being too low level. But we should definitely reference it in the docs, because separately developed clients will need it.

PR #61, opened 2016-12-02
Master docs emergency fixes

eysteins:
Temporarily revert a few commits that refer to master code.
The real problem is that we publish docs from the master branch, however revert these commits until we publish stable branch.


eysteins:
Revert of these commits in PR at https://github.com/mendersoftware/mender-docs/pull/62

PR #64, opened 2016-12-07
SYSTEMD_AUTO_ENABLE is a package-local variable, don't set it globally.

kacf:
Not only will you trigger a rebuild of a lot of packages by changing
it, but you will influence a lot of packages you didn't intend.


kacf:
@bboozzoo: I haven't tested this yet, but it makes sense, right?

eysteins:
Thanks for catching this; I just noticed the Mender daemon did not start with our default config, as it used to (and should).

PR #80, opened 2017-01-03
move APIs and Throubleshooting structure to make room for new section

estenberg:
Line 184 & 199 in deploy_docs script needs update (from "06.APIs" to "07.APIs") together with this merge.
@Enemico can you pls make that change in deploy_docs, update the infra and merge this together with it?


Enemico:
The merge requires changes on deploy_docs: 

https://github.com/mendersoftware/infra/blob/master/files/mender/deploy/deploy_docs.sh#L184
https://github.com/mendersoftware/infra/blob/master/files/mender/deploy/deploy_docs.sh#L199

And is changing 16 files in mender-docs. 
Since we are automatically deploying in production every 3 hours during the day, this requires testing. 
I will need to find a way to deploy the changes to my test host ( that now is being used to work on https://tracker.mender.io/browse/INF-399 ) locally, in order to verify that this changes are actually not breaking the deployment. 

Alternatively, i can just merge this and make the changes on deploy_docs and hope for the better. 
You still prefer cholera?  :-)

michaelatmender:
Suggestion: we give APIs and Troubleshooting directories higher numbers, so we leave some breathing space for adding more categories later. Grav doesn't care if the numbers are consecutive - you could make 200.API and 201.Troubleshooting, and they will be the last 2 on the menu

estenberg:
@michaelatmender good idea, I updated it now. I also tested it works (no int overflow :).

@Enemico yes, still a cholera-man. :) I tested this locally, and the worst thing that can happen is that the APIs/ section will be empty, so I think it's ok to just merge it, change deploy scripts, deploy to staging. Then we can revert the commit immediately (git revert commit-id) if not good.

Enemico:
Ok, i will do it now, since production won't be affected until tomorrow
morning.

PR #88, opened 2017-01-09
administration: notes on administration & deployment

bboozzoo:
@maciejmrowiec @mendersoftware/rndity @eysteins

### DIFF
--- a/06.Administration/01.Overview/docs.md
+++ b/06.Administration/01.Overview/docs.md
@@ -1,0 +1,71 @@
+---
+title: Overview
+taxonomy:
+    category: docs
+---
+
+Mender backend is composed of a number of microservices, each implementing a
+small, well defined piece of functionality.
+
+The integration environment, previously used
+in [Create a test environment](../../Getting-started/Create-a-test-environment)
+chapter, brings together the following services:
+
+- [Mender Device Admission Service](https://github.com/mendersoftware/deviceadm)
+- [Mender Device Authentication Service](https://github.com/mendersoftware/deviceauth)
+- [Mender Deployment Service](https://github.com/mendersoftware/deployments)
+- [Mender Device Inventory Service](https://github.com/mendersoftware/inventory)
+- [Mender User Administration Service](https://github.com/mendersoftware/useradm)
+- [Mender API Gateway](https://github.com/mendersoftware/mender-api-gateway-docker)
+- [Minio](https://www.minio.io/) object storage
+- Storage service proxy based on [OpenResty](https://openresty.org/en/)
+
+Services are delivered in form of Docker images, available from
+official [Mender Docker repository](https://hub.docker.com/r/mendersoftware/).
+When required, each service can be built directly from its source code. Consult
+respective repositories for build instructions.
+
+While it is possible to `docker pull` individual images and start containers
+manually,
+the [integration repository](https://github.com/mendersoftware/integration)
+provides a convenient setup based on Docker Compose tool. The
+`docker-compose.yml` file describes the following setup:
+
+```
+        |
+        |                                            +-------------------------+
+        |                                            |                         |
+        |                                       +--->|  Device Authentication  |
+        |                                       |    |  (mender-device-auth)   |
+        |                                       |    +-------------------------+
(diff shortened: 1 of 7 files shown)


maciejmrowiec:
This differs a little from the docs architecture plan: https://docs.google.com/document/d/1cFFiusOP8aErgcjnPgJG_0W4w2G0y7dIoZEDqBRl0nw/edit

However I find this split more intuitive than the one from google doc.

bboozzoo:
There's a plan ;) ok, i'll check it out and update the PR.

maciejmrowiec:
I favour your structure. @eysteins perhaps we can adjust the google doc plan?

estenberg:
A few minor notes:

* Administration/Overview section is really good! Could have been merged immediately if it was a separate commit.

* Do you use https://github.com/mendersoftware/mender-docs-site? Several places have formatting issues with warning notes (e.g. "Make sure that the certificate matches the domain name..." is split so half of sentence only is warning-formatted).

estenberg:
Regarding the main task (currently Administration/Deployments), it is currently too "bottom up" for the use case at hand (generate and use secure certificates everywhere). I do need to generate certs for all the services (which is in separate locations). I don't quite understand the current split. Do we need 3 different certificates (API Gateway, Minio, Token signing)? How do I generate them (only noted in Token signing)? Where do I put them on the machine that runs docker-compose (can be inferred by looking closely at the mount points)? Should I change the perms on them to make them strict? How do I make the client use the certs (it won't work otherwise)?

What I suggest for now:
* Remove Administration/Deployments for now, instead:
* Merge all the certificate generation and include everything (see notes above) into a tutorial-style separate section Administration/Certificates.
* Split out storage into it's own section Administration/Storage (currently two headers "Storage location" and "S3 storage backend"). Also include any storage limits here (max size per artifact? max artifacts?).
* Split out "Bandwidth and connection limiting" header into its own section Administration/Bandwidth.

I've updated the plan to include Storage as well, otherwise please try to follow that going forward to make sure everything is consistent: https://docs.google.com/a/cfengine.com/document/d/1cFFiusOP8aErgcjnPgJG_0W4w2G0y7dIoZEDqBRl0nw/edit?usp=sharing

bboozzoo:
Ok, i"ll update the PR to be in line with documentation plan.

I'm not convinced about Administration/Bandwidth though. It gives the impression that this applies to the whole system while it's only storage download/connection limiting. We have no means of policing device/management requests.

bboozzoo:
With current documentation plan, where do I put generating keys for token signing?

bboozzoo:
I've updated the PR to be mostly in line with the documentation plan. The only divergence is 'Token sinign section', I did not have any idea where to put it, so it's become a seprate chapter.

estenberg:
Thanks, I merged it and polished it a bit now.
It's available at https://docs.mender.io/Administration

I will continue to polish it - also see https://github.com/mendersoftware/integration/pull/113 and https://github.com/mendersoftware/mender-docs/pull/89.

estenberg:
By the way, how come the deviceauth and useradm only have private keys? How are these shared between other services for verification? Or are the resulting signatures only verified by the services themselves?

bboozzoo:
deviceauth and useradm need private keys to sign the tokens. Verification happens at the gateway, before forwarding the request to the upstream service nginx does an `auth_request` to either deviceauth or useradm (depending on the endpoint being accessed). Basically, the whole incoming request (sans the data obviously) gets forwarded to auth endpoint. Original headers are preserved (Authorization which carries token in our case) and 2 extra are added - X-Original-URI and X-Original-Method. When the request hits the upstream service eventually, it's already after auth verification.

estenberg:
OK, thanks, I will include this in the docs.
So we could have used a MAC, such as HMAC-SHA256 instead of public key crypto for these services as well? Not that important now, but this will probably be quite a lot more efficient.

bboozzoo:
Yes, we could have used HMAC but that requires you to use a shared secret. So for instance if a 3rd party wants to verify if token was issued by mender (not that there's a valid case for this atm), we'd need to have a public token verification service (we could not share the secret obviously). With RSA, we can just give them the public key.

PR #89, opened 2017-01-10
Add separate section about production builds.

kacf:
And document properly how to insert certificate in client build.


kacf:
@eysteins: I don't know if you want to go through all of it, but do you agree with the structure? The information seemed to be lacking about the difference between production/demo builds. The way I do it here, the production part is the secondary goal, with testing/demoing being the primary one; I did this primarily thinking about ease of use / shortest time to value, but it can be debated whether this is the right order to introduce things.

estenberg:
Looks great! Agree on the easy demo point.
I just need to sync with https://github.com/mendersoftware/mender-docs/pull/88 and will merge shortly.

estenberg:
Thanks, I added a commit to polish it a bit here: https://github.com/mendersoftware/mender-docs/commit/ac686ae90744551453d7bc63ea45e94e51e77131

Two minor notes:
* When linking to external pages (outside docs.mender.io), please open URL in new tab/window using https:/...?target=_blank
* Use the non-abbreviated forms like "do not" instead of "don't" as in these more "formal" settings :)

kacf:
Noted!

PR #91, opened 2017-01-18
Explain how to generate the server.crt client cert from server certs.

kacf:


estenberg:
Nice!

PR #96, opened 2017-01-30
MEN-932: Document that we use standard https port now, not 8080.

kacf:
Since this is the default, some sections can be removed, since no
change is required anymore.


estenberg:
Nice, simplified. :)

The tricky thing here is the master/stable out-of-sync issue again. So I think we have to wait with merging..

PR #97, opened 2017-01-31
MEN-932: Document how to change server and port using Yocto.

kacf:


estenberg:
Nice, thanks!

PR #98, opened 2017-01-31
MEN-932: Rename section to "Image configuration" to be more generic.

kacf:
A subtle wording change, but the clue is that the section should cover
any and all client configuration meant to be done during the build,
even if some changes are not really a build system feature.

The only exception for this section is configuration for production
devices, which is still covered in its own section.


estenberg:
Makes sense, but please make sure to update content that links to the old section.

kacf:
Updated!

estenberg:
Thanks!

PR #100, opened 2017-02-02
apis/integration: make room for useradm docs

mchalski:
we'll automatically generate these docs just like all other.
goes with https://github.com/mendersoftware/useradm/pull/32
Issues: MEN-976

Signed-off-by: mchalczynski <marcin.chalczynski@rndity.com>


estenberg:
Makes sense, feel free to merge once the scripting integration is completed.

PR #117, opened 2017-02-14
administration: production: update URL for checking if mender backend is up

bboozzoo:
URL API versioning scheme was updated at the gateway, but the docs were left
behind using the older scheme.

Fixes: [Mender #1009]

@maciejmrowiec @mendersoftware/rndity @pasinskim


maciejmrowiec:
needs to get into 1.0 branch

PR #127, opened 2017-03-17
Some more UBI related updates

mirzak:


mirzak:
@kacf Any comments on this?

kacf:
@mirzak: Btw, we were discussing in the office how to prioritize UBI recently. It's not at the top of our list currently, but given the work that has already gone into it, we don't want to leave it for a long time either. It would be useful to know your future plans for this feature.

mirzak:
> @mirzak: Btw, we were discussing in the office how to prioritize UBI recently. It's not at the top of our list currently, but given the work that has already gone into it, we don't want to leave it for a long time either. It would be useful to know your future plans for this feature.

My plan is to continue the work and as it seems at the moment I will able to continue during the coming weeks, starting with this. I am currently investigating how this should be implemented in the Mender client and how found some "issues" that need to be resolved/discussed. I will try and summarize this to get your opinion

I will try and create a "proper" plan on the Mender tracker as well and then it would probably be easier for you to track and prioritize in the future.

kacf:
Cool! You can use this ticket to post plans/progress: https://tracker.mender.io/browse/MEN-1126

kacf:
@estenberg: UBI support in both the client and the meta-mender layer has been merged now. We're not going to document all of this now (this is covered in [this ticket](https://tracker.mender.io/browse/MEN-1130)), but I think it deserves a mention as an experimental feature. Where do you think the best place is? Under "3. Device Integration -> Partition Layout -> Flash Memory types" maybe?

estenberg:
@kacf Awesome! Yes, that would be a good place in the docs for it.

kacf:
Alright, added here: #163.

PR #140, opened 2017-03-31
Add documentation for boot loader variables.

kacf:


kacf:
Cherry-picked to 1.0.x as well.

PR #147, opened 2017-04-18
Document MENDER_ARTIFACT_SIGNING_KEY variable.

kacf:


estenberg:
Would note that the artifact can be signed any time after it is built (link to how?), which is best practice for more security (in cases where the build server is compromised).

kacf:
@estenberg: Good point. I added fbda15fccc47fc7e6b3b83fdebf5d169a9a8a960 as an amendment. Note that the referenced section doesn't actually contain any information about signing yet, but I expect this to be written as part of completing the epic.

kacf:
Ah yes, there is https://tracker.mender.io/browse/MEN-1024 (or MEN-2^10 if you prefer.... *crickets*)

PR #150, opened 2017-04-25
Document Artifact signing and verifying.

pasinskim:
Signed-off-by: Marcin Pasinski <marcin.pasinski@mender.io>


04.Artifacts/06.Building-for-production/docs.md
```diff
@@ -61,3 +61,10 @@ SRC_URI_append_pn-mender = " file://server.crt"
 ```
 
 Note in particular the `:` after the directory; this is mandatory.
+
+
+## Artifact signing and verification keys
+
+The private key used for signing the Mender Artifact should be protected and kept outside of the build system, thus there are no extra steps needed to add it to any part of the build system or either Client or the Server environment. Only the public key, which is used by the Mender Client to verify the signed Artifact must be included in the Client build. 
+The best way to include a public verification key in the client is to set the name of the verification key to `artifact-verify-key.pem` and place in `SRC_URI` before building the Yocto client image. For more information about some alternate approaches please follow [MENDER_ARTIFACT_VERIFY_KEY documentation](../Variables#mender-artifact-verify-key)).
```
estenberg: This section is really important - how to put the verification key on the client and enforce signature verification. From this it is not that clear to me, where do I find `SRC_URI`? Would have exact steps here.

estenberg:
Sorry for the late feedback. Overall very nice, good structure. I will merge and do some minor polishing. But pls see the comment above.

04.Artifacts/06.Building-for-production/docs.md
kacf: I propose https://github.com/mendersoftware/mender-docs/pull/153.

PR #161, opened 2017-06-01
MEN-1221: Fix incorrect ECDSA key generation arguments.

kacf:
Technically the first key generated is valid, but Go refuses to parse
one that has a parameter section, so get rid of that using the second
line.


drewmoseley:
Acked-by: Drew Moseley \<drew.moseley@mender.io\>

kacf:
Cherry-picked to 1.1.x as well.

PR #163, opened 2017-06-02
Mention UBI support as an experimental feature.

kacf:


kacf:
@mirzak: Anything else you think I should add? :-)

estenberg:
Good, and I would tone it a bit more positive, given that we are going to have support for it soon. :)

Maybe something like this:

Mender officially supports  *block devices* and there is *experimental* support for raw flash and UBI volumes and file systems. To enable UBI support, inherit the `mender-full-ubi` class in your `local.conf` and take a look at the various UBI related variables in `mender-install.bbclass`. Please keep in mind that this is an experimental feature so please provide feedback on the [mailing list](https://groups.google.com/a/lists.mender.io/forum?target=_blank#!forum/mender) if you have any questions or issues. Raw flash support will be officially supported once it has been thoroughly tested and it has proven high quality.

kacf:
Done and done!

PR #186, opened 2017-07-13
Manually revert previous change for testing purposes only.  Do not me…

drewmoseley:
…rge.


drewmoseley:
I pulled the trigger a bit quickly on the previous PR and there was an issue reported by Travis.  I'm using the PR to recheck but will _not_ merge this.

drewmoseley:
OK.  Looks like the previous test failure was a temporary network glitch with github connections.  This one succeeded so I think we are good.

PR #198, opened 2017-08-09
Automated spellchecking

drewmoseley:
This is a script and custom dictionary to allow for running ispell. It won't catch everything as I had to add a lot of strange things to the dictionary to cover crypto hashes, bitbake variables, etc but it is a bit easier than spell checking by hand.


drewmoseley:
@michaelatmender @GregorioDiStefano @estenberg what do you guys think of this?  I've got similar changes in the other branches I can submit if we like this approach.

GregorioDiStefano:
good idea, we should have used this earlier on!

PR #201, opened 2017-08-10
1.2.x spelling

drewmoseley:


estenberg:
Looks great to me, please feel free to merge.

We should pick to other branches too, especially master, right? In general the standard workflow is to fix master first then cherry-pick to stable branches so we are sure that we don't miss important fixes in master and issues resurface.

estenberg:
ah, I see you have 1.1.x in #202

drewmoseley:
Yeah, for the spelling changes I just did them as mostly separate commits rather than cherry picking.  It was just easier that way.

PR #202, opened 2017-08-10
1.1.x spelling

drewmoseley:
@estenberg are we still taking changes on this branch?


estenberg:
Yes, there will be 1.1.1 so we should fix this.

PR #212, opened 2017-09-01
Fix docs for PREFERRED_VERSION settings

drewmoseley:


drewmoseley:
@kacf does this look like what we discussed?

04.Artifacts/01.Building-Mender-Yocto-image/docs.md
```diff
@@ -128,9 +128,20 @@ INHERIT += "mender-full"
...
+#
+# Note the following mapping from Mender client version to Mender artifact version.
+# Client => Artifact
+# 1.0.0     1.0.0
+# 1.0.1     1.0.0
+# 1.1.0     2.0.0
+# 1.2.0     2.1.0
+PREFERRED_VERSION_pn-mender = "1.2.0"
+PREFERRED_VERSION_pn-mender-artifact = "2.1.0"
+PREFERRED_VERSION_pn-mender-artifact-native = "2.1.0"
```
kacf: I don't think we should use this in our default recommendation, it should rather just pick the latest, and this can be in a section describing how to get a specific version. Does that makes sense?

In any case it should be `1.2.%` to allow automatic patch upgrades.

drewmoseley: @kacf I'm not quite clear on your suggestion here.  Are you saying specifically for the master branch to not have values of PREFERRED_VERSION_blah set? Or should this also be done for other branches?

kacf: Mainly for master, but maybe also for the latest branch.

drewmoseley: Will the later named version be selected over the _git version automatically?

kacf: Yes.

PR #213, opened 2017-09-02
improved readability of state script docs

estenberg:
Most notably
* Split out use cases
* Simplified transitions and ordering pieces
* Split out most common state transitions (they were one very very long line)
* Uppercase Artifact and Mender Client


estenberg:
FYI @pasinskim. Should also be backported to 1.2.x once merged.

estenberg:
Thanks, I will merge now as we are on a release schedule here.

PR #259, opened 2017-09-14
Use wrapper scripts to create default user.

drewmoseley:


drewmoseley:
In the current git integration environment (git://github.com/mendersoftware/integration) as well as the archive tarball (https://github.com/mendersoftware/integration/archive/master.tar.gz) I am getting an error when creating the initial user as documented in the production installation section.

```
$ sudo docker-compose exec mender-useradm /usr/bin/useradm create-user --username=drew.moseley@northern.tech --password=mysecretpassword
ERROR: No container found for mender-useradm_1
```

If I wrap that in the provided script so that it uses the same docker-compose yaml files it works:
```
$ ./run exec -T mender-useradm /usr/bin/useradm create-user --username=drew.moseley@northern.tech --password=mysecretpassword
2b0915e3-6354-4ca7-84ff-7e17d0776a60
```

This change updates the docs for both the production installation and the demo installation so use the wrapper script.  I don't know why this only _just_ started happening or why it doesn't _seem_ to happen in the demo setup but this seems like the right change to make regardless.

Anybody have any thoughts?

PR #260, opened 2017-09-18
MEN-595: Document new MENDER_FEATURES_ENABLE|DISABLE variables.

kacf:


kacf:
Cherry-picked to 1.1.x and 1.2.x as well.

PR #277, opened 2017-10-11
Document that client certificate is only needed for self signed certs.

kacf:


drewmoseley:
Looks good to me.

PR #286, opened 2017-10-19
Clarify the need to add the beaglebone BSP layer.

drewmoseley:


drewmoseley:
@kacf does this make sense?

PR #295, opened 2017-10-26
Fixed typo

yene:


drewmoseley:
This looks correct however I suspect the multiplier should be adjusted to 4096 to account for the switch from MB to KB. @kacf does that sound right?

It also looks like this should be applied to the 1.2.x and 1.1.x branches.  @kacf can you confirm?

kacf:
> This looks correct however I suspect the multiplier should be adjusted to 4096 to account for the switch from MB to KB. @kacf does that sound right?

Yes, except that for clarity I would write `4 * 1024`, since 4 is the number of partitions, not some size, so as to not confuse them.

> It also looks like this should be applied to the 1.2.x and 1.1.x branches. @kacf can you confirm?

Agreed!

drewmoseley:
Corrected version available here: https://github.com/mendersoftware/mender-docs/pull/300

Closing

PR #297, opened 2017-10-30
Standalone deployments

estenberg:
Move from getting started to architecture.


drewmoseley:
@estenberg should this be backported?

estenberg:
@drewmoseley already done, nothing more to do here :)

PR #298, opened 2017-10-30
reorganize getting started to make it simpler

estenberg:


drewmoseley:
@estenberg is this worth backporting?

estenberg:
@drewmoseley yep, I will take care of it.

PR #303, opened 2017-11-03
Do not use sudo to run docker

drewmoseley:


drewmoseley:
I updated this to clarify the need to be in the docker group.  Once this is done I'll pull it to the older branches.

estenberg:
:+1:

PR #309, opened 2017-11-10
1.2.2 update

GregorioDiStefano:
@eysteins @estenberg

### DIFF
--- a/01.Getting-started/02.Create-a-test-environment/docs.md
+++ b/01.Getting-started/02.Create-a-test-environment/docs.md
@@ -17,7 +17,7 @@ environment:
 [start_autoupdate_integration_archive_x.x.x.tar.gz]: #
 
 ```bash
-curl -L https://d1b0l86ne08fsf.cloudfront.net/1.2.1/integration/integration-1.2.1.tar.gz | tar xz
+curl -L https://d1b0l86ne08fsf.cloudfront.net/integration/integration-1.2.2.tar.gz | tar xz
 ```
 
 [end_autoupdate_integration_archive_x.x.x.tar.gz]: #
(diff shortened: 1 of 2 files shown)


GregorioDiStefano:
oops, ignore!

04.Artifacts/01.Building-Mender-Yocto-image/docs.md
```diff
@@ -153,7 +153,7 @@ MACHINE = "<YOUR-MACHINE>"
 # To get your tenant token, log in to https://hosted.mender.io,
 # click your email at the top right and then "My organization".
 # Remember to remove the meta-mender-demo layer (if you have added it).
-# We recommend Mender 1.2.1 and Yocto Project's pyro or later for Hosted Mender.
+# We recommend Mender 1.2.2 and Yocto Project's pyro or later for Hosted Mender.
```
estenberg: Is there something specific in 1.2.2 that makes it better with Hosted? I knew it worked on 1.2.1 so that's why I added it, but I think it is ok as-is if it works well with 1.2.1 and later..

estenberg:
What about the links to the Artifacts and .sdimg files? E.g. those referenced here https://docs.mender.io/1.2/getting-started/deploy-to-virtual-devices

04.Artifacts/01.Building-Mender-Yocto-image/docs.md
GregorioDiStefano: Mender client 1.2.2 doesn't even exist, so this should be 1.2.1, I'll fix it.

GregorioDiStefano:
the 1.2.1 links are the same, because the client is still 1.2.1, and we are referencing the images for that client version.

estenberg:
Thanks!

PR #310, opened 2017-11-13
Artifacts/state-scripts: Added state-sript retry-later documentation

olepor:
Changelog: Title


kacf:
Thanks, tweaked it a little and resubmitted here: #313.

olepor:
Nice.

Looks good.

Also, I will not come into the office tomorrow, as I get a funeral to attend in the middle of the day at home, home (Hadeland).
So I will be available in the morning, and then later again in the evening. Funeral is at 13:00.

- O

PR #316, opened 2017-11-13
Specify remote name with git log command.

drewmoseley:


drewmoseley:
Note, this is only needed on the master branch.  The other branches use tags which don't exist per remote.

drewmoseley:
@GregorioDiStefano @kacf does this look right?

PR #317, opened 2017-11-13
Use fully qualified branch for git log command.

drewmoseley:

### DIFF
--- a/06.Administration/02.Production-installation/docs.md
+++ b/06.Administration/02.Production-installation/docs.md
@@ -534,7 +534,7 @@ At this point your commit history should look as follows:
 
 <!--AUTOMATION: ignore -->
 ```bash
-git log --oneline 1.2.x..HEAD
+git log --oneline 1.2.2..HEAD
 ```
 > 7a4de3c production: final configuration  
 > 41273f7 production: adding generated keys and certificates  


drewmoseley:
@GregorioDiStefano @kacf does this look right?

estenberg:
Note that this would need to be updated every release (like now we are about to release 1.2.2). If we want this then we definitely need some kind of automation here.

drewmoseley:
True but I'm more worried that we are inconsistent between the branches.  I can change this to use the branch name which would not require updating.  ie "git log origin/1.2.x..HEAD"

Is that a preferred change?

estenberg:
Makes sense to me. :)

drewmoseley:
OK.  I updated all the branches and PRs so that the git clone command uses the proper tag but the git log command uses the fully qualified branch name.  That means for point releases that we won't need to update the text of the git log command.

kacf:
Automation is on its way, don't worry about that, this is a drop in the ocean compared to all the other version replacements. But what is most correct from a user perspective? I think the tag is the most correct, and this is indeed what I added to the 1.3.x branch. This also saves us from having to specify the upstream name, because it's not guaranteed to always be `origin`.

Do you agree?

estenberg:
sure, sounds good.

drewmoseley:
Updated.  Once the CI tests complete I'll merge this and the other branches.

PR #321, opened 2017-11-15
Move changelogs back into one list only.

kacf:


kacf:
@estenberg

estenberg:
Thanks, I cherry-picked it to 1.3.x as well.

PR #323, opened 2017-11-16
where to find client logs

estenberg:
@kacf @pasinskim can you please make sure this is (still) accurate?

In particular on my QEMU instance, /var/lib/mender/logs/ does not exist, which was a bit suspicious.


kacf:
> In particular on my QEMU instance, /var/lib/mender/logs/ does not exist, which was a bit suspicious.

I haven't tested it, but from what I can piece together by quickly skimming the code, it looks like they are placed one level higher, in `/var/lib/mender`. [This](https://github.com/mendersoftware/mender/blob/master/main.go#L392) is the main piece of the puzzle, it assigns the data store path directly, and I can't see there are any further path additions after that in the constructor.

estenberg:
Thanks @kacf, I tested and you were right. The deployment logs are directly under `/var/lib/mender`.

PR #326, opened 2017-11-21
beaglebone: Rename to beaglebone-yocto.

drewmoseley:
This is due to a change in upstream poky.

### DIFF
--- a/06.Administration/02.Production-installation/docs.md
+++ b/06.Administration/02.Production-installation/docs.md
@@ -318,7 +318,7 @@ git commit -m 'production: adding generated keys and certificates'
 ### Persistent storage
 
 Persistent storage of backend services' data is implemented using
-named [Docker volumes](https://docs.docker.com/engine/tutorials/dockervolumes/?target=_blank).
+named [Docker volumes](https://docs.docker.com/engine/admin/volumes/volumes/?target=_blank).
 The template is configured to mount the following volumes:
 
 - `mender-artifacts` - artifact objects storage


drewmoseley:
This corresponds to https://github.com/mendersoftware/meta-mender/pull/421

01.Getting-started/03.Download-test-images/docs.md
```diff
@@ -21,16 +21,16 @@ Download the Artifacts for your desired device types below:
...
 | Raspberry Pi 3   | [mender-raspberrypi3.sdimg.gz][autoupdate_mender-raspberrypi3_x.x.x.sdimg.gz] | [raspberrypi3_release_1.mender][autoupdate_raspberrypi3_release_1_x.x.x.mender] | [raspberrypi3_release_2.mender][autoupdate_raspberrypi3_release_2_x.x.x.mender] |
 
 
 [autoupdate_vexpress_release_1_x.x.x.mender]: https://d1b0l86ne08fsf.cloudfront.net/master/vexpress-qemu/vexpress_release_1_master.mender
 [autoupdate_vexpress_release_2_x.x.x.mender]: https://d1b0l86ne08fsf.cloudfront.net/master/vexpress-qemu/vexpress_release_2_master.mender
 
-[autoupdate_mender-beaglebone_x.x.x.sdimg.gz]: https://d1b0l86ne08fsf.cloudfront.net/master/beaglebone/mender-beaglebone_master.sdimg.gz
-[autoupdate_beaglebone_release_1_x.x.x.mender]: https://d1b0l86ne08fsf.cloudfront.net/master/beaglebone/beaglebone_release_1_master.mender
-[autoupdate_beaglebone_release_2_x.x.x.mender]: https://d1b0l86ne08fsf.cloudfront.net/master/beaglebone/beaglebone_release_2_master.mender
+[autoupdate_mender-beaglebone-yocto_x.x.x.sdimg.gz]: https://d1b0l86ne08fsf.cloudfront.net/master/beaglebone-yocto/mender-beaglebone-yocto_master.sdimg.gz
```
estenberg: I am not sure if it will be renamed here (prebuilt Artifact / sdimg names). @GregorioDiStefano  or @kacf needs to decide. I guess it makes sense but we probably don't know the right one yet (this new link is broken).

estenberg:
I really hate this upstream change, but not much we can do about it I suppose. If we do not have to rename the images/Artifacts we build I think that would be preferable (writing beaglebone-yocto everywhere is longer and feels strange). But if it could be confusing (some people use beaglebone MACHINE and others use beaglebone-yocto) then I agree this should be done.

01.Getting-started/03.Download-test-images/docs.md
kacf: I think it will be, most likely. I'll keep this in the queue until we actually do an update from master which is not due until the beginning of December. Will let you know.

drewmoseley:
I agree and have the same concern with potential confusion.  The artifacts from the bitbake build will all have the "-yocto" suffix so I thought it would make most sense to add that for the manually created items as well.

01.Getting-started/03.Download-test-images/docs.md
drewmoseley: I did a build locally and all the bits created by bitbake have the "-yocto" appended to the names.

estenberg: OK, sounds good. But why do not our master sdimg/.mender files have that name? We should obviously not have broken links, so it needs to be in sync. :)

drewmoseley: I suspect they will be once we get a successful build with this change integrated.

drewmoseley:
Will resubmit with a better branch name

PR #332, opened 2017-12-05
Remove github links to raw flash components.

drewmoseley:
This is more consistent with usage in other portions of the docs and removes one
more place to update when versions change


drewmoseley:
@estenberg @kacf any concerns with this change?

kacf:
Fine with me, I don't think it's important for users to know the details behind those links.

estenberg:
:+1:

PR #345, opened 2017-12-21
beaglebone: Rename to beaglebone-yocto.

drewmoseley:
This is due to a change in upstream poky.


drewmoseley:
@kacf is it time to look at this one again?  I seem to recall that this rename is complete in meta-mender.  I've basically just done a global search/replace here but I'm not sure if every replacement here is correct.  I replaced the text in the invocations of mender-artifact, as well as in all the pre-generated images and such.  Are we planning to update those?

kacf:
We will have to hold off on this for now, because Mender 1.4.0 is released in March, long before the next Yocto release which will be the first one to contain the new `beaglebone-yocto` name. So the documentation will have to refer to the old name for the time being, so that 1.4 will be correct when we branch it.

That is, unless we want to mention both. But there are quite a lot of places, so I suggest that we don't and just make the move once the next Yocto release is out.

PR #357, opened 2018-01-22
migrate from dynomite to redis 

mchalski:
goes with: https://github.com/mendersoftware/integration/pull/429

@mendersoftware/rndity @maciejmrowiec


maciejmrowiec:
```
ERROR: Volume mender-dynomite-db declared as external, but could not be found. Please create the volume manually using `docker volume create --name=mender-dynomite-db` and try again.
Traceback (most recent call last):
  File "./test_docs.py", line 119, in <module>
    assert False, "shell script extracted from docs failed."
AssertionError: shell script extracted from docs failed.
```

mchalski:
yes, it needs the referenced PR to run.

I'll start a parametrized build then

PR #365, opened 2018-02-07
1.3.x

drewmoseley:


oleorhagen:
It is only merged in master, so no.

drewmoseley:
OK.  I pushed an updated branch with the reboot hardening bits removed.

PR #384, opened 2018-03-13
 Replaced wget with git clone in getting-started doc.

oleorhagen:


drewmoseley:
Is there a reason there are 3 commits with the same title?  Should they be squashed into one?

01.Getting-started/02.Create-a-test-environment/docs.md
```diff
@@ -14,22 +14,14 @@ taxonomy:
 In a working directory, download the Mender integration
 environment:
 
-[start_autoupdate_integration_archive_x.x.x.tar.gz]: #
-
 ```bash
-curl -L https://github.com/mendersoftware/integration/archive/master.tar.gz | tar xz
+git clone https://github.com/mendersoftware/integration integration-master
```
kacf: You need to include which branch/tag to check out. Use the example from [the production setup docs](https://docs.mender.io/1.4/administration/production-installation#basic-preparation-setup).

01.Getting-started/02.Create-a-test-environment/docs.md
```diff
@@ -14,22 +14,14 @@ taxonomy:
 In a working directory, download the Mender integration
 environment:
 
-[start_autoupdate_integration_archive_x.x.x.tar.gz]: #
-
```
kacf: Just leave these for now. They don't currently work, but I will eventually get around to fixing these, and at that point I'd like to be able to search for them.

oleorhagen:
Fixed! and apologies. Was supposed to be a quick one-liner. Lesson: Always keep your master clean.

oleorhagen:
Also: question? why keep the autoupdate tar tags, when it is no longer in use?

kacf:
They kind of are in use, but only manually. The plan was to make them automatic, but this has not happened yet.

kacf:
You probably need to cherry-pick this with the necessary adjustments to the `-b` parameter.

PR #389, opened 2018-03-15
Replaced wget with git clone in getting-started doc.

oleorhagen:
Changelog: Title

Signed-off-by: Ole Petter <olepor@matnat.uio.no>


drewmoseley:
@oleorhagen I looked at all the different branch versions of this.  I checked out the git versions and compared them with my previous tarball downloads and they all look good to me.  From my perspective these can be merged.

I'm not sure if there are other things dependent on these so I'll wait for an ack from @kacf to merge.

PR #393, opened 2018-04-24
Update artifact names for 1.4.0 release

drewmoseley:

### DIFF
--- a/01.Getting-started/04.Deploy-to-physical-devices/docs.md
+++ b/01.Getting-started/04.Deploy-to-physical-devices/docs.md
@@ -202,7 +202,7 @@ the steps needed to edit the root file system contained in this Artifact are:
 [start_autoupdate_beaglebone_release_1_x.x.x.mender]: #
 
 ```bash
-mkdir beaglebone_release_1 && tar -C beaglebone_release_1 xvf beaglebone_release_1_master.mender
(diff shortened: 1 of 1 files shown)


drewmoseley:
Looks like we missed a few artifact name changes for 1.4.0.  Did I get this right?

estenberg:
Yes, this looks like a correct change. However, this should have been done by automation, I think. Maybe @kacf can comment on why it did not automatically update?

drewmoseley:
@estenberg I think the reason that particular invocation requires the leading '-' is that there was a previous command line parameter that used it; ie the '-C <dir>' option.  It definitely failed for me without it although a direct "tar xvf blah" does work.  I am cherry picking that back to older branches as well as to master.

kacf:
> Yes, this looks like a correct change. However, this should have been done by automation, I think. Maybe @kacf can comment on why it did not automatically update?

No, that task is [still open](https://tracker.mender.io/browse/MEN-1140), I'm afraid. There has been talk of moving away from Grav, which would impact this a lot, so I haven't made a move here yet. @michaelatmender, any news on that?

michaelatmender:
@kacf nothing specifically, just a general possibility that one day we'll move the docs. We should continue assuming we'll build the site with Grav for the medium-longish term future

kacf:
Ok, I will act as if we stay with Grav then.

PR #401, opened 2018-05-03
Add new section for Server Integration topics

drewmoseley:


estenberg:
I would actually move this before "Administration", as "Administration" is typically something you do just once. But I still think APIs should be below "Administration" because "APIs" is more of a reference documentation.

However, I do understand you want to keep "APIs" and "Server integration" next to each other, though, so I don't have a strong position on this.

07.Server-Integration/01.Overview/docs.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Overview
+taxonomy:
+    category: docs
+---
+
+The Mender backend microservices are all accessible using an HTTPS API.  These APIs can be used to configure the server (for example, pre-authorizing devices) or implementing custom workflows (for example, integrating the Mender server into an existing device management system.)
```
estenberg: We use "server" instead of "backend".

I am also thinking this first one should be "Using the APIs" instead of "Overview" (which feels more like a architecture/diagram type of title rather than use-case focused).
In "Using the APIs" we can explain quickly the two types (management vs. device), how to set up a CLI environment and (soon) how to set up mender-cli.

estenberg: And tokens, management vs. device. :)

07.Server-Integration/chapter.md
```diff
@@ -0,0 +1,12 @@
...
+title: Server Integration
+taxonomy:
+    category: docs
+---
+
+### Chapter 7
+
+# Server Integration
+
+Using the Mender APIs for custom setups.
```
estenberg: Would say "Integrating the Mender server through the REST APIs." In particular I don't think this is necessarily "custom", but maybe I am getting a bit too nit-picky. :)

estenberg:
In the directory name, would use lowercase "integration", so "..Server-integration", like the existing directories.

07.Server-Integration/01.Overview/docs.md
drewmoseley: I updated all per your suggestions except for this one.  I'm not sure what text you want for:
> explain quickly the two types (management vs. device), how to set up a CLI 
> environment and (soon) how to set up mender-cli.

and

> And tokens, management vs. device

estenberg:
Thanks, just one last one would rename: 06.Server-integration/01.Overview/ to 06.Server-integration/01.Using-the-apis/ to align with title.

drewmoseley:
Done.  Plus I (think) I fixed the travis failure.

drewmoseley:
@estenberg is this ok to merge now?

estenberg:
Looks good! Should ensure it goes into 1.5 if that is already branched (otherwise it will automatically), as well as 1.4. But maybe wait until we have some more content?

PR #409, opened 2018-05-09
MEN-1805: Setting wifi ssid and password in an image

oleorhagen:
Changelog: Title

### DIFF
--- a/01.Getting-started/04.Deploy-to-physical-devices/docs.md
+++ b/01.Getting-started/04.Deploy-to-physical-devices/docs.md
@@ -126,6 +126,20 @@ Gateway=$IP_OF_MENDER_SERVER_FROM_DEVICE
 
 ! If you have a static IP address setup for several devices, you need several disk images so each get different IP addresses.
 
+## Wifi connectivity
+
+The raspberrypi demo image comes with Wifi connectivity enabled by default, thus the only thing needed in order for your device to connect to your network is setting the correct `<ssid>` and `<password>` in the `wpa_supplicant-nl80211@wlan0.conf` file on your device. First set your `<password>` and `<ssid>`, and `<sdimg>` path as shell variables:
+```bash
+NW_SSID=<ssid>
+NW_PASSWORD=<password>
+MENDER_IMGPATH=<sdimg>
+```
+And then running:
+```bash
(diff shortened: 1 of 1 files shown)


01.Getting-started/04.Deploy-to-physical-devices/docs.md
```diff
@@ -126,6 +126,14 @@ Gateway=$IP_OF_MENDER_SERVER_FROM_DEVICE
 
 ! If you have a static IP address setup for several devices, you need several disk images so each get different IP addresses.
 
+## Wifi connectivity
+
+The raspberrypi demo image comes with Wifi connectivity enabled by default, thus the only thing needed in order for your device to connect to your network is setting the correct `<ssid>` and `<password>` in the `wpa_supplicant-nl80211@wlan0.conf` file on your device. Therefore running:
+```bash
+mender-artifact cat <sdimg>:/etc/wpa_supplicant/wpa_supplicant-nl80211-wlan0.conf | sed 's#[@]MENDER_DEMO_WIFI_PASSKEY[@]#<password>#' | sed 's#[@]MENDER_DEMO_WIFI_PASSKEY[@]#<ssid>#' | mender-artifact cp <sdimg>:/etc/wpa_supplicant/wpa_supplicant-nl80211-wlan0.conf"
```
estenberg: Would set these as shell variables first, so the user does not need to edit this long command.

E.g.

Set your ssid and password...
SSID="example-ssid"
PASSWORD="abc"

Then run the following command to insert your wifi into the wpa-supplicant configuration...
mender-artifact cat...

estenberg:
Does this work for Hosted Mender images as well? If so it would be really great if we could update https://hosted.mender.io/ui/#/help/connecting-devices/provision-a-demo/raspberry-pi-3 as well.

oleorhagen:
I'm not sure what the hosted mender images look like @estenberg, but if they are in the format that is handled by the mender artifact tool, it should work. What do the mender images look like @maciejmrowiec ?

estenberg:
good! @oleorhagen it is a separate task, but could you please update the Hosted Mender RPi3 help as well then? It is in the GUI repository.

oleorhagen:
Sure: https://github.com/mendersoftware/gui/pull/363

oleorhagen:
Btw, Is this ready to be merged?

PR #412, opened 2018-05-22
Add diagrams for state script transitions and use cases

drewmoseley:


drewmoseley:
This adds my state machine diagram as well as customized versions for the documented use cases. It's hard to see what this looks like here in PR form.  I ran mender-docs-site locally and did a print to PDF.  It's not a perfect representation but should give you an idea of how it looks.
[State scripts _ Mender documentation.pdf](https://github.com/mendersoftware/mender-docs/files/2028823/State.scripts._.Mender.documentation.pdf)

kacf:
Actually you can see it if you click the little document icon on each file ("rich diff").

Big thumbs up from me!

drewmoseley:
Yeah, I thought the same thing but the Google Draw utility didn't have anything better.  If you come up with a better choice we can always update them later.  The source documents are in Google drive.

pasinskim:
Really good stuff @drewmoseley!

PR #419, opened 2018-05-26
Rename IMAGE_BOOTLOADER* variables to MENDER_IMAGE_BOOTLOADER*.

drewmoseley:
These are mender-specific; rename them so it's more obvious.


drewmoseley:
This goes with: https://github.com/mendersoftware/meta-mender/pull/517

kacf:
For stability I won't cherry-pick a variable rename to rocko. Do you think we should merge (and cherry-pick) this PR immediately, even though rocko is still the most recent stable branch of ours? I think I lean towards yes, even though it is slightly premature. It won't be long before it's true and this way we won't forget.

drewmoseley:
Yeah, the independent versioning of meta-mender and mender-docs does cause some confusion here.  I think it's ok to have this one cherry-picked.

kacf:
Alright, cherry-picked to 1.4.x and 1.5.x as well.

PR #425, opened 2018-06-04
Clarify state scripts and managed mode.

drewmoseley:


04.Artifacts/07.State-scripts/docs.md
```diff
@@ -6,6 +6,8 @@ taxonomy:
 
 The Mender Client has the ability to run pre- and postinstall scripts, before and after it writes the root file system. However, Mender state scripts are more general and useful than pre/postinstall scripts because they can be run between *any* state transition, not just (before/after) the install state. For some examples of usage, see [example use cases](#example-use-cases).
 
+!!! Note that state scripts are a feature of managed mode.  State scripts are not executed when running Mender in [standalone mode](../../architecture/overview#modes-of-operation)).
+
```
mirzak: IMO it should be enough to to put the "note" at this location as it is the relevant section for the "state script" feature.

Readers reading 01.Overview/docs.md, will get introduced to something called "state scripts" without having read 07.State.scripts/docs.md, which might be confusing.

estenberg: I agree, this is the place to put it (not in architecture)!

PR #430, opened 2018-06-05
variables: make it clear that IMAGE_ROOTFS_SIZE is expressed in kiB

mirzak:
It states in the Yocto docs that it is expressed in kiB, but if user never clicks the link they will probably assume that it is in MB (happened to me)


mirzak:
Thanks!

PR #437, opened 2018-06-06
Update links to Yocto docs

mirzak:
To future proof the links updated links to use the /latest/
link which should always point to the latest release of Yocto.

Note that the Yocto Project Quick Start manual does not seem
to support /latest/ link, so instead update to the latest
available which is 2.4.

Signed-off-by: Mirza Krak <mirza.krak@endian.se>


mirzak:
Thanks!

PR #441, opened 2018-06-13
GRUB/BIOS documentation

kacf:


kacf:
GRUB integration guide is still less than one page! :-D

kacf:
Yes, 1.4.x and 1.5.x need this. Both had no conflicts at all, so I went ahead and pushed both directly.

estenberg:
Nice!

PR #459, opened 2018-07-24
Update compatibility tables

drewmoseley:


drewmoseley:
Note specifically that I added an "available" status in the client/Yocto table to indicate that it was not supported.  Objections?

Also, I'm not sure about mender-artifact versions.  I see a 2.2 in the repo but I don't know where it should be in the table.  Please let me know.

estenberg:
@drewmoseley very nice!

* Instead of "available" I would use "community". This aligns with how YP refers to its support level (https://wiki.yoctoproject.org/wiki/Releases) and what we will have for device support soon (the plan is reference for BBB and RPi3 and community level for anything else). Instead of "yes" we should probably say "stable" then also (like YP).

* Don't know about mender-artifact versioning, perhaps @kacf could comment on that?

* Correct, only API v1 so far, but we will get API v2 in Mender 1.7 (end of this year).

drewmoseley:
OK.  I'll wait for @kacf to return to comment on the mender-artifact versioning

02.Architecture/03.Compatibility/docs.md
```diff
@@ -39,6 +41,9 @@ The [Mender Artifact format](../mender-artifacts) is managed by the [Mender Arti
 | Mender 1.0.x / mender-artifact 1.0.x | yes         | no          |
 | Mender 1.1.x / mender-artifact 2.0.x | yes         | yes         |
 | Mender 1.2.x / mender-artifact 2.1.x | yes         | yes         |
+| Mender 1.3.x / mender-artifact 2.1.x | yes         | yes         |
+| Mender 1.4.x / mender-artifact 2.1.x | yes         | yes         |
+| Mender 1.5.x / mender-artifact 2.1.x | yes         | yes         |
```
kacf: 1.4.x and 1.5.x use 2.2.x of mender-artifact.

drewmoseley:
OK.  Updated.  Once the tests complete I'll merge and backport.

PR #468, opened 2018-08-13
Remove description of MENDER_STORAGE_DEVICE_BASE.

kacf:
It is now fully automatic, and should not need to be set by anyone
anymore.

### DIFF
--- a/03.Devices/02.Partition-layout/docs.md
+++ b/03.Devices/02.Partition-layout/docs.md
@@ -74,12 +74,10 @@ In order to select the storage device where the partitions are expected to be lo
 MENDER_STORAGE_DEVICE = "/dev/mmcblk0"
 ```
 
-! For memory card storage and some other types of storage, the default way to refer to partitions is to add a "p" and then the number of the partition (for example `/dev/mmcblk0p1`). If you're using a storage type which doesn't follow this scheme (for example `/dev/sda1`), then you also need to set `MENDER_STORAGE_DEVICE_BASE` to the correct value, such as `MENDER_STORAGE_DEVICE_BASE = "${MENDER_STORAGE_DEVICE}"`. The default is the value of `MENDER_STORAGE_DEVICE` plus a "p".
-
 
 ###More detailed storage configuration
 
-If you need more fine grained control over which partitions Mender will use, you can set one or more the following variables to specific partition strings:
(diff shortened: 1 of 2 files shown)


03.Devices/02.Partition-layout/docs.md
```diff
@@ -74,12 +74,10 @@ In order to select the storage device where the partitions are expected to be lo
 MENDER_STORAGE_DEVICE = "/dev/mmcblk0"
 ```
 
-! For memory card storage and some other types of storage, the default way to refer to partitions is to add a "p" and then the number of the partition (for example `/dev/mmcblk0p1`). If you're using a storage type which doesn't follow this scheme (for example `/dev/sda1`), then you also need to set `MENDER_STORAGE_DEVICE_BASE` to the correct value, such as `MENDER_STORAGE_DEVICE_BASE = "${MENDER_STORAGE_DEVICE}"`. The default is the value of `MENDER_STORAGE_DEVICE` plus a "p".
-
 
 ###More detailed storage configuration
 
-If you need more fine grained control over which partitions Mender will use, you can set one or more the following variables to specific partition strings:
+If you need more fine grained control over which partitions Mender will use, you can set one or more the following variables to specific partition strings, using `MENDER_STORAGE_DEVICE_BASE`:
```
drewmoseley: Should this refer to MENDER_STORAGE_DEVICE instead?

estenberg: also, below this there are mentions of MENDER_STORAGE_DEVICE_BASE too:

MENDER_BOOT_PART = "${MENDER_STORAGE_DEVICE_BASE}1"
MENDER_DATA_PART = "${MENDER_STORAGE_DEVICE_BASE}5"
MENDER_ROOTFS_PART_A = "${MENDER_STORAGE_DEVICE_BASE}2"
MENDER_ROOTFS_PART_B = "${MENDER_STORAGE_DEVICE_BASE}3"

estenberg: Maybe this entire section should be removed?

kacf: No, this is the only place where you do need `MENDER_STORAGE_DEVICE_BASE`, and you cannot replace it with `MENDER_STORAGE_DEVICE`; the result will be wrong. The difference from before is that you no longer need to *set* `MENDER_STORAGE_DEVICE_BASE`, only use it.

kacf:
Thanks! :-)

kacf:
Cherry-picked to 1.6.x, but not older branches.

PR #473, opened 2018-09-04
mention usage of 'jq' utility to decoded and pretty-print JSON

mirzak:
It is commonly used together with 'curl' and JSON, but it is also
required to decode certain values in responses, e.g an artifact
URI will contain raw values that need to be decoded by a JSON
parses.

Some more background info can be found in MEN-2082.

Changelog: None

### DIFF
--- a/06.Server-integration/01.Using-the-apis/docs.md
+++ b/06.Server-integration/01.Using-the-apis/docs.md
@@ -47,7 +47,13 @@ For example, to upload an Artifact you can run the following command:
 
 ## Set up shell variables for cURL
 
-There are many ways to call http-based REST APIs, but the most generally available method is most likely the command line utility `curl`.
+There are many ways to call http-based REST APIs, but the most generally available method is the command line utility `curl`. The `jq` utility is commonly used to decode JSON messages and display them in human-readable format. REST API responses can contain raw data and must run through a JSON parser (`jq`).
+
+You can easily get both using:
+
+```bash
+sudo apt-get install curl jq
+```
 
 Open a terminal, which we will use in the following to call the Mender server's REST APIs. First set a shell variable with the URI of your server:
 ```bash
@@ -80,7 +86,7 @@ You should now have an API token you can use to call any of the [Mender server m
 To verify you can call the server APIs, list the users of your Mender server instance:
 
 ```bash
-curl -H "Authorization: Bearer $JWT" $MENDER_SERVER_URI/api/management/v1/useradm/users
+curl -H "Authorization: Bearer $JWT" $MENDER_SERVER_URI/api/management/v1/useradm/users | jq
 ```
 
 If it succeeds it will return something like the following:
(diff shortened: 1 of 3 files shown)


drewmoseley:
Looks good to me.  Once @maciejmrowiec approves we'll get this merged. I think this should also go back to the older docs branches.

estenberg:
There are other references in the doc to use python -m json.tool. I would change those to jq as well then.

mirzak:
Updated, replaced `python -m json.tool` occurrences with `jq`. Also added hint on how to install `jq` in the using the APIs section.

06.Server-integration/01.Using-the-apis/docs.md
```diff
@@ -47,7 +47,13 @@ For example, to upload an Artifact you can run the following command:
 
 ## Set up shell variables for cURL
 
-There are many ways to call http-based REST APIs, but the most generally available method is most likely the command line utility `curl`.
+There are many ways to call http-based REST APIs, but the most generally available method is most likely the command line utility `curl`. The `jq` utility is commonly used to decode JSON messages and to get "pretty-print". REST API responses can contain raw data and must run trough a JSON parses (`jq`).
```
drewmoseley: The wording here is clunky.  Can you change it to "the most generally available method is the command line utility"?
Also I would say "The jq utility is used to decode JSON messages and display them in human-readable format".
Finally, the word "through" is misspelled

06.Server-integration/02.Preauthorizing-devices/docs.md
```diff
@@ -89,11 +89,11 @@ Secondly, To make sure that the device is not in an existing authentication set,
...
 
 ```bash
-curl -H "Authorization: Bearer $JWT" $MENDER_SERVER_URI/api/management/v1/admission/devices | python -m json.tool > /tmp/admission.json
-curl -H "Authorization: Bearer $JWT" $MENDER_SERVER_URI/api/management/v1/devauth/devices | python -m json.tool > /tmp/devauth.json
+curl -H "Authorization: Bearer $JWT" $MENDER_SERVER_URI/api/management/v1/admission/devices | jq > /tmp/admission.json
+curl -H "Authorization: Bearer $JWT" $MENDER_SERVER_URI/api/management/v1/devauth/devices | jq > /tmp/devauth.json
 ```
 
-!!! To make the response more readable, we use the `json.tool` module of python to indent it. If it is not available on your system you can omit this pipe or replace it with a different indentation tool (remove or replace `| python -m json.tool`).
+!!! To make the response more readable, we use the `jq` utility to decode and pretty-print. If it is not available on your system you can omit this pipe or replace it with a different indentation tool (e.g `python -m json.tool`).
```
drewmoseley: I would just say "use the jq utility to decode it." We already say that we are doing that to make it more readable.

mirzak:
Thanks for the review @drewmoseley , updated.

PR #475, opened 2018-09-05
Add troubleshooting section for rocko -> sumo update

mirzak:
As there is now a patch in sumo that targets a specific U-boot version,
which will not apply if an older U-boot fork is used.

### DIFF
--- a/201.Troubleshooting/01.Yocto-project-build/docs.md
+++ b/201.Troubleshooting/01.Yocto-project-build/docs.md
@@ -216,3 +216,53 @@ By default, meta-mender will produce a UEFI image (`uefiimg`) when integrating w
 MENDER_FEATURES_ENABLE_append = " mender-image-sd"
 MENDER_FEATURES_DISABLE_append = " mender-image-uefi"
 ```
+
+# When I update Yocto version from rocko to sumo U-boot patches do not apply
+
+This is what the error message might look like.
+
+```
+ERROR: u-boot-custom-2017.03-r0 do_patch: Command Error: 'quilt --quiltrc /home/user/git/poky/build/tmp/work/imx6ullevk-poky-linux-gnueabi/u-boot-custom/2017.03-r0/recipe-sysroot-native/etc/quiltrc push' exited with 0  Output:
+Applying patch 0006-env-Kconfig-Add-descriptions-so-environment-options-.patch
+can't find file to patch at input line 19
+Perhaps you used the wrong -p or --strip option?
+The text leading up to this was:
+--------------------------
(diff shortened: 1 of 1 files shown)


mirzak:
It is from this mailing list thread, https://groups.google.com/a/lists.mender.io/forum/#!topic/mender/AstDvNzHiB0.

mirzak:
I like the idea. We also have the "generic" patches that are different for different versions which we would be useful to apply based on U-boot version.

Though PV might not be the right variable to parse. A lot of forks do not follow the "upstream" convention for their U-boot recipes. Meaning that PV could be something made-up or set to "git" and as this tries to solve a problem that is common with forks it might not end being that useful. 

Though at the tip of the Makefile there is version information that should be "more accurate", if it possible to parse that would be more reliable.

kacf:
> Though at the tip of the Makefile there is version information that should be "more accurate", if it possible to parse that would be more reliable.

I think this would be too late in the parsing. `Makefile` depends on `SRC_URI` which depends on "tip of Makefile" = circular dependency. :-(

> Though PV might not be the right variable to parse. A lot of forks do not follow the "upstream" convention for their U-boot recipes. Meaning that PV could be something made-up or set to "git" and as this tries to solve a problem that is common with forks it might not end being that useful.

True for some forks, but I have seen many of them follow the version scheme as well.

We could make it follow this logic then:

1. If `PV` version is in YYYY.MM format (possibly we can allow trailing stuff like "-gitabcd123" or "-r3"):
    1. Add patch if 2018.01 or later
    1. Else don't add patch
1. Version is unknown/vendor-specific: Add patch

Then we could still keep these docs for the bottom case, where we cannot detect problems automatically.

mirzak:
Yeah makes sense.

drewmoseley:
I definitely agree that if we can fix it in code rather than in docs that is preferable. Can we do that in a reasonable time given the current team load?

mirzak:
But even if it is fixed in code, it will not fix every case and think that this is still valid.

I vote merge this and update when we have a fix in code (if needed). With 1.6 being released soon I suspect people will start migrating and will hit this.

kacf:
@mirzak: Can you cherry-pick this? It will be needed at least for 1.5.x and 1.6.x.

mirzak:
Will cherry-pick, but does it really need to be in 1.5.x? Isn't that based on Rocko?

kacf:
They are independent releases, so it's wrong to say it's based on any one of them. What matters is which meta-mender branches contain the 1.5 recipes, and both rocko and sumo contain them, which means that upgrading from rocko to sumo is possible with 1.5.

mirzak:
Thanks for explaining. Cherry-picked to both 1.5.x and 1.6.x

PR #476, opened 2018-09-06
MEN-1140: Implement autoversioning of Mender components

kacf:


kacf:
> Looks like a great feature, but hard to see if it does the right thing or not without testing a bit the output.

Yes, I thought about adding a test, but I think manual verification is actually better here, because it is something we have to verify anyway at release time. The test could easily have the same flaw as the code it's testing.

PR #484, opened 2018-09-10
FIX: Do not read and write to a file in the same pipe-chain.

oleorhagen:
Changelog: None

### DIFF
--- a/01.Getting-started/04.Deploy-to-physical-devices/docs.md
+++ b/01.Getting-started/04.Deploy-to-physical-devices/docs.md
@@ -88,7 +88,7 @@ server when it starts.
 ### Insert the address of Mender server
 
 ```bash
-mender-artifact cat <imgname>.sdimg:/etc/hosts | sed "\$a ${IP_OF_MENDER_SERVER_FROM_DEVICE} docker.mender.io s3.docker.mender.io" | mender-artifact cp <imgname>.sdimg:/etc/hosts
+mender-artifact cat IMGNAME.sdimg:/etc/hosts | sed "\$a ${IP_OF_MENDER_SERVER_FROM_DEVICE} docker.mender.io s3.docker.mender.io" > tmpf; mender-artifact cp tmpf IMGNAME.sdimg:/etc/hosts && rm tmpf
 ```
 
 Then you can check the contents of your 'etc/hosts' file by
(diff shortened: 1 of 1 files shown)


oleorhagen:
Fix for: https://tracker.mender.io/browse/MEN-2087

01.Getting-started/04.Deploy-to-physical-devices/docs.md
```diff
@@ -88,7 +88,7 @@ server when it starts.
 ### Insert the address of Mender server
 
 ```bash
-mender-artifact cat <imgname>.sdimg:/etc/hosts | sed "\$a ${IP_OF_MENDER_SERVER_FROM_DEVICE} docker.mender.io s3.docker.mender.io" | mender-artifact cp <imgname>.sdimg:/etc/hosts
+mender-artifact cat <imgname>.sdimg:/etc/hosts | sed "\$a ${IP_OF_MENDER_SERVER_FROM_DEVICE} docker.mender.io s3.docker.mender.io" > tmpf; mender-artifact cp tmpf <imgname>.sdimg:/etc/hosts && rm tmpf
```
oleorhagen: You reckon this tempfile will do? Is the name appropriate?

kacf: I think it's ok. Using `mktemp` will be complicated here because the name needs to be saved between the two calls, and this is supposed to be a oneliner. For simplicity I think I would stick to a hardcoded name in this particular example.

kacf: However, given that the `>` character is actually used in the command now, perhaps you can switch `<imgname>` with `IMGNAME` (note `<>` gone), so that it both stands out more, and can't be confused with actual redirection.

oleorhagen: Another option could be: ```
TMPFILE=$(mktemp); mender-artifact cat <imgname>.sdimg:/etc/hosts | sed "\$a ${IP_OF_MENDER_SERVER_FROM_DEVICE} docker.mender.io s3.docker.mender.io" > $TMPFILE; mender-artifact cp $TMPFILE <imgname>.sdimg:/etc/hosts && rm $TMPFILE
```

kacf:
Thanks, confirmed working!

Merged manually in cae63ab80bd12c3e8ead793db08dfbf12b92fa8e.

PR #486, opened 2018-09-11
Updated device integration list to include mender_boot_part_hex variable

dellgreen:
At some point a new 'mender_boot_part_hex' variable was introduced.
This variable needs to be changed along with 'mender_boot_part' from linux
to perform a successful kernel swapping test, otherwise you end up with
a swapped rootfs, but the kernel from the previous rootfs.

It should be noted that if testing from uboot shell then you can
run the 'mender_altbootcmd' which will swap both variables for you.

Changelog: None

Signed-off-by: Dell Green <dell.green@ideaworks.co.uk>


mirzak:
Looks good to me

drewmoseley:
@dellgreen thanks for the contribution.  I'll cherry pick this to the older docs branches as well.

PR #487, opened 2018-09-11
Updated device integration checklist with 'pgrep' alternative.

dellgreen:
On some Yocto builds, depending on the image config, 'pgrep' is not
configured to be installed by default as part of busybox recipe.
However 'pidof' is available. Documentation has been updated to
give the user the option of either depending upon their setup.

Changelog: None

Signed-off-by: Dell Green <dell.green@ideaworks.co.uk>


dellgreen:
yes indeed, my image is based on a minimal image

kacf:
Agree we should just drop pgrep entirely. Closed and replaced with #493.

PR #493, opened 2018-09-12
Switch pgrep to pidof, which is Busybox friendly.

kacf:
Thanks to Dell Green <dell.green@ideaworks.co.uk> for spotting this.


kacf:
Cherry-picked to 1.5.x and 1.6.x as well.

PR #495, opened 2018-09-12
1.6.x update with new mender-artifact steps

kacf:


kacf:
@estenberg: Your changes with mine and Ole's as well, which should cover all bases

PR #509, opened 2018-09-21
Clarify systemd settings per Yocto release.

drewmoseley:


drewmoseley:
@kacf is this enough clarification?

estenberg:
I can't verify if this is true or not. But assuming it is, it is in the right place and clear to me.

PR #515, opened 2018-09-25
Use validate subcommand for signature check.

drewmoseley:


drewmoseley:
This should address part of https://tracker.mender.io/browse/MEN-2155

PR #519, opened 2018-09-27
Minor use case examples improvements

kacf:
```
commit 8187d799facea6d8b985a0ece0ddf621ce127d20
Author: Kristian Amlie <kristian.amlie@northern.tech>
Date:   Thu Sep 27 16:11:26 2018

    Expand the use case example slightly for Update Confirmations.
    
commit 49bf95e03418a1eab2e14bfa9e3c1e4e2ba0fc24
Author: Kristian Amlie <kristian.amlie@northern.tech>
Date:   Thu Sep 27 16:07:29 2018

    Avoid suggesting state script that can break migration.
    
    Clarify that it is ArtifactInstall_Leave that should be used for
    migrations, not ArtifactReboot_Enter. This is important because if a
    spontaneous reboot happens while inside ArtifactReboot_Enter, it will
    be ignored, and hence the new partition will be entered with an
    incomplete migration in place. With ArtifactInstall_Leave it will
    instead cause a rollback, triggering a potential ArtifactFailure
    script which can undo the half done migration.
    
```

### DIFF
--- a/04.Artifacts/07.State-scripts/docs.md
+++ b/04.Artifacts/07.State-scripts/docs.md
@@ -100,7 +100,7 @@ In this case, application data like a user profile is stored in an SQLite databa
 #### Update confirmation by end user
 For many devices with a display that interacts with an end user, it is desirable to ask the user before applying the update. You have probably seen this on a smartphone, where it will ask you if you want to update to the latest release of Android or iOS and it only starts after you hit "Apply".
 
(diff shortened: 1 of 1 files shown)


kacf:
@drewmoseley: Just a couple of things that caught my eye while researching for update modules.

kacf:
I think so, so I went ahead and cherry-picked to both of them.

PR #522, opened 2018-09-28
Board integration is a new taxonomy concept.

estenberg:
It only applies to the actual Mender integration of the board
and any related low-level sections like the reference boards, output images.

Devices will still be the common term to describe a product,
identity and the over-arching concept.

Signed-off-by: Eystein Måløy Stenberg <eystein@mender.io>

### DIFF
--- a/01.Getting-started/04.Deploy-to-physical-devices/docs.md
+++ b/01.Getting-started/04.Deploy-to-physical-devices/docs.md
@@ -27,9 +27,9 @@ a SD card to store the OS, so you will need one SD card
 
 ### Disk image and Artifacts
 
-Get the disk image and Artifacts for your device(s) from [Download demo images](../download-test-images).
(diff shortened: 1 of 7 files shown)


drewmoseley:
@estenberg does this need to be backported?

estenberg:
@drewmoseley yes, I think we should backport to 1.6 and 1.5 as it might be confusing if we use a different taxonomy in different branches.

This might be a bit tricky backport as master diverged quite a bit. Let me know if you want me to do it!

drewmoseley:
@estenberg yes, I'll leave the backport to you. ;-)

estenberg:
Backported to 1.6.x.
Skipped 1.5.x backport because it had diverged quite a bit wrt to board support (with GRUB, UEFI introduced in 1.6).

PR #524, opened 2018-10-02
Added documentation for "Servers" configuration entry

alfrunes:
backlog ID: MEN-2175


05.Client-configuration/04.Configuration-file/50.Configuration-options/docs.md
```diff
@@ -32,6 +32,16 @@ system based on Yocto configuration and rarely needs to be modified.
 The Linux device that contains root filesystem B. This is set by the build
 system based on Yocto configuration and rarely needs to be modified.
 
+#### Servers
+
+An array of json objects of the format
+`[{ServerURL: "https://mender-server.com"},
+{ServerURL: "https://mender-server2.com"}, ...]`, where `ServerURL` is as
+explained in the next paragraph. If `Servers` entry is specified, the
```
kacf: Would rephrase to 
```
... where `ServerURL` has the same format as the plain [`ServerURL` attribute](#serverurl).
```
in case we add anything in between here.

05.Client-configuration/04.Configuration-file/50.Configuration-options/docs.md
```diff
@@ -32,6 +32,16 @@ system based on Yocto configuration and rarely needs to be modified.
...
+configuration cannot contain an additional `ServerURL` entry in the top level
+of the json configuration. Upon an unserved request (4XX/5XX-response codes) the
+client will attempt the next server on the list in the given order.
+
 #### ServerURL
 
 The server URL which is used as the basis for API requests. This should be set
```
kacf: I would add a note under `ServerURL` as well, mentioning that `Servers` conflicts with `ServerURL`.

alfrunes:
@kacf; The requested changes has been made.

estenberg:
This is more of a design comment, but can we
* Use the existing attribute and just test if it's a list or string (for backward compatibility)?
* Why do we use JSON **object** inside Servers, if the URL is the only entry? Wouldn't it be better to have something like servers : [ "server1", "serve2", ...] rather than servers: [{ "ServerURL" : "server1"}, { "ServerURL" : "server2"}  ] ? Is there a reason for this deeper nesting?

kacf:
> Why do we use JSON object inside Servers, if the URL is the only entry? Wouldn't it be better to have something like servers : [ "server1", "serve2", ...] rather than servers: [{ "ServerURL" : "server1"}, { "ServerURL" : "server2"} ] ? Is there a reason for this deeper nesting?

Yes, it's because the tenant token can also be part of the object. But it's good you brought it up, because that is missing from the documentation! @alfrunes, TenantToken should be mentioned there as well, and a link should be added from the original TenantToken entry, similar to the ServerURL entry.

> Use the existing attribute and just test if it's a list or string (for backward compatibility)?

Not really a good fit due to the comment above about TenantToken.

alfrunes:
As @kacf says the idea was that multiple per-server related variables such as `TenantToken`, unique `ServerCertificate` paths and possible other fields would be added later on. At the moment these are left out due to the amount of restructuring that it would imply to maintain a list of `TenantToken`s.

kacf:
Ah, so it's not currently possible?

alfrunes:
No.

alfrunes:
Sure thing.

PR #529, opened 2018-10-05
fix Raspberry Pi alignment example

mirzak:
It was still using the old MENDER_PARTITION_ALIGNMENT_KB variable
which was convert to MENDER_PARTITION_ALIGNMENT and the value has
to be in bytes now.


mirzak:
Or do we want want "on this version and older you have to do this and on newer to that"? Because I noticed that we have that for legacy method of choosing U-boot on RPi

drewmoseley:
@mirzak yes I think detailing which versions need which is the right approach since our docs version is independent of Yocto versioning.

mirzak:
@drewmoseley , updated

mirzak:
Re-triggered the test, green. Merging.

PR #542, opened 2018-10-18
reference Mender Hub alongside commercial board integration

estenberg:
Removed reference devices I saw in these areas, this status should probably be moved to Mender Hub.
Though I wonder about the virtual ones - they are not (yet) present in Mender Hub..

### DIFF
--- a/03.Devices/02.Yocto-project/docs.md
+++ b/03.Devices/02.Yocto-project/docs.md
@@ -4,7 +4,7 @@ taxonomy:
     category: docs
 ---
 
-!!! Mender has four reference devices already integrated with the Yocto Project: [Raspberry Pi 3](https://www.raspberrypi.org/products/raspberry-pi-3-model-b?target=_blank), [BeagleBone Black](https://beagleboard.org/black?target=_blank) and two virtual devices (`qemux86-64` and `vexpress-qemu`). If you would like assistance supporting your device and OS, please refer to the [commercial device support offering](https://mender.io/product/board-support?target=_blank).
+!!! Check out the board integrations at [Mender Hub](https://hub.mender.io?target=_blank) to see if your board is already integrated. If you would like professional assistance supporting your board and OS, please refer to the [commercial device support offering](https://mender.io/product/board-support?target=_blank).
 
 ##Yocto Project
 Although it is possible to compile and install Mender independently, we have optimized the installation experience for those who build their Linux images using [Yocto Project](https://www.yoctoproject.org?target=_blank).
--- a/03.Devices/chapter.md
+++ b/03.Devices/chapter.md
@@ -9,4 +9,6 @@ taxonomy:
 # Board integration
 
 A description of how to integrate Mender with your board.
+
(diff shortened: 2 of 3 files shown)


mirzak:
Should be pretty straight forward to add the QEMU devices to Mender Hub and I think it would valuable to people.

I have prepared this in [meta-mender-community|https://github.com/mendersoftware/meta-mender-community/tree/sumo/meta-mender-qemu] just need some testing and a post.

03.Devices/02.Yocto-project/docs.md
```diff
@@ -4,7 +4,7 @@ taxonomy:
     category: docs
 ---
 
-!!! Mender has four reference devices already integrated with the Yocto Project: [Raspberry Pi 3](https://www.raspberrypi.org/products/raspberry-pi-3-model-b?target=_blank), [BeagleBone Black](https://beagleboard.org/black?target=_blank) and two virtual devices (`qemux86-64` and `vexpress-qemu`). If you would like assistance supporting your device and OS, please refer to the [commercial device support offering](https://mender.io/product/board-support?target=_blank).
+!!! Check out the community-supported board integrations at [Mender Hub](https://hub.mender.io?target=_blank) to see if your board is already integrated. If you would like professional assistance supporting your board and OS, please refer to the [commercial device support offering](https://mender.io/product/board-support?target=_blank).
```
kacf: This makes it sound like there are no officially supported boards at all. Is that the intention?

From looking at https://hub.mender.io/c/board-support-resources, it is not really clear to me what is officially supported and what isn't.

estenberg: @kacf the intention was to move "board support status" to hub.mender.io indeed.

@mirzak also told me that CI has not run on the Rpi3 nor BBB for a year now, and I think it is a stretch to call those officially supported if that's the case..?

However, I did not mean to say there are *no* officially supported boards, merely that their status should be found in hub.mender.io, not the docs (I don't think we should make claims about specific boards in our docs anymore now).

What about "Check out the existing board integrations and their status at Mender Hub to see if your board is already integrated..."?

kacf: > also told me that CI has not run on the Rpi3 nor BBB for a year now, and I think it is a stretch to call those officially supported if that's the case..?

They are tested manually with each release, but that's not the same, obviously. So maybe that's true. And just to be clear: We do want to keep manually testing them, right? For our downloads and tutorials?

Regarding Mender Hub I just found it peculiar that it doesn't seem to have any elevated status at all, yet we offer both tutorials and downloads for it in our docs.

estenberg: Makes sense, I left a note of the reference devices. We do have an elevated status in Mender Hub for automated testing; however we can't use it before the tests are actually automated and I am not sure we should introduce yet-another status before getting our reference boards automated..

kacf: :+1:

PR #545, opened 2018-11-05
MEN-2125: Allow trailing AUTOVERSION tags for page headers.

kacf:
If the tag is above the header, the page doesn't render correctly.


drewmoseley:
Looks ok to me although I'm not terribly familiar with the details.  But from a general docs perspective this is fine.

kacf:
Thanks!

PR #555, opened 2018-11-19
Fix instructions for wpa-supplicant on Rpi.

kacf:


drewmoseley:
LGTM. Do we need this on any older branches?

estenberg:
:+1: 

Let's get it into 1.7 at least, but would not worry about 1.6 as we are switching soon and we don't have the support there.

kacf:
Cherry-picked to 1.7.x as well.

PR #566, opened 2018-12-04
fixed in MEN-1961

estenberg:
https://tracker.mender.io/browse/MEN-1961 is now closed.

This removes the notice from master, but:
* what about 1.7 branch?
* potentially add note in 1.6 branch that this is fixed in thud/sumo...where is it fixed exactly?

@DominikAdamski @kacf please comment

### DIFF
--- a/201.Troubleshooting/02.Running-Yocto-Project-image/docs.md
+++ b/201.Troubleshooting/02.Running-Yocto-Project-image/docs.md
@@ -58,16 +58,3 @@ help here:
   #define CONFIG_AUTOBOOT_DELAY_STR "\r"
   #define CONFIG_AUTOBOOT_DELAY_STR2 "\n"
   ```
-
-
-## GRUB prints error: no such device: ((hd0,gpt1)/EFI/BOOT)/EFI/BOOT/grub.cfg.
-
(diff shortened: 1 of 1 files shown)


kacf:
I actually think I would leave it in 1.7. People are slow to update, and they could still be hitting this for quite some time.

A note about where it's fixed would be useful though. Neither thud nor sumo have the fix yet, but I plan to cherry-pick to both as the last step of the 1.7 release. So maybe you can put those two, and make a note in the release checklist for 1.7 to merge it when the time comes?

kacf:
Actually, strike sumo from that list. This doesn't cherry-pick cleanly, due to a slightly different build setup for GRUB in sumo. It's not super hard to fix, but since this is mostly cosmetic, I would vote for dropping that.

estenberg:
Thanks!

I did not see any negative effects of mentioning this now, so note added to 1.7 and 1.6:
* https://github.com/mendersoftware/mender-docs/commit/bd5c3efec91cb8a9ad672fa8836e2137b07abed0
* https://github.com/mendersoftware/mender-docs/commit/d74258830a5706c7da6536f944f768ff3275b572

PR #568, opened 2018-12-08
Updates to yocto setup

mirzak:


04.Artifacts/01.Yocto-project/01.Building/docs.md
```diff
@@ -177,6 +168,39 @@ DISTRO_FEATURES_BACKFILL_CONSIDERED = "sysvinit"
...
+#
+# Uncomment below and update the URL to match your configured domain
+# name and provide the path to the generated server.crt file.
+#
+# NOTE! It is recommend that you provide below information in your custom
+# Yocto layer and this is only for demo purposes. See linked documentation
+# for additional information.
+#MENDER_SERVER_URL = "https://docker.mender.io"
+#FILESEXTRAPATHS_prepend_pn-mender := "<DIRECTORY-CONTAINING-server.crt>:"
+#SRC_URI_append_pn-mender = " file://server.crt"
```
drewmoseley: Do we also want to comment that this is only needed for self-signed certs?

estenberg: If we can mention it in a very short/concise way I think. Most people would probably use self-signed I suppose.

mirzak: Pushed an update which mentions this. Good enough?

mirzak:
In regards to back-ports, this should go all they way back to 1.6.x?

drewmoseley:
Yeah, I think it's applicable to all supported branches.

PR #589, opened 2018-12-28
changes: there is no 1.7.0 branch, but a 1.7.x branch instead

RobertBerger:
dir:
 mender-docs

files:
 07.Administration/02.Production-installation/docs.md

### DIFF
--- a/07.Administration/02.Production-installation/docs.md
+++ b/07.Administration/02.Production-installation/docs.md
@@ -531,7 +531,7 @@ At this point your commit history should look as follows:
 <!--AUTOVERSION: "git log --oneline origin/%..HEAD"/integration -->
 <!--AUTOMATION: ignore -->
 ```bash
-git log --oneline origin/1.7.0..HEAD
+git log --oneline origin/1.7.x..HEAD
 ```
 > 7a4de3c production: final configuration
 > 41273f7 production: adding generated keys and certificates


mirzak:
A couple of quick comments about process,

- Changes typically go to master first if not very specific to a certain branch, e.g your first commit is correct to target 1.7.x. The second change should probably go trough master first and then cherry-picked to relevant stable branches.

- We require a Signed-off-by tag on all commits

- The `changes:` prefix seems superfluous to me

RobertBerger:
Please check if this is better now.

1) there should be a signed-off by 
2) extra comments were removed
3) only one the commit for the 1.7.x branch

In case this did not work we can drop it and I make a new pull request

07.Administration/02.Production-installation/docs.md
```diff
@@ -531,7 +531,7 @@ At this point your commit history should look as follows:
 <!--AUTOVERSION: "git log --oneline origin/%..HEAD"/integration -->
 <!--AUTOMATION: ignore -->
 ```bash
-git log --oneline origin/1.7.0..HEAD
+git log --oneline origin/1.7.x..HEAD
```
mirzak: After looking at this for a while, I think it would be better to change this to:

```
git log --oneline 1.7.0..HEAD
```

The reason being is that the initial clone command is:

     git clone -b 1.7.0 https://github.com/mendersoftware/integration mender-server

That is the `1.7.0` tag and not the `1.7.x` branch. And there is a chance that they differ if there are "staging" commits in `1.7.x` which have not yet been tagged. 

Also these strings are actually generated by a script, based on this tag:

    

Above probably needs an update on master, and maybe on the stable branches?

I will leave the final say to @kacf, who has better insight on how this is all connected.

mirzak:
The format looks good now, I left a comment on the change as well.

07.Administration/02.Production-installation/docs.md
RobertBerger: OK now I know what you are trying to do;)

Sorry for being so picky. Just trying to navigate somehow through the process and I am posting things which don't work as documented.

kacf: Hmm, hold on, now I realize that this will be wrong no matter what we do. If we change it not to include the `origin/` string, it is no longer correct on the master branch, where it *is* indeed a branch.

I'm actually not sure how to solve this. This may need some additional logic added to our autoversion mechanism.

In any case, I'm fine with fixing it as I suggested for this branch specifically, since this is user facing, but I will have to think about a better way once the next version comes along.

oleorhagen:
@mirzak, @RobertBerger, looks dead to me. Will close by the end of the day, if no activity occurs.

PR #590, opened 2018-12-28
maybe different path to remove meta-mender-demo

RobertBerger:
Signed-off-by: Robert Berger <robert.berger@ReliableEmbeddedSystems.com>


04.Artifacts/01.Yocto-project/03.Building-for-production/docs.md
```diff
@@ -16,6 +16,14 @@ When building for production, we need to remove the demo layer, `meta-mender-dem
 bitbake-layers remove-layer ../meta-mender/meta-mender-demo
 ```
 
+!!! Note that in case the above does not work you might want to check the inluded layers with
```
drewmoseley: Typo here: "included"

RobertBerger: should be fixed now

PR #592, opened 2019-01-02
add polling intervals to "build for production section"

mirzak:
It makes sense to me to mention it in the our "building for production" section.

### DIFF
--- a/04.Artifacts/01.Yocto-project/03.Building-for-production/docs.md
+++ b/04.Artifacts/01.Yocto-project/03.Building-for-production/docs.md
@@ -25,6 +25,12 @@ and then remove the demo layer e.g.
 bitbake-layers remove-layer meta-mender-demo
 ```
 
+## Configure polling intervals
+
+For security reasons, the Mender client does not require any open ports at the embedded device. Therefore, all communication between the Mender client and the server is always initiated by the client and it is important to configure the client so that the frequency of sending various requests to the server is reasonable for a given setup.
+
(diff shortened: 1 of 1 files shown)


mirzak:
I am not 100 % sure if I got the relative link correct, how do you normally test them? Github does not seem to resolve them?

04.Artifacts/01.Yocto-project/03.Building-for-production/docs.md
```diff
@@ -25,6 +25,12 @@ and then remove the demo layer e.g.
 bitbake-layers remove-layer meta-mender-demo
 ```
 
+## Configure polling intervals
+
+For security reasons, the Mender client does not require any open ports at the embedded device. Therefore, all communication between the Mender client and the server is always initiated by the client and it is important to configure the client so that the frequency of sending various requests to the server is reasonable for a given setup.
+
+Please refer to [client configuration - polling intervals](../../../client-configuration/configuration-file/polling-intervals), for information on how to choose and how to set polling intervals.
```
estenberg: Could drop the "client configuration" part, just use "polling intervals" as link text.

mirzak: Done

mirzak:
> To test links you have to set up the docs site locally: https://github.com/mendersoftware/mender-docs-site/ - though this looks right to me.

Tested the link by setting up above :+1:

mirzak:
@drewmoseley @eysteins , should this be back-ported as well?

drewmoseley:
I would say if it cherry picks cleanly then sure

mirzak:
It sure does, created #594 and #595

PR #596, opened 2019-01-18
 Update to "Partition configuration"

ajithpv:
This PR has the following changes:
1) The folder name changed from "Partition layout" to "Partition configuration" for both Yocto and Debian
2) The Title name changed from "Partition layout" to "Partition configuration" for both Yocto and Debian
3) The above changes will provide better readability and more specific to the contents.

### DIFF
--- a/03.Devices/02.Yocto-project/01.Partition-configuration/docs.md
+++ b/03.Devices/02.Yocto-project/01.Partition-configuration/docs.md
@@ -1,5 +1,5 @@
 ---
-title: Partition layout
+title: Partition configuration
 taxonomy:
     category: docs
 ---
--- a/03.Devices/03.Debian-family/01.Partition-configuration/docs.md
+++ b/03.Devices/03.Debian-family/01.Partition-configuration/docs.md
@@ -1,5 +1,5 @@
 ---
-title: Partition layout
+title: Partition configuration
 taxonomy:
     category: docs
 ---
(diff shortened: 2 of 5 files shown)


mirzak:
Looks good to me, though it is missing updating all references to the old file names.

If you do a:

    grep -r "partition-layout" ./*

In the mender-docs source folder you will see a lot sections linking to these `..partition-layout...`, e.g

   [Flash memory types](../../devices/yocto-project/partition-layout#flash-memory-types)

These need to be updated as well, otherwise the reference links will be broken.

ajithpv:
OOPS! I forgot to change the linking sections. I will update all references to the new file name(s).

ajithpv:
@drewmoseley and @estenberg , I have updated the references. Please have a look

drewmoseley:
Looks ok.  Any reason not to squash all into a single commit?

estenberg:
Looks good, agree a squash would make sense.

We should run a dead link checker on development docs after it is merged to master.

ajithpv:
@drewmoseley I was not able to do it from my PC due to some reason. Hence, I done it via Github directly. Even I agree a squash would be better. Should I drop this PR and create a new one with squash all into a single commit?

ajithpv:
@mirzak , @drewmoseley and @estenberg : I have squashed all commits into a single one. Please let me know whether this is fine or anything need to be modified.

PR #598, opened 2019-01-28
MEN-2336: mender-artifact modify only works if payload is an ext4 image

oleorhagen:


oleorhagen:
@mirzak This should cover it. A few of them are located in the same section.

PR #603, opened 2019-02-12
artifacts: do not recommend to store the artifact verify key on the d…

mirzak:
…ata part

The documentation also stated that we do this by default, which we do not and
the key is stored in '/etc/mender/' by default which is also the recommended
location to be able to rotate it as part of an rootfs update.

Changelog: None

### DIFF
--- a/04.Artifacts/01.Yocto-project/03.Building-for-production/docs.md
+++ b/04.Artifacts/01.Yocto-project/03.Building-for-production/docs.md
@@ -86,8 +86,6 @@ The private key used for signing the Mender Artifact should be protected and kep
 thus there are no extra steps needed to add it to any part of the build system, Mender Client nor Server.
 
 Only the public key, which is used by the Mender Client to verify the signed Artifact must be included in the Mender Client build.
-The public verification key should be stored on *persistent storage* on the device where the Mender client runs,
-as the key should not change across deployments (except when doing key rotation). By default it is stored on the data partition.
 
 The best way to include a public verification key in the client is to add it to your own layer. Set the name of the verification key to `artifact-verify-key.pem` and append it to `SRC_URI` of the `mender` application before building the Yocto client image. For example:
 


mirzak:
This PR is based on discussions from IRC, transcript below:

```
<Alex___> rrcs-24-30-232-222.nyc.biz.rr.com/24.30.232.222 I'm working with signed keys via mender. I noticed that via the documentation here: https://docs.mender.io/1.6/artifacts/building-for-production#artifact-signing-and-verification-keys , that the verification key should be on persistent storage. However my image seems to be deploying it to /etc/mender/artifact-verify-key.pem
17:37 Is there a configuration setting that I am missing to force the verification key onto the /data/ partition?
17:38 <aduskett> That will be awesome!
17:40 <mirzak> Mirza Krak Indeed the code is not behaving as the documentation, 
17:40 It even says ""By default it is stored on the data partition."
17:40 It is not :P
17:41 And no there no variable to force it, you could create a mender_%.bbappend and adding a do_install_append step to move it to the data part
17:45 <Alex___> rrcs-24-30-232-222.nyc.biz.rr.com/24.30.232.222 I wondered if that was my only option. I would also need to update /etc/mender/mender.conf to point to the new location as well?
17:45 <mirzak> Mirza Krak Yepp
17:45 Or symlink /etc/mender/artifact-verify-key.pem -> /data/artifact-verify-key.pem
17:48 <Alex___> rrcs-24-30-232-222.nyc.biz.rr.com/24.30.232.222 The main difference I see between putting it in the rootfs vs /data is that in the rootfs you can use a mender rootfs to do a key rotation. (which is convenient if you don't lose the original key).  However if the key is on the data partition you will need a secondary mechanism to do key rotation. (Note we are using mender in standalone mode).
17:53 Also from a quick test it looks like you can do mender -rootfs from an uncommitted partition. Is this an intended feature? It would seem you would only want to be able to do a rootfs upgrade if the current partition is active.
18:04 <kacf1> kacf mirzak: Yeah, the docs are wrong in this respect. We are definitely (and should be) storing it on the rootfs, so that it can be rotated like everything else in the rootfs image
```

PR #608, opened 2019-02-18
MEN-2074: Document Update Modules

lluiscampos:
First draft for the documentation I am preparing for Update Modules. I added TODOs in places where IU am unsure about what should be done.

Your feedback is appreciated.


04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,210 @@
...
+
+```bash
+./mender-artifact --version
+```
+
+It should be version X.Y.Z or newer
+
+!!TO-DO: mender-artifact version?
+
+Future releases will allow for "artifact creation" for Update Modules from the web, so that a build machine will not be required anymore. But for now this is the only way to prepare the Update Modules' payloads.
```
kacf: I would leave this out for now. We're not exactly sure what we are going to offer, or how, so we can rather change this when we have something solid.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Basic example: File copy Update Module
+
+In this first basic example we will create an Update Module that copies files into `/data` directory on the device.
```
kacf: As mentioned elsewhere, I think this is too specific to our dual rootfs offering. I suggest choosing a different directory, such as `/var/www` (a common place for a device to serve html files from). This will require changes in many places, obviously.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,210 @@
...
+    SupportsRollback)
+        echo "Yes"
+        ;;
+
+    ArtifactInstall)
+        find /data -maxdepth 1 -type f  -exec tar -rf $prev_files_tar {} \;
+        find /data -maxdepth 1 -type f | xargs rm
+        expected_md5sums=$(md5sum "$FILES"/files/* | cut -d' ' -f1)
+        cp "$FILES"/files/* /data
+        actual_md5sums=$(find /data -maxdepth 1 -type f | xargs md5sum | cut -d' ' -f1)
```
kacf: I think this is susceptible to sorting order (`ls` sorts, but `find` does not). Maybe insert a `| sort` after the `cut` in both places?

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,210 @@
...
+        tar -xf $prev_files_tar -C /data
+        ;;
+
+esac
+exit 0
+```
+
+This new version of `data-file` Update Module implements a simple mechanism to restore the previous state (i.e. files in /data before the update).
+
+What is doing, is saving a tarball with all current files in `/data` (only the single files, not directories) before removing them and copying the new set of files coming from the archive. After copying the files, it will check that the md5sum of all copied files is the same as the files contained in the incoming Artifact. If they differ, the script will exit with error code `1` for this `ArtifactInstall` phase. Finally, the mender client will interpret this error code as a failure, check if our Update Module supports rollback, and then call again the module with `ArtifactRollback`option, in which the module will restore the old files.
```
kacf: As an example, I suggest one paragraph where the user can experience what happens if a rollback is triggered. Creating a directory in `/data` which conflicts with one of the files is one way, for example.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,210 @@
...
+        [[ -f $prev_containers_file ]] || exit 1
+        find /data -maxdepth 1 -type f | xargs rm
+        tar -xf $prev_files_tar -C /data
+        ;;
+
+esac
+exit 0
+```
+
+This new version of `data-file` Update Module implements a simple mechanism to restore the previous state (i.e. files in /data before the update).
```
kacf: Missing monospace quotes around "/data".

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,210 @@
...
+* **deb** Simple update module to install all `.deb` packages sent in the artifact using `apt-get`.
+* **rpm** Similar to the previous one for `.rpm` packages.
+* **docker** This module allows to setup a given combination of docker images in the target
+
+!! TO-DO: Add links here
+
+## Further reading
+
+* [Update modules v3 protocol](https://github.com/mendersoftware/mender/blob/master/Documentation/update-modules-v3-file-api.md)
+* [Mender Artifact file format v3](https://github.com/mendersoftware/mender-artifact/blob/master/Documentation/artifact-format-v3.md)
```
kacf: I think we can omit the format v3 link. It's only needed if you get deep into the update module protocol, and there already is a link inside that document if you need it.

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+#### Writable directory for mender
+
+You need to set a writable directory for mender at `/var/lib/mender`
+
+For example:
+
+```bash
+sudo ln -s /data/mender /var/lib/mender
```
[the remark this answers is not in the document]
lluiscampos: You are right. I agree it is important to separate (as much as possible) the use cases of rootfs updates and Update Module possibilities.

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,95 @@
...
+
+Then install in the identity tools, see [../client-configuration/identity] for more information.
+
+```bash
+install -d /var/lib/mender/identity
+install -t /var/lib/mender/identity -m 0755 \
+            ${MENDER_SRC_PATH}/support/mender-device-identity
+```
+
+!! TO-DO: Use MENDER_SRC_PATH (and explain it) or rather wget from GitHub?
```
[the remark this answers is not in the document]
lluiscampos: My doubt here is that if we provide a prebuilt binary, then the reader will not have a cloned repo. So I don't know if the doc should require a full clone or not.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,210 @@
...
+
+You will need a device with mender client v2.0 or newer, which supports Update Modules.
+
+There are three ways to obtain this:
+
+* You can build a new rootfs image for your device using [Yocto build](../yocto-project)
+* You can use a prebuilt client for your architecture and [install the client manually](../../client-configuration/installation).
+* You can [cross-compile](../../client-configuration/cross-compiling) the client from a build machine and [install the client manually](../../client-configuration/installation).
+
+!!TO-DO: prebuilt client?
```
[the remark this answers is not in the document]
lluiscampos: Ok. I will not remove just yet to let @estenberg comment

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,210 @@
...
+
+To verify that `mender-artifact` supports Update Modules, execute:
+
+```bash
+./mender-artifact --version
+```
+
+It should be version X.Y.Z or newer
+
+!!TO-DO: mender-artifact version?
```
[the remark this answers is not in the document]
lluiscampos: That is my suggestion also, to hardcode 2.0 here

04.Artifacts/08.Update-Modules/docs.md
lluiscampos: Right, I agree in taking it out. It doesn't add any value at the moment anyway.

04.Artifacts/08.Update-Modules/docs.md
lluiscampos: Good idea, I think a web server files copy is a good use case. I am changing it.

04.Artifacts/08.Update-Modules/docs.md
lluiscampos: That is a good example, thanks.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,208 @@
...
+<!--AUTOVERSION: "version % or newer"/ignore-->
+It should be version 3.0.0 or newer.
+
+## Basic example: File copy Update Module
+
+In this first basic example we will create an Update Module that copies files into `/var/www` directory on the device.
+
+### Create the Update Module script
+
+In the device terminal, go to `/usr/share/mender/modules`
```
lluiscampos: @kacf Should this be `/usr/lib/mender/modules/v3`?

It certainly works without the v3, but the Documentation of the protocol states we need the v3 subdir...

lluiscampos: ref https://github.com/mendersoftware/mender/pull/330/files?short_path=1122e6d

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
...
+
+esac
+exit 0
+```
+
+This new version of `data-file` Update Module implements a simple mechanism to restore the previous state (i.e. files in `/var/www` before the update).
+
+What is doing, is saving a tarball with all current files in `/var/www` (only the single files, not directories) before removing them and copying the new set of files coming from the archive. After copying the files, it will check that the md5sum of all copied files is the same as the files contained in the incoming Artifact. If they differ, the script will exit with error code `1` for this `ArtifactInstall` phase. Finally, the mender client will interpret this error code as a failure, check if our Update Module supports rollback, and then call again the module with `ArtifactRollback`option, in which the module will restore the old files.
+
+!!! You can experiment with the rollback mechanism by forcing the update to fail. For example, creating a directory under `/var/www` with the same name as one of the files contained in the incoming artifact will trigger an error in the `cp` command. Once this happens, the Update Module should be able to restore the previous files.
```
estenberg: Artifact

estenberg: "should be able to restore" -> "should restore"

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
...
+
+## More Update Modules
```
estenberg: Since we refer to Mender Hub and mention some of them in the beginning I think we can drop this header & section. My main concern with calling them out explicitly is that this might get outdated if we remove some or add more out of the box.

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,97 @@
+---
+title: Installing the mender client
```
estenberg: Mender

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,97 @@
...
+taxonomy:
+    category: docs
+---
+
+If you use system updates (builtin rootfs type) you must follow [board integration tutorials](../../devices), but if you are just relying on [Update Modules](../../artifacts/update-modules) you can follow this tutorial for setting up the client manually on the target system.
```
estenberg: "If you use" -> "If you would like support for"

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,97 @@
...
+
```
estenberg: Would add:
"
## Prerequisites

#### Mender client for your device

You need a Mender client binary for your device. Follow the steps in [Cross compiling](../cross-compiling) to obtain one.
"
Here is also where we could link to any prebuilt binaries, once we have them.

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,97 @@
...
+## Manually installing and configuring mender client on device
```
estenberg: A bit long header, what about "Install and configure the Mender client"

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+#### Copy mender binary
+
+Copy the prebuilt or cross-compiled binary to directory `/usr/share/mender`. From the directory where you have downloaded/built the binary:
```
estenberg: "Copy the Mender client binary to a working directory on your device, e.g. using `scp`:

```bash
scp mender root@<IP-OF-DEVICE> /root
```

Then open a shell on your device (e.g. using SSH) and change to the working directory:

```bash
ssh root@<IP-OF-DEVICE>
cd /root
```

From your working directory, run the following commands to install the Mender binary:"

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+You need to set a writable directory for mender at `/var/lib/mender`
+
+For example:
+
+```bash
+install -d /var/lib/mender
+```
+
+! If you are planning to do rootfs updates, `/var/lib/mender` will then need to be in a separate partition. For this case, we recommend setting `/var/lib/mender` as a symbolic link to a directory in such partition.
```
estenberg: I think you can skip this line as this tutorial is not for people doing system/rootfs updates, this is noted at the top.

05.Client-configuration/06.Installation/docs.md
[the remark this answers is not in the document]
estenberg: Hmm, this needs to be fixed somehow indeed.

Maybe we should just bite the bullet and make this into a small shell script that will fetch from the github repo, given a Mender client version (2.0.0 or newer)? Then it can be extended later to also obtain a prebuilt client.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
...
+title: Getting started with Update Modules
+taxonomy:
+    category: docs
+---
+
+## Introduction
+
+Starting from release 2.0, mender now supports Update Modules feature. Common update modules for application updates are provided in the release, such as support for file-based, package based and container-based updates. But in this section we will guide you into how to write your own Update Module for your custom use case.
+
+An update module implements one or more of the actions that the Mender client carries out when installing an Artifact payload. The *core* action all update modules must implement is the Install action, which implements how an Artifact payload is installed on the device. However, there are other actions that may be implemented depending on the desired functionality and use case for an update module, such as Rollback.
```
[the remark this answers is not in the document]
lluiscampos: I missed this one, thanks. I've been follow such convention.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
+---
+title: Getting started with Update Modules
+taxonomy:
+    category: docs
+---
+
+## Introduction
+
+Starting from release 2.0, mender now supports Update Modules feature. Common update modules for application updates are provided in the release, such as support for file-based, package based and container-based updates. But in this section we will guide you into how to write your own Update Module for your custom use case.
```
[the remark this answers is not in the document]
lluiscampos: Yes, good idea.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
...
+### The state machine workflow
+
+Update modules are modelled after the same execution flow as [state scripts](../state-scripts). For the development of Update Modules it is important to have a basic understanding of the workflow.
+
+![Update Modules state machine](update-modules-state-machine.png)
+
+The most relevant states for developing an Update Module are:
+
+* `ArtifactInstall` state should be used to install the update into its final destination.
+* `ArtifactCommit` state should be used to make the update permanent, in cases where it's still possible to roll back.
```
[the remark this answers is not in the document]
lluiscampos: An Update Module can implement `ArtifactCommit`even though it does not support rollback. It is still optional, for sure, but it could be used for other purposes.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
...
+
+### The state machine workflow
+
+Update modules are modelled after the same execution flow as [state scripts](../state-scripts). For the development of Update Modules it is important to have a basic understanding of the workflow.
+
+![Update Modules state machine](update-modules-state-machine.png)
+
+The most relevant states for developing an Update Module are:
+
+* `ArtifactInstall` state should be used to install the update into its final destination.
```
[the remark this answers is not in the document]
lluiscampos: Similarly, I think there is no really "required" step. I cannot find a use case where `ArtifactInstall` would not be implemented, but as per design there are no required steps.

@kacf please correct me if I am wrong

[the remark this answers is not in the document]
lluiscampos: Regarding the three suggestions to append "optional" or "mandatory" in these states, I think it is better to omit them altogether, as they might mislead the user.

If any, I can add a paragraph before/after saying that all the states are actually optional and it will be application dependent to implement them, enforcing the idea than the main action will most likely happen in `ArtifactInstall`

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
...
+
+There are three ways to obtain this:
+
+* You can build a new rootfs image for your device using [Yocto build](../yocto-project)
+* You can use a prebuilt client for your architecture and [install the client manually](../../client-configuration/installation).
+* You can [cross-compile](../../client-configuration/cross-compiling) the client from a build machine and [install the client manually](../../client-configuration/installation).
+
+!!TO-DO: prebuilt client?
+
+### Build machine with mender-artifact
```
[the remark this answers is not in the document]
lluiscampos: Yeah, it is a better term.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
...
+chmod +x data-file
+```
+
+Your Mender client is now will be able to handle any artifact containing a `data-file` Update Module payload.
+
+### Create an Artifact with a payload for the new update module
+
+In your build machine, change directory to where you downloaded (or built) mender-artifact above.
+
+First, simply create a file you want to install to the data partition:
```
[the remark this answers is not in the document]
lluiscampos: Good catch, thanks.

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
...
+
+```bash
+echo 'Installed by Mender!' > hello-world
+```
+
+Now we can create a new Mender Artifact with this as the only payload, for our new `data-file` Update Module.
+
+```
+./mender-artifact write module-image -t qemux86-64  -o ~/Downloads/data-file-1.mender -T data-file -n data-file-1.0 -f hello-world
```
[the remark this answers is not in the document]
lluiscampos: Good idea. Thank you.

04.Artifacts/08.Update-Modules/docs.md
lluiscampos: Yes, I think it is a good idea to remove the section once we are referring to Mender Hub in the introduction.

05.Client-configuration/06.Installation/docs.md
lluiscampos: For the sake of simplification, I agree. And if we do the script we cannot take this into account anyway.

05.Client-configuration/06.Installation/docs.md
lluiscampos: Yes, better not to omit that the binary goes from workstation to device via SSH. Thanks

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Create the Update Module script
+
+In the device terminal, go to `/usr/share/mender/modules`
+
+```bash
+cd /usr/share/mender/modules
+```
+
+and create an script named `data-file` with the following content:
```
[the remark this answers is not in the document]
lluiscampos: `web-file` is a good suggestion, thanks.

We will provide a UM for generic files, so I think is a good idea to make a bigger differentiation.

04.Artifacts/08.Update-Modules/docs.md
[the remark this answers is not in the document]
lluiscampos: Fixed as suggested.

04.Artifacts/08.Update-Modules/docs.md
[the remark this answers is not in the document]
lluiscampos: I take your proposal almost as-is :smiley:

05.Client-configuration/06.Installation/docs.md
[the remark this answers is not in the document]
lluiscampos: My new suggestion here is to automate the install, but not as a separate script but just as an install target in mender repo `Makefile`. This will ease [MEN-2367](https://tracker.mender.io/browse/MEN-2367) but also have consequences for `meta-mender` repo.

Then this tutorial would still keep all the steps (replacing the clone with wget from GitHub). Adding also the indication at the start that all the steps can be automatically run with defaults values with "make install" (so that the tutorial is for customization)

lluiscampos:
Thank you both for the thorough feedback. I applied most of the suggestions that you made, marking them with :+1: emoji.

I just updated the PR with all these changes

Then, the few open discussions remaining are:

1. clone repo, wget scripts, or new script for installation [link](https://github.com/mendersoftware/mender-docs/pull/608#discussion_r257960830)
2. Mention to required/optional states for Update Modules [link](https://github.com/mendersoftware/mender-docs/pull/608#discussion_r258272109)
3. Possible inconsistency between doc and implementation of UM v3 protocol [link](https://github.com/mendersoftware/mender-docs/pull/608#discussion_r258078108)

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,208 @@
...
+        find /var/www -maxdepth 1 -type f  -exec tar -rf $prev_files_tar {} \;
+        find /var/www -maxdepth 1 -type f | xargs rm
+        expected_md5sums=$(md5sum "$FILES"/files/* | sort | cut -d' ' -f1)
+        cp "$FILES"/files/* /var/www
+        actual_md5sums=$(find /var/www -maxdepth 1 -type f | xargs md5sum | sort | cut -d' ' -f1)
+        [[ "$actual_md5sums" == "$expected_md5sums" ]] || exit 1
+        ;;
+
+    ArtifactRollback)
+        [[ -f $prev_containers_file ]] || exit 1
```
lluiscampos: Typo: `prev_files_tar`

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,114 @@
...
+
+```bash
+{
+    "InventoryPollIntervalSeconds": 60,
+    "RetryPollIntervalSeconds": 30,
+    "UpdatePollIntervalSeconds": 5,
+    "ServerURL": "https://mymenderserver.net",
+    "ServerCertificate": "/etc/site-conf/server.crt"
+}
+```
```
lluiscampos: Convert into heredoc this configuration file also

05.Client-configuration/06.Installation/docs.md
```diff
@@ -0,0 +1,114 @@
...
+For example:
+
+```bash
+install -d /var/lib/mender
+```
+
+Then install the identity tools, see [../client-configuration/identity] for more information.
+
+```bash
+install -d /var/lib/mender/identity
```
lluiscampos: This is wrong. Here and below (inventory) should be `/usr/share/mender/...` instead of `/var/lib`

04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -0,0 +1,208 @@
+---
+title: Getting started with Update Modules
```
estenberg: I missed this, sorry, but I would just call it "Update Modules" (same as the name in the navigation to the left).

04.Artifacts/08.Update-Modules/docs.md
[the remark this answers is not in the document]
estenberg: Sounds good, you know this better than me. :)

Would add a similar comment that you suggest above, maybe even saying that something like "Every state is optional for an Update Module to implement, however in practice all Update Modules would likely start with implementing the ArtifactInstall state."

05.Client-configuration/06.Installation/docs.md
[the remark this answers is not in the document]
estenberg: Interesting idea, I just still think it would be easier on the user to have this fully scripted, instead of
* cross-compile client
* scp client
* SSH to device
* git clone ...
* cd ...
* Make
* update config (e.g. server url, tenant token)
* start Mender service

I am pressing on this because this will be the default path for people to evaluate Mender as well (e.g. when signing up for Hosted Mender). So it is really important there are as few steps as possible, and some kind of script/tool to do all this would seem like an obvious candidate.

We can even make this a script that you can run on your laptop on a remote device, which probably would be the easiest. Perhaps you have other ideas?

That said, we can make it easier in iterations, and if you think a Makefile can be re-used and we can simply add some tooling on top of it to install Mender then we can certainly start with that.

05.Client-configuration/06.Installation/docs.md
estenberg: Just one thing to note on the configuration itself. We have this hard-coded already in:
* meta-mender
* mender-convert
* now here

In addition, there is "demo mode" settings and "production" settings (mainly affecting polling intervals). So we could end up with potentially 6 different configurations, when only 2 would be needed.

Not sure when we should unify this, but perhaps now is the time? Perhaps it relates to the scripting/makefile idea (can be reused across meta-mender, mender-convert, and here)?

05.Client-configuration/06.Installation/docs.md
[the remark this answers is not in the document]
lluiscampos: I have been going back and forth with this today and I still don't have the solution.

The install of the Mender client is already implemented in `meta-mender` and `mender-convert`. I believe that creating yet another script/tool (even as a Makefile target) is not a good idea anymore, as it will make it hard to keep everything in sync (same problem as pointed out with the `mender.conf`)

My idea now is to refactor `mender-convert` shell scripts a bit so that it can be used to install mender client only in addition to the current use (to convert the rootfs **and** to install mender into it). Then this new use of `mender-convert` can be documented in this doc page (i.e. remove all manual steps from here).

I have implemented a prototype for this solution and seems feasible. What do you think about it?

Sorry for the confusion going back and forth, I am still trying to get the full overview of our ecosystem.

05.Client-configuration/06.Installation/docs.md
lluiscampos: I agree in removing this (and basically all install steps) to not duplicate info. Refer to the other discussion [here](https://github.com/mendersoftware/mender-docs/pull/608#discussion_r259333728).

05.Client-configuration/06.Installation/docs.md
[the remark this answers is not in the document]
mirzak: > My new suggestion here is to automate the install, but not as a separate script but just as an install target in mender repo `Makefile`. This will ease [MEN-2367](https://tracker.mender.io/browse/MEN-2367) but also have consequences for `meta-mender` repo.
> Then this tutorial would still keep all the steps (replacing the clone with wget from GitHub). Adding also the indication at the start that all the steps can be automatically run with defaults values with "make install" (so that the tutorial is for customization)

Sorry for going another cycle on this and commenting late in the process.

I believe that it is indeed time to implement a `make install` step in [mender client repo](https://github.com/mendersoftware/mender). This might have not made sense earlier as there was no point of installing Mender without the rest of the integration provided by build systems as is done in [meta-mender](https://github.com/mendersoftware/meta-mender) but with update modules this changes in my opinion (even though I think that made sense earlier as well :)).

I would also like to extend our list of duplication's:
- [meta-mender](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender/mender.inc)
- [Buildroot](https://github.com/buildroot/buildroot/blob/dd97b055b11b8708438974555d1746a18924ceba/package/mender/mender.mk)
- [mender-convert](https://github.com/mendersoftware/mender-convert/blob/master/convert-stage-4.sh)
- [Blog post](https://mender.io/blog/porting-mender-to-a-non-yocto-build-system)
    - which we said we would revise and update
- [Mender client debian package](https://packages.debian.org/sid/main/mender-client)
- OpenWRT conversion scripts (not available publicly yet)

So it is **really** time to try to unify these steps in the [mender repo](https://github.com/mendersoftware/mender), meaning that anything that is not build system specific should be part of the `make install`, another way of saying that is anything that is shared between all the duplication's should be handled by `make install`. We have the information necessary now that the Mender client is installed on so many different platforms.

Example what it would look like:
```
go get github.com/mendersoftware/mender
cd $GOPATH/src/github.com/mendersoftware/mender
make
sudo make install
```
Result of above should be a running Mender client that is trying to connect to a server. Then `make install` can be shared by all the implementations.

Above can be put in a script if you want have more automation. e.g setting up go if not present or download a pre-built client or input demo server ip address or Hosted Mender token and so on. These are typically things that are handled trough the different build systems that are supported.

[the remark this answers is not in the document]
estenberg: Makes a lot of sense. How would this work for installing on a target device, though?
It is rare that you would compile the Mender binary on the target itself, so we'd have to support:

* cross-compile (or fetch prebuilt)
* scp to target
* make install (?)

But the target doesn't necessarily have Make or built tools, right (so it can't run make install)?

Maybe make install-remote can be supported?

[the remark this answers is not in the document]
estenberg: I am assuming an already-running target here, with no package manager support. I think this will be the common case for evaluating Mender, at least in the beginning (until Mender is "downstreamed" to most OSes). A different approach to solve (most cases) is to simply package Mender ourselves soon.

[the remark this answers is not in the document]
mirzak: Oki, if we are assuming that we have a device that basically only has `busybox` installed and a internet connection then we have to resort to tricks :).

```
wget https://github.com/mendersoftware/mender/archive/1.7.0.tar.gz
tar xvf 1.7.0.tar.gz -C /tmp/
cd /tmp/mender-1.7.0
./make.sh
./make-install.sh
```

`make.sh` and `make-install.sh` are "pure shell" and not `bash` for maximum portability.

The ordinary `make` and `make install` are pass-trough to the the `.sh` equivalents, for the use-case where `make` is present.

`make.sh`, needs to figure out if it is possible to compile, if not fetch pre-built binary for the running architecture. If unsupported apologize and exit :D.

`make-install.sh`, installs either compiled mender binary or the downloaded pre-built one, along with all the necessary files to get an operating Mender client.

 > A different approach to solve (most cases) is to simply package Mender ourselves soon.

This would still require a package manager running on target then?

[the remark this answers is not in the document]
estenberg: Nice ideas!

What I was thinking in the no package manager case, is mainly for development boards and initial evaluation of Mender. So if someone today comes along with a Raspbian, how do they install Mender? In this case the device has a *package manager* but not a *package* (yet). :)

Maybe I am trying to conflate too many things into one solution. It seems we should separate the *build* from the *install* step.

* *build* (e.g. make): Will in most cases be cross-compilation (we can support native building but would not have this as a priority initially). Output: Mender binary and needed configuration available.

* *install* (e.g. make install): Assumes you are in build directory, installs to the current system. Does not seem very useful, as it is rare to do native builds. Output: Mender & configuration files installed from build directory.

* *package tar* (e.g. make package-tar): Similar to make install but instead of installing on running system it creates a tar.gz that can be unpacked to / on the target device.

* *package deb* (e.g. make package-deb): Same as package tar but different package format.

In addition to these we can add a script that would install Mender to a running device over SSH. E.g.:
* ./mender-install 10.0.5.177
** fetches mender from github
** (future: SSH into device to get right make options)
** runs make 
** runs make package-tar
** scp mender.tar.gz to 10.0.5.177 and unpacks it to /
** asks about server config (e.g. Hosted Mender tenant token)
** starts Mender

Also, once we get prebuilt binaries (e.g. tar files or deb packages, the mender-install script can rather use those than the make command).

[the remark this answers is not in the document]
mirzak: Just wanted to clarify that all my "ideas" are based on running on a device, and not involving a host machine at all. This does rely on the fact that we would provide pre-built binaries of the Mender client.

I also would really like us to avoid inventing stuff that is already solved. In the UNIX/Linux world it is "fairly standard" how you compile and install something.

Slightly simplified as there might be a configure step depending if you use a Makefile generator (e.g cmake or autotools), but after running the generators you do:

    make
    make install

Done! We already support these steps today in the Mender client, though our install steps does not actually install anything (it runs go install)

So IMO step 1, make sure `make install` installs a working Mender configuration, working can mean that the Mender client starts only, and configuration is necessary to connect to a server.

From step 1, you can build specialized solutions like you mention below, `make package-tar etc...`, so basically a bottom up approach. Trying to "patch it together" with scripts without having this fundamental functions is hacking away :).

> * _build_ (e.g. make): Will in most cases be cross-compilation (we can support native building but would not have this as a priority initially). Output: Mender binary and needed configuration available.

We already support this, both native and cross-compilation.

Note, `make` should only compile the binary and nothing else. Meaning Output: Mender binary!

> * _install_ (e.g. make install): Assumes you are in build directory, installs to the current system. Does not seem very useful, as it is rare to do native builds. Output: Mender & configuration files installed from build directory.

One note, does not assume to be installing on the current running system but relies on a prefix which might default to `/`. But could point to local directory and probably a prerequisites to `mender package-tar` and `mender package-deb`

Third note, this is the most important step to implement imo. This can be re-used by everything, Yocto, Buildroot, mender-convert, on-boarding scripts etc...

> * _configure_ (e.g. make configure): Select demo vs. non-demo templates. Can input server config, tenant token, etc.

This to me is a post-install step, meaning that `make install` installs a utility script (e.g mender-conf), which you can call to configure Mender easily with some pre-configured flags, or asking the user for input. This way you can also re-configure easily.

It can also be required to call `mender-conf` at least once to enable the systemd service and generate a custom configuration file.

`mender-conf` represents the lack of build system, e.g Yocto, Buildroot, deb package (even if it could be used together with deb package).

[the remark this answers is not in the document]
lluiscampos: Thanks for your valuable discussion.

I will try to summarize the state and to suggest the way forward.

At this point, we have this task (and PR) to **document** Update Modules. However, along the way we have seen a need to document/automate the mender client install for devices not being built through yocto or `mender-convert`. I think we all agree in that automating the install step is the way to go and that we want to have it in the `Makefile` of the main `mender` repo.

But what it doesn't seem completely clear is how to design this automation; specially how to support the two build/install scenarios (build and install natively, and build and install remotely), and to figure out how to configure (and re-configure?) step.

In any case I hope we also agree that this is an important scope creep. So at this point my suggestion is to focus the current PR in the documentation and work on a new task to automate the client install step, where we can gather all input and agree on a design for the process. This new task is independent of Update Modules and an actual dependency to do [MEN-2367](https://tracker.mender.io/browse/MEN-2367).

For the current PR to proceed, I am removing the `Installation` docs page completely and rework `Prerequisites > Client` from `Update modules` doc page to refer only to the current yocto and debian options (temporary until we have the automation script in place).

Does it sound like a plan?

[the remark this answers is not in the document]
lluiscampos: Created [MEN-2383](https://tracker.mender.io/browse/MEN-2383)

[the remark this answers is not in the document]
estenberg: Sounds good & agreed!

This is ready to merge from my PoV, and let's follow up with discussions for installation. I also agree we should have standard make / make install behaviour, but need something more than this as well to make it work in real scenarios. But let's save this for a meeting, then adjust the followup tasks.

04.Artifacts/08.Update-Modules/docs.md
kacf: Yes, but it should be `/usr/share/mender/modules/v3`, not with `lib`, but with `share`. This is wrong in the spec file right now, but I have fixed it locally.

05.Client-configuration/06.Installation/docs.md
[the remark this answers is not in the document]
kacf: I second pretty much everything that @mirzak said! :-)

>> * _install_ (e.g. make install): Assumes you are in build directory, installs to the current system. Does not seem very useful, as it is rare to do native builds. Output: Mender & configuration files installed from build directory.
> 
> One note, does not assume to be installing on the current running system but relies on a prefix which might default to `/`. But could point to local directory and probably a prerequisites to `mender package-tar` and `mender package-deb`

Yes, it is also fairly standard on UNIX (at least anything using automake) to use the `prefix` variable to set a prefix that all other installation folders are relative to, like this:

```
make prefix=$HOME/deb-pkg-folder install
```

And then files end up in folders like `$HOME/deb-pkg-folder/usr/bin/mender` and such, and this can be used to build deb packages, or to transfer an entire tree to a device.

04.Artifacts/08.Update-Modules/docs.md
kacf: And also wrong in the implementation btw, so it's good we caught this! :-)

lluiscampos: :+1:

lluiscampos:
@kacf Updated the doc with the `v3` :heavy_check_mark:

PR #616, opened 2019-03-08
Remove install of custom toolchain on debian page

mirzak:
For additional context regarding gcc6 see https://github.com/mendersoftware/mender-convert/pull/78

### DIFF
--- a/04.Artifacts/02.Debian-family/docs.md
+++ b/04.Artifacts/02.Debian-family/docs.md
@@ -95,7 +95,7 @@ Then adjust to the correct paths below and run the conversion:
             --mender-client <PATH-TO-MENDER-CLIENT-BINARY>  \
             --artifact-name golden-image-1                  \
             --bootloader-toolchain arm-linux-gnueabihf      \
-            --demo-host-ip <IP-OF-DEMO-SERVER> --keep
+            --demo-host-ip <IP-OF-DEMO-SERVER> --demo
 ```
 
 !!! The conversion may take 10 minutes, depending on the resources available on your machine.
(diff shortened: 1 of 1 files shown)


estenberg:
Awesome!
Sorry to feature-creep, but could you do the same in the docker-mender-convert container? Probably not urgent but good for maintenance reasons.

mirzak:
This [commit](https://github.com/mendersoftware/mender-convert/commit/a231c92edde3c01d7902dde03772bad60e2fbc5b) did remove it from the Dockerfile of the docker-mender-convert container.

So all is good

estenberg:
Ah, nice!
Once BeagleBone conversion works well inside docker container (WIP), we should probably update this doc to use it?

mirzak:
> Once BeagleBone conversion works well inside docker container (WIP), we should probably update this doc to use it?

That should also work now with @apodogrocki latest changes. Today at least I did convert beagle images using the docker-mender-convert.

mirzak:
It is time to create a Mender Hub post for Beagle as well.

estenberg:
Yep and yep! Let's follow up on that once we can get around to it. It's been on my list for a while, and now seems like good timing to "clean" this up.

mirzak:
Should this be back-ported to 1.7? 

The 1.7 docs are using version 1.0.0 of mender-convert so I would assume not unless we also update the version which includes the changes done.

estenberg:
No, as you mention mender-convert 1.0.0 was released with 1.7, so let's release 1.1.0 with 2.0 and make it available there. :)

PR #619, opened 2019-03-18
Document the swapping of StateScriptRetry configuration values

lluiscampos:
Changelog: None

### DIFF
--- a/05.Client-configuration/05.Configuration-file/50.Configuration-options/docs.md
+++ b/05.Client-configuration/05.Configuration-file/50.Configuration-options/docs.md
@@ -67,14 +67,20 @@ standard certificate trust chains.
 
 #### StateScriptRetryIntervalSeconds
 
+The timeout before a state script that previously returned the special retry
+return code (21), will be run again.
+
+<!--AUTOVERSION: "mender v%"/ignore-->
+*Note*: Before mender v2.0.0 release, this option used to be called `StateScriptRetryTimeoutSeconds`.
+
+#### StateScriptRetryTimeoutSeconds
+
 The interval for which the script is allowed to keep retrying. After this
 interval has expired, another retry attempt will be treated as a failure and if
 possible, the update will be rolled back.
 
-#### StateScriptRetryTimeoutSeconds
-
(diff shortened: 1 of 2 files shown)


lluiscampos:
@drewmoseley please have a look.

drewmoseley:
Looks ok to me.  @kacf how should we call this out? It seems like we really want to make sure users are aware of this change.

05.Client-configuration/05.Configuration-file/50.Configuration-options/docs.md
```diff
@@ -67,14 +67,14 @@ standard certificate trust chains.
...
+The timeout before a state script that previously returned the special retry
+return code (21), will be run again.
 
 #### StateScriptRetryTimeoutSeconds
 
-The timeout before a state script that previously returned the special retry
-return code (21), will be run again.
+The interval for which the script is allowed to keep retrying. After this
+interval has expired, another retry attempt will be treated as a failure and if
+possible, the update will be rolled back.
```
kacf: I would add a "historical" note to both of these, which says that the behavior used to be swapped before Mender 2.0.0.

kacf:
> how should we call this out? It seems like we really want to make sure users are aware of this change.

My plan was to use the changelog only, since it's a fairly detailed configuration setting. Do you think that's too low key?

05.Client-configuration/05.Configuration-file/50.Configuration-options/docs.md
lluiscampos: :+1:

lluiscampos:
New changes. Added the note for history reference.

drewmoseley:
A note in the changelog is fine by me.

lluiscampos:
FYI @drewmoseley, this will be the text in the Changelog:

MEN-2409: Swapped definition of StateScriptRetryTimeout and
StateScriptRetryInterval for the names to represent what they are
actually doing. This change breaks compatibility with current usage of
these configurable parameters. See documentation for correct usage.

PR #620, opened 2019-03-18
Document LZMA support and dependency

lluiscampos:
Changelog: None

### DIFF
--- a/.ispell_english
+++ b/.ispell_english
@@ -146,6 +146,7 @@ autoprefixer
 autorest
 autotools
 AUTOVERSION
+av
 AvianFlu
 AvrAgMBAAE
 awk
@@ -829,6 +830,7 @@ inode
 insertafter
 installables
 Insua
+int
 internet
 InventoryPollIntervalSeconds
 io
@@ -963,6 +965,7 @@ lib
 libc
 libdir
 libgcc
+liblzma
 libpseudo
 libs
 libtool
@@ -1013,6 +1016,7 @@ luiscouto
 lut
 ly
 Lydell
+lzma
 LZO
 mac
 mach
--- a/03.Devices/01.General-system-requirements/docs.md
+++ b/03.Devices/01.General-system-requirements/docs.md
@@ -5,7 +5,13 @@ taxonomy:
 ---
 
 ##Device capacity
-The client binaries are about 7 MB in size, or about 4 MB when debug symbols are stripped (using the `strip` tool). This includes all dependencies for the client, such as the http, TLS, and JSON libraries.
+The client binaries are about 7 MB in size, or about 4 MB when debug symbols are
+stripped (using the `strip` tool). This includes most of the dependencies for
+the client, such as the http, TLS, and JSON libraries.
+
+The client depends on the LZMA library for Artifact compression, which is
+present in most Linux distributions, including those based on the Yocto Project
+and the Debian family.
 
 ##Bootloader support
 To support atomic rootfs rollback, Mender integrates with the bootloader of the device. Currently Mender supports [GRUB](https://www.gnu.org/software/grub/?target=_blank) and [U-Boot](http://www.denx.de/wiki/U-Boot?target=_blank). The bootloader installation, features and requirements vary depending on the target OS in use.  Please see the [Yocto bootloader support](../yocto-project/bootloader-support) or [Debian bootloader support](../debian-family/bootloader-support) for more information.
--- a/04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
+++ b/04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
@@ -99,16 +99,20 @@ mender-artifact write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o 
 
 !!! If you would like to generate a *signed Artifact*, simply add the `-k` option with the path to your *private key*. In our example above, the full command would be `mender-artifact write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o artifact-signed.mender -k private.key`.
 
-## Artifact Compression
+## Artifact compression
 
-Currently the standard behaviour is to compress the artifact using `gzip`, however you can also opt out of the standard
-compression format by giving the `mender-artifact` command the global flag `compression`, and the argument `none`,
-meaning that no  compression will be done. Example:
+Mender Artifacts are compressed with the `gzip` algorithm by default. Mender
+supports three compression algorithms: `gzip` compression, `lzma` compression
+and no compression. This can be selected when creating artifacts with
+`mender-artifact` command with the global flag `compression`. For example, to
+select `lzma` compression instead of the default `gzip`, run the following
+command:
 
 ```bash
-mender-artifact --compression none write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o artifact.mender
+mender-artifact --compression lzma write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o artifact.mender
 ```
 
+Use `mender-artifact --help` to list all available compression options.
 
 ## Signing after modification
 
--- a/02.Architecture/04.Mender-Artifacts/docs.md
+++ b/02.Architecture/04.Mender-Artifacts/docs.md
@@ -64,7 +64,8 @@ interrupted, using [HTTP range requests](https://tools.ietf.org/html/rfc7233?tar
 To enable streaming and control based on metadata, like aborting the download
 if the Artifact is not compatible with the device, the Mender Artifact itself
 is not compressed. Instead, the root file systems within Artifacts are
-compressed, currently with the [gzip compression algorithm](https://en.wikipedia.org/wiki/gzip?target=_blank).
+compressed, currently with the [gzip compression algorithm](https://en.wikipedia.org/wiki/gzip?target=_blank)
+by default.
 
 
 ## Signing and verification
--- a/04.Artifacts/01.Yocto-project/99.Variables/docs.md
+++ b/04.Artifacts/01.Yocto-project/99.Variables/docs.md
@@ -345,7 +345,7 @@ value is `1024`.
 <!--AUTOVERSION: "introduced in the Yocto Project 2.7 (%)"/ignore-->
 !!! This variable was introduced in the Yocto Project 2.7 (warrior) release.
 
-The size of the optional swap partiton in the generated `.biosimg`, `.sdimg` or
+The size of the optional swap partition in the generated `.biosimg`, `.sdimg` or
  `.uefiimg` file. By default no swap partition is created, setting this variable 
 to a value greater than zero will add one after the other partitions.
 


lluiscampos:
@kacf @estenberg please have a look.

04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
```diff
@@ -101,14 +101,22 @@ mender-artifact write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o
...
+Currently the standard behaviour is to compress the artifact using `gzip`. Mender Artifact supports three
+compression modes: `gzip` compression, `lzma` compression and no compression. This can be selected when creating
+artifacts with `mender-artifact` command with the global flag `compression`. For example, to opt out of the
+compression you use argument `none` meaning that no compression will be done:
 
 ```bash
 mender-artifact --compression none write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o artifact.mender
 ```
 
+Similarly, to select `lzma` compression instead of the default `gzip`:
```
estenberg: I don't think it's necessary to have two examples about compression.
lzma will probably be way more popular than none, so I suggest removing the example of "none" compression. They can just check help for all available options.

03.Devices/01.General-system-requirements/docs.md
```diff
@@ -5,7 +5,9 @@ taxonomy:
 ---
 
 ##Device capacity
-The client binaries are about 7 MB in size, or about 4 MB when debug symbols are stripped (using the `strip` tool). This includes all dependencies for the client, such as the http, TLS, and JSON libraries.
+The client binaries are about 7 MB in size, or about 4 MB when debug symbols are stripped (using the `strip` tool). This includes dependencies for the client, such as the http, TLS, and JSON libraries.
+
+The only external dependency for the client is LZMA compression library, present already in most common Linux systems.
```
estenberg: This looks a bit concerning; any dependency required for an embedded device is a potential problem.

It is only required if the user uses the lzma compression type, right? So if I use gzip I don't have to have the LZMA library installed on my device?

If that is the case, I would make this clear here and reference the "Artifact compression" header for more information about different compression types.

estenberg: If my assumptions are right, I would add a new header "LZMA library (optional)" and also explain what exactly is required (is it a kernel setting, or what is the library called on e.g. ubuntu, yocto?).

kacf: > It is only required if the user uses the lzma compression type, right? So if I use gzip I don't have to have the LZMA library installed on my device?

The default is to compile with it. @lluiscampos, maybe you can document this way of building:

```
make TAGS=nolzma
```

And maybe check that it works also. I haven't tried it in practice.

.ispell_english
```diff
@@ -891,6 +893,7 @@ jwt
 JxSJ
 JyZL
 jzRXCnnu
+kacf
```
kacf: Such an honor!

04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
lluiscampos: Makes sense :+1:

03.Devices/01.General-system-requirements/docs.md
lluiscampos: > This looks a bit concerning; any dependency required for an embedded device is a potential problem.

That was exactly my intention when launching this PR, to create awareness of the dependency.

Anyway, as Kristian said it can be opt out during compilation (`make TAGS=nolzma` works), so I am rewriting this part of the doc following your suggestions.

lluiscampos:
New change set following @estenberg suggestions.

Also launched a [minor PR](https://github.com/kacf/mender/pull/12) for `mender` with the build instructions to opt-out LZMA compression

03.Devices/01.General-system-requirements/docs.md
```diff
@@ -7,6 +7,9 @@ taxonomy:
 ##Device capacity
 The client binaries are about 7 MB in size, or about 4 MB when debug symbols are stripped (using the `strip` tool). This includes all dependencies for the client, such as the http, TLS, and JSON libraries.
 
+For LZMA Artifact compression support, the LZMA shared library needs to be present in the target system. See
```
estenberg: "in the target system" -> "on the device" ?
We usually refer to this as a device, not target.

04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
```diff
@@ -99,16 +99,25 @@ mender-artifact write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o
...
 ```bash
-mender-artifact --compression none write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o artifact.mender
+mender-artifact --compression lzma write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o artifact.mender
 ```
 
+Use `mender-artifact --help` to list all available compression options.
+
+### LZMA compression support
+
+If you wish to use LZMA compression for Mender Artifacts, library LZMA shall be present in your system.
```
estenberg: "library LZMA shall be present in your system" -> "the LZMA library needs to be present on your device"

04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
```diff
@@ -99,16 +99,25 @@ mender-artifact write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o
...
+
+It is already present in most Linux Operating systems, but if you need to install it:
```
estenberg: Operating -> operating (lower caps)

.ispell_english
estenberg: Congratulations! I wonder when it is my turn..

lluiscampos:
Reworked following the last comments.

apodogrocki:
@lluiscampos for me it looks ok

04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
```diff
@@ -99,16 +99,25 @@ mender-artifact write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o
...
 ```bash
-mender-artifact --compression none write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o artifact.mender
+mender-artifact --compression lzma write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o artifact.mender
 ```
 
+Use `mender-artifact --help` to list all available compression options.
+
+### LZMA compression support
+
+If you wish to use LZMA compression for Mender Artifacts, the LZMA library needs to be present on your device.
```
kacf: We offer a binary download, so the dependency is not optional in that sense, so I think this should be changed to something like:

```
To use mender-artifact, the LZMA library needs to be present on your device.
```

But then maybe it should not be under its own header aboue LZMA?

This is all the more reason that we should offer mender-artifact in a package, not as a file.

lluiscampos: I disagree, I think we can still keep it optional.

Reasoning: I assume a `mender-artifact` compiled with LZMA (like the one we will offer for download) can still run in a system without liblzma, it will only get a runtime error when trying to compress/decompress an Artifact with LZMA.

Am I wrong?

kacf: Unfortunately yes:

```
/test # ./mender-artifact 
/bin/sh: ./mender-artifact: not found
/test # ldd mender-artifact 
	/lib64/ld-linux-x86-64.so.2 (0x7fcac53c4000)
Error loading shared library liblzma.so.5: No such file or directory (needed by mender-artifact)
	libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7fcac53c4000)
	libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7fcac53c4000)
Error relocating mender-artifact: lzma_code: symbol not found
Error relocating mender-artifact: lzma_easy_encoder: symbol not found
Error relocating mender-artifact: lzma_auto_decoder: symbol not found
Error relocating mender-artifact: lzma_end: symbol not found
/test # 
```

lluiscampos: Ouch! I think I confused my experiments here. I experimented compiling mender without having liblzma on the system, and maybe not executing it.

This brings us to square one: the dependency is mandatory. @estenberg insisted in that this is a problem.

Then I think the only place in the docs where it can be mentioned how to opt out the dependency is in the instructions for building the client (Yocto build and Cross compiling), everywhere else should be stated as mandatory (general requirements, install instructions, and mender artifact page).

estenberg: Ouch indeed! This was discovered a bit late.. :-/

Hard dependencies, especially on the device, which can have OSes based on Yocto, Buildroot, Debian, Ubuntu, Raspbian, OpenWRT, etc. will for sure lead to problems. The problem is probably less on the workstation/laptop side as it seems to be a common library. But is it present on MacOS? I have a feeling this will lead to lots of unhappy experiences & repeated questions/diagnostics if we don't follow up.

Is it much work to disable LZMA by default, both for client (all OSes, assuming Makefile) and mender-artifact builds? If not, I would suggest doing this quickly (ideally before beta).

If this is feasible then create two more tasks (can be done after beta):
- Document how to enable LZMA (Yocto, all other client builds, mender-artifact)
- Research support for LZMA across OSes to see if we should enable by default, at least in some cases

Does that make sense @kacf @lluiscampos ?

kacf: First off, let's keep this discussion on mender-artifact. liblzma is *extremely* common on Linux systems, systemd depends on it, GRUB depends on it, dpkg depends on it. I had to go out of my way to produce that error message (Alpine, which is supposed to be bare minimum, and very old versions of CentOS were the only ones I could find that didn't have liblzma in the minimal install).

In fact, once I found out those dependencies, I think I changed my position regarding this: We should not document this at all, because *everyone* has this library. *Maybe* one entry in the troubleshooting guide if we want to be safe.

MacOS is irrelevant in this discussion I think, since we will decide the level of support for mender-artifact once we produce that package.

That's for mender-artifact. For mender, I still think that this is no problem at all, since we already depend on systemd, and systemd depends on liblzma. But here we should probably document it, since embedded systems are a bit more sensitive in this regard. I'm posting another comment in the device part.

03.Devices/01.General-system-requirements/docs.md
```diff
@@ -7,6 +7,9 @@ taxonomy:
 ##Device capacity
 The client binaries are about 7 MB in size, or about 4 MB when debug symbols are stripped (using the `strip` tool). This includes all dependencies for the client, such as the http, TLS, and JSON libraries.
 
+For LZMA Artifact compression support, the LZMA shared library needs to be present on the device. See
+[Artifact Compression](../../artifacts/modifying-a-mender-artifact#artifacts-compression) for more details.
```
kacf: This is good, but maybe add: "...for more details, particularly how to disable LZMA support."

04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
lluiscampos: Discussed with @kacf in person now.

Yes, I think I scaled this a bit much and it is not a big deal after all.

The motivation behind the this PR was more of a principle than a practical problem. I thought that we could not "claim" anymore to have all our dependencies statically linked and that was a problem for us.

But I agree with @kacf that it is quite unlikely that someone actually bumps into the problem.

I am going to rewrite my changes to focus in the following ideas:
* In "General system requirements" page
  * There is a dependency on LZMA, but you don't need to worry about
  * If you want to disable it: follow these instructions in README to cross-compile the client or modify your yocto environment accordingly (we don't need to write instructions in how to disable LZMA for mender-artifact)
* In "Artifacts" page
  * Info about how to use lzma for compression (assume always present)

lluiscampos: I decided to skip also the comments on how to opt out. For the cross-compile instructions it will be feasible, but for the Yocto part will require deep changes as it needs to change the build options of mender.inc

lluiscampos:
I pushed a simplified change set for this issue.

(I did it in a different commit in case I have to revert, I will squash the commits once we are ok with the text)

03.Devices/01.General-system-requirements/docs.md
```diff
@@ -5,7 +5,9 @@ taxonomy:
 ---
 
 ##Device capacity
-The client binaries are about 7 MB in size, or about 4 MB when debug symbols are stripped (using the `strip` tool). This includes all dependencies for the client, such as the http, TLS, and JSON libraries.
+The client binaries are about 7 MB in size, or about 4 MB when debug symbols are stripped (using the `strip` tool). This includes most of the dependencies for the client, such as the http, TLS, and JSON libraries.
+
+The client depends on LZMA library for Artifact compression, which is present in most of Linux distributions including Yocto project and Debian family ones.
```
estenberg: ...Yocto Project... (capital P)

04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
```diff
@@ -99,16 +99,17 @@ mender-artifact write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o
 
 !!! If you would like to generate a *signed Artifact*, simply add the `-k` option with the path to your *private key*. In our example above, the full command would be `mender-artifact write rootfs-image -t beaglebone -n release-1 -u rootfs.ext4 -o artifact-signed.mender -k private.key`.
 
-## Artifact Compression
+## Artifact compression
 
-Currently the standard behaviour is to compress the artifact using `gzip`, however you can also opt out of the standard
-compression format by giving the `mender-artifact` command the global flag `compression`, and the argument `none`,
-meaning that no  compression will be done. Example:
+Currently the standard behaviour is to compress the Artifact using `gzip`. Mender Artifact supports three
```
estenberg: Would say "Mender Artifacts are compressed with the `gzip` algorithm by default. Mender supports three compression algorithms: ..."

03.Devices/01.General-system-requirements/docs.md
lluiscampos: :+1:

04.Artifacts/05.Modifying-a-Mender-Artifact/docs.md
lluiscampos: :+1:

lluiscampos:
Great, thanks!

I modified the last details and also squashed the commits into one. I think this is ready to be merged.

03.Devices/01.General-system-requirements/docs.md
```diff
@@ -5,7 +5,12 @@ taxonomy:
 ---
 
 ##Device capacity
-The client binaries are about 7 MB in size, or about 4 MB when debug symbols are stripped (using the `strip` tool). This includes all dependencies for the client, such as the http, TLS, and JSON libraries.
+The client binaries are about 7 MB in size, or about 4 MB when debug symbols are
+stripped (using the `strip` tool). This includes most of the dependencies for
+the client, such as the http, TLS, and JSON libraries.
+
+The client depends on LZMA library for Artifact compression, which is present in
```
estenberg: ...depends on *the* LZMA...

03.Devices/01.General-system-requirements/docs.md
```diff
@@ -5,7 +5,12 @@ taxonomy:
...
+most of Linux distributions including Yocto Project and Debian family ones.
```
estenberg: ...present in most Linux... (no "of")
also would add , after "distributions"
also, reword "ones" and say based on.

In total: "..., which is present in most Linux distributions, including those based on the Yocto Project and the Debian family."

lluiscampos:
Fixed :+1:

lluiscampos:
Great! Thanks for a thorough review.

PR #622, opened 2019-03-26
MEN-2391 update modules spec link

lluiscampos:
As the file is not present in master, we cannot use AUTOREVISION atm; we
will fix it after the final merge into master.

Changelog: None

### DIFF
--- a/04.Artifacts/08.Update-Modules/docs.md
+++ b/04.Artifacts/08.Update-Modules/docs.md
@@ -204,4 +204,5 @@ It saves a tarball with all current files in `/var/www` (only the single files, 
 
 ## Further reading
 
-* [Update modules v3 protocol](https://github.com/kacf/mender/blob/update_modules/Documentation/update-modules-v3-file-api.md)
+<!--AUTOVERSION: "blob/%/Documentation"/mender -->
+* [Update modules v3 protocol](https://github.com/mendersoftware/mender/blob/2.0.0b1/Documentation/update-modules-v3-file-api.md)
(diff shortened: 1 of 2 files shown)


04.Artifacts/08.Update-Modules/docs.md
```diff
@@ -204,4 +204,5 @@ It saves a tarball with all current files in `/var/www` (only the single files,
 
 ## Further reading
 
-* [Update modules v3 protocol](https://github.com/kacf/mender/blob/update_modules/Documentation/update-modules-v3-file-api.md)
+<!--AUTOVERSION: "blob/%/Documentation"/ignore -->
+* [Update modules v3 protocol](https://github.com/mendersoftware/mender/blob/2.0.0b1/Documentation/update-modules-v3-file-api.md)
```
kacf: Use `/mender` in the AUTOVERSION tag. We want to update this on the next release, no?

lluiscampos: But won't that generate a dead link in development version of the docs? That is what I was trying to avoid.

kacf: Since the master branch stays constantly on master, it is never actually updated. So it's fine to keep a non-master tag there for now, as long as we remember to switch it back manually.

If anyone did try to update master, then indeed this would break, but I don't expect that.

lluiscampos: Understood. Fixed now.

lluiscampos:
Now that i am beta testing the .deb package, I have few fixes also for the Installing documentation. Let's wait to get them here and we do one merge only

lluiscampos:
Now I think it can be merged.

Thanks for the `tee` tip

lluiscampos:
I am merging without your approval to be able to close the task in the sprint. This is an exception on the procedure.

PR #625, opened 2019-03-28
Document various Update modules behavior changes

kacf:


04.Artifacts/07.State-scripts/docs.md
```diff
@@ -65,13 +72,13 @@ All other return codes are reserved for future use by Mender and should not be u
...
+
+!!! Tip: Since a power loss in `ArtifactReboot_Enter` will not be marked as a failure, this enables the user to stall the daemon in `ArtifactReboot_Enter` state forever, and thus an install will only be installed on a power-cycle of the device.
 
-!!! Note that there is one exception. A power loss in `ArtifactReboot_Enter` will not be marked as a failure. This enables the user to stall the daemon in reboot_enter state forever, and thus an install will only be installed on a power-cycle of the device.
+This means that state scripts should be written to be idempotent. This means that re-running the script several times, even partially, should have the same effect as running it once, as long as the last execution is a complete one.
```
lluiscampos: This means that .... This means that...

Suggestions:
* "For this reason, state scripts should be..."
* or just crop the first part like "State scripts should be written to be idempotent. This means that..."

lluiscampos: (if you change it, do it also in the Update Modules docs below)

04.Artifacts/07.State-scripts/docs.md
```diff
@@ -65,13 +72,13 @@ All other return codes are reserved for future use by Mender and should not be u
...
+## Power loss
+
+In general power loss means that Mender will transition into an error state, either ArtifactRollback or ArtifactFailure, depending on the level of rollback support of the [Update Module](../update-modules) in question. If a power loss happens inside one of the error states, that state will be repeated until it succeeds without a power interruption. However, there are some exceptions: Power loss is allowed to happen within any Reboot state, since it may be indistinguishable from a normal reboot. Also `ArtifactCommit_Leave` will be repeated just like the error states, since after committing it is too late to do a rollback.
```
mirzak: > depending on the level of rollback support of the [Update Module](../update-modules) in question

This should be an exception for update modules right? The wording now suggests that this is only supported by update modules.

kacf:
@lluiscampos, @mirzak: I fixed both your concerns now. See if you agree with the new wording.

mirzak:
Yepp, looks good.

lluiscampos:
Perfect! :+1:

kacf:
The build is green, it's just the Github status which is stuck for some reason.

PR #626, opened 2019-03-28
Release updates

kacf:


202.Release-information/01.Release-notes-changelog/docs.md
```diff
@@ -4,6 +4,424 @@ taxonomy:
     category: docs
 ---
 
+## mender-convert 1.1.0b1
```
lluiscampos: That mender-convert 1.1.0b1 comes before mender 2.0.0b1 is done manually? Could we swap them?

kacf: For consistency I think it makes sense to keep them in tag age order (the topmost components build on the lower components), and there will anyway be yet another set of changelogs on top of this once I tag meta-mender. I argued for keeping each component on its own page some time ago, but this was voted down.

lluiscampos: Ok. Thanks for clarification.

PR #627, opened 2019-03-28
Add changelog for meta-mender/thud-v2019.03.

kacf:


kacf:
Travis is extremely slow today. These are pretty standard changes, so will just merge to unblock the release.

PR #630, opened 2019-04-01
Update mender artifact docs

lluiscampos:
The main change is the second commit:

Improve instructions for mender-artifact build: version and deps
    
We want the instructions to point to the latest released version. This
is important now that for the 3.0.0b release the code is not merged into
master, so without the checkout we cannot build mender-artifact suitable
for Update Modules
    
Also, document de LZMA depdendency with specific instructions for Mac OS
users, as we do not provice a native binary for them (yet).


lluiscampos:
@mzedel Please beta test the new instructions on Mac OS

@eysteins Please have a look. I will port this to 2.0.x docs release once it is approved.

estenberg:
:+1: 

At some point we should split out all the build & tool download instructions into a separate page / area of the docs. But that is out of scope here I think. Looks good!

lluiscampos:
Thank you @mzedel for the beta testing.

I merged it before your feedback, so I needed to open a new PR for the last change, see #634 Sorry for the inconvenience.

PR #639, opened 2019-04-11
Fix mender standalone '-install' argument.

kacf:


kacf:
Yes, we need this in master and 2.0.x, not 1.7.x.

lluiscampos:
Yes, but that is my point, that in master the mender client still uses the -rootfs argument

kacf:
Ah, you're right. I suggest we just leave it though, it won't be long.

Cherry-picked to 2.0.x.

PR #642, opened 2019-04-15
client: install: remove mention of demo server

mirzak:
All the provided configuration files point to Hosted Mender, and the
section "Use demo settings" only updates polling intervals and not
server url, meaning it is still using Hosted Mender.

Changelog: None


estenberg:
This makes sense as it is now, however I think we should have a section on how to use the demo server..

mirzak:
Merged this for now. 

I have some unrelated follow up changes and I do think as well that we should a a section on how to connect to demo server, but this is more of a extension then replacing the current instructions. The certificate is part of the Mender client source coude `demo.crt`

PR #648, opened 2019-04-18
update mender install instructions

mirzak:
I am also thinking if we should remove this section:

- https://docs.mender.io/2.0/client-configuration/cross-compiling

It is now covered in the Mender client [README](https://github.com/mendersoftware/mender/tree/update_modules#installing-from-source)

### DIFF
--- a/05.Client-configuration/06.Installing/docs.md
+++ b/05.Client-configuration/06.Installing/docs.md
@@ -4,64 +4,95 @@ taxonomy:
     category: docs
 ---
 
-This page describes how to install the Mender client in an existing Linux system.
+This page describes how to install the Mender client in an existing Linux system. Installing this way does not offer a a full Mender integration. However, it is possible to use Update Modules and update parts of the system.
 
-If rootfs updates are also desired, full board integration is required. In this case, follow instead the device documentation on [Yocto Project](../../devices/yocto-project) or [Debian family](../../devices/debian-family).
+If full board integration is desired, follow the device documentation on [Yocto Project](../../devices/yocto-project) or [Debian family](../../devices/debian-family).
 
-## Install prebuilt package
+## Install from source
 
-An official `.deb` package with the Mender client that supports the ARM-v6 architecture is provided below. This package should work on most operating systems in the Debian family (e.g. Debian, Ubuntu, Raspbian) and devices (e.g. Raspberry Pi 2/3, Raspberry Pi Zero W, BeagleBone).
+<!--AUTOVERSION: "mender/tree/%#installing-from-source"/mender -->
+It is possible to install the Mender client from source by following the guidelines outlined in the [README.md](https://github.com/mendersoftware/mender/tree/master#installing-from-source) of the Mender client source repository.
 
-Run the following commands on your device to install the package:
+## Install Mender provided Debian package
+
+A Debian package (`.deb`) is provided for convince to install on e.g Debian, Ubuntu or Raspbian. We provide packages for the following architectures:
+
+- armhf (ARM-v6)
+    - Raspberry Pi, BeagleBone and other ARM based devices.
+
+Download the package:
 
 <!--AUTOVERSION: "cloudfront.net/%/"/mender "mender_%-1_armhf.deb"/mender -->
 ```bash
 wget https://d1b0l86ne08fsf.cloudfront.net/master/dist-packages/debian/armhf/mender_master-1_armhf.deb
+```
+
+Install the package:
+
+<!--AUTOVERSION: "mender_%-1_armhf.deb"/mender -->
+```bash
 sudo dpkg -i mender_master-1_armhf.deb
 ```
 
-## Setup
+### Setup
 
-After successfully installing the Mender client, some initial setup is required.
+After successfully installing the Mender client Debian package, some initial setup is required.
 
-### Mender configuration file
+#### Mender configuration file
 
-First, we have to configure the Mender client with the configuration file at `/etc/mender/mender.conf`.
+First, we have to configure the Mender client using the configuration file at `/etc/mender/mender.conf`.
 
-To configure the Mender client for Hosted Mender, you need to edit this file and insert your Tenant Token
-where it says "Paste your Hosted Mender token here". Run the following commands:
-
-```bash
-TENANT_TOKEN="<INSERT YOURS FROM https://hosted.mender.io/ui/#/settings/my-organization>"
-sudo sed -i "s/Paste your Hosted Mender token here/$TENANT_TOKEN/" /etc/mender/mender.conf
-```
-
-#### Use demo settings (optional)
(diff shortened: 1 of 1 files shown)


05.Client-configuration/06.Installing/docs.md
```diff
@@ -4,64 +4,96 @@ taxonomy:
...
+```bash
 sudo sed -i "s/Paste your Hosted Mender token here/$TENANT_TOKEN/" /etc/mender/mender.conf
 ```
 
-### Device type
+#### Device type
 
-The device type is a string that defines your device and the type of updates it supports. Run the following
-command to set `raspberrypi3` as device type:
+The device type is a arbitrary string that defines your device and is used to ensure software compatibility by comparing the device type set in a [Mender Artifact](../../architecture/mender-artifacts) with the string on the device.
```
estenberg: Would drop "arbitrary"

estenberg:
Nice improvements!

I agree the cross compiling section is probably too "low level" for the docs and better suited for README.md. However, the docs page is a bit more prescriptive (e.g. I would not know what to do from this in the README.md: "Download the cross-compiler required for your device"). What is in the docs should work for most ARM based devices? If so maybe we should copy something over to the README.md before removing it?

Also, I know I made some inbound links to cross compiling in docs when describing mender-convert. But it looks like they are removed now -- good. :)

05.Client-configuration/06.Installing/docs.md
```diff
@@ -4,64 +4,96 @@ taxonomy:
     category: docs
 ---
 
-This page describes how to install the Mender client in an existing Linux system.
+This page describes how to install the Mender client in an existing Linux system. Installing this way does not offer a complete system updater. However, it is possible to use update modules and update other parts of the system.
```
drewmoseley: Do we use "complete system' updater elsewhere or is there a clearer term we can use here? In conversation I generally use "full image" updates but I'm not sure how consistently we use that.

lluiscampos: So far we didn't need to distinguish, so I don't think there is a clear nomenclature for it. It is a good point to define these, though, and be consistent in our wording in the docs, Mender Hub, blog posts...

To avoid the problem, I would say something like "...does not offer full Mender integration. However,...". Reasoning: it is consistent with the later reference to "full board integration" and the link to the corresponding doc section.

lluiscampos: Also, please use Update Modules (capitalized). That is something I've been using everywhere in the docs.

05.Client-configuration/06.Installing/docs.md
```diff
@@ -4,64 +4,96 @@ taxonomy:
...
-## Setup
+### Setup
+
+After successfully installing the Mender client Debian package, some initial setup is required.
 
-After successfully installing the Mender client, some initial setup is required.
+#### Mender configuration file
 
-### Mender configuration file
+First, we have to configure the Mender client using the configuration file at `/etc/mender/mender.conf`.
```
lluiscampos: Missing a quote? \`/etc/mender...\`

lluiscampos:
Even though I don't necessarily think it is a problem to duplicate info between docs/README (in my eyes, there have different target audience), in this case I agree we can remove the Cross-compile section from the docs site, as we already have other options to get the client in the device (native build + install, or deb package).

If you do this, make sure we transfer the necessary info from the docs to the README.md. If you don't have time to do this, I can take it later this week out of this PR

05.Client-configuration/06.Installing/docs.md
```diff
@@ -4,64 +4,96 @@ taxonomy:
...
+This page describes how to install the Mender client in an existing Linux system. Installing this way does not offer a complete system updater. However, it is possible to use update modules and update other parts of the system.
 
-If rootfs updates are also desired, full board integration is required. In this case, follow instead the device documentation on [Yocto Project](../../devices/yocto-project) or [Debian family](../../devices/debian-family).
+If filesystem image updates are also desired, full board integration is required. In this case, follow instead the device documentation on [Yocto Project](../../devices/yocto-project) or [Debian family](../../devices/debian-family).
 
-## Install prebuilt package
+## Install from source
 
-An official `.deb` package with the Mender client that supports the ARM-v6 architecture is provided below. This package should work on most operating systems in the Debian family (e.g. Debian, Ubuntu, Raspbian) and devices (e.g. Raspberry Pi 2/3, Raspberry Pi Zero W, BeagleBone).
+<!--AUTOVERSION: "mender/tree/%#installing-from-source"/mender -->
```
lluiscampos: This AUTOVERSION is not being properly processed. Is it due to the following blank line?

kacf: Yes, blank lines reset the expression to empty, except inside source blocks.

05.Client-configuration/06.Installing/docs.md
mirzak: Where? According to my eyes this is wrapped correctly :)

mirzak:
Pushed updates, and I believe I have resolved all comments/suggestions.

I will not remove the cross-compile section as part of this PR.

05.Client-configuration/06.Installing/docs.md
mirzak: Removed the blank line, and how do you check if these are resolved correctly?

lluiscampos: You can run `./autoversion.py --check` locally

05.Client-configuration/06.Installing/docs.md
lluiscampos: Yes, you are right. I don't know what happened with "my eyes" :smile:

05.Client-configuration/06.Installing/docs.md
```diff
@@ -4,64 +4,95 @@ taxonomy:
...
+Download the package:
 
 <!--AUTOVERSION: "cloudfront.net/%/"/mender "mender_%-1_armhf.deb"/mender -->
 ```bash
 wget https://d1b0l86ne08fsf.cloudfront.net/master/dist-packages/debian/armhf/mender_master-1_armhf.deb
+```
+
+Install the package:
+
+<!--AUTOVERSION: ""mender_%-1_armhf.deb"/mender -->
```
lluiscampos: Typo: extra quotes

mirzak: Fixed

mirzak:
@drewmoseley  @eysteins , can you to take on last look before we merge this?

drewmoseley:
LGTM

lluiscampos:
@mirzak I will merge this one and port it to 2.0.x :+1:

mirzak:
Thanks, and thanks for fixing the "AUTOVERSION" tag for 2.0.x

PR #663, opened 2019-04-30
master branch updates

drewmoseley:


drewmoseley:
@estenberg look ok?

estenberg:
Yep! We could probably drop some of the older Mender client versions too (e.g. one year or older), but  nice improvement on its own.

PR #666, opened 2019-05-03
add a note to user creation in demo and on-prem setup

mirzak:

### DIFF
--- a/01.Getting-started/02.Create-a-test-environment/docs.md
+++ b/01.Getting-started/02.Create-a-test-environment/docs.md
@@ -52,6 +52,8 @@ The UI requires you to log in with a valid username and password. You must first
 ./demo exec mender-useradm /usr/bin/useradm create-user --username=myusername@example.com --password=mysecretpassword
 ```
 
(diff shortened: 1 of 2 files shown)


drewmoseley:
I'll cherry-pick into 2.0.x as well.

estenberg:
Nice! Some thoughts:
* It was (at least not my) intention to set a secure password on the CLI.. this will get stored in the history
* Maybe it is possible to add single quotes to avoid shell expansion

mirzak:
> It was (at least not my) intention to set a secure password on the CLI.. this will get stored in the history

This is what we recommend in our on-prem tutorial, are there other ways of setting the password? I guess one could set a temp password and the change it using the GUI.

Maybe we should add a warning or write a wrapper/helper script that takes password as input using the `read` command, this way one can avoid it being echoed back and will not be visible in history.

> Maybe it is possible to add single quotes to avoid shell expansion

True, it might have helped. Though unknown if it covers all cases.

estenberg:
Indeed, I think the UI would be the "right" place to set a secure password, it also has a generator.

So I think at least for now we can leave it as an exercise for the user.. though a small script would be welcome (can rather say run this script than this long command and it would take everything as input).

PR #676, opened 2019-05-07
Rename Mender .deb package to mender-client

lluiscampos:
From release 2.0 on, the Debian package is named `mender-client` and
must force the users to remove the beta package `mender`

Changelog: None

### DIFF
--- a/05.Client-configuration/06.Installing/docs.md
+++ b/05.Client-configuration/06.Installing/docs.md
@@ -22,16 +22,16 @@ A Debian package (`.deb`) is provided for convince to install on e.g Debian, Ubu
 
 Download the package:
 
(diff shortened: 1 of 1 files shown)


lluiscampos:
See related [mender-dist-packages PR](https://github.com/mendersoftware/mender-dist-packages/pull/6)

mirzak:
I personally do not think that we should add this to our docs, it will get out of date fast. It is kind of expected for packages transitioning from "beta" to final to break things.

Though I do feel that this is worth mentioning in our "2.0 final release announcement".

lluiscampos:
@mirzak Thanks for your input. I don't like it neither so I am close to agree with you in just dropping it.

I felt it will generate many questions and then decided to launch the PR to get opinions. @kacf ?

lluiscampos:
Great! Thank you both. Reworked to only include the renaming in the PR

PR #684, opened 2019-05-08
update requirements for on-prem installation

mirzak:
I do not actually know if the values in the requirements are appropriate but recently we have had a number of questions/issues on the forums reporting failure to setup the on-prem server, and most where related to RAM memory which we seem to have increased recently due to update of Elasticsearch.

### DIFF
--- a/07.Administration/02.Production-installation/docs.md
+++ b/07.Administration/02.Production-installation/docs.md
@@ -16,11 +16,12 @@ repository.
 
 ## Prerequisites
 
-- A machine with Ubuntu 18.04 (e.g. an instance from AWS EC2 or DigitalOcean) with
-  the following tools installed:
+- A machine with Ubuntu 18.04 (e.g. an instance from AWS EC2 or DigitalOcean) with the following tools installed:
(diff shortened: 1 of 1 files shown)


estenberg:
This is almost impossible to quantify generically unfortunately - it depends on the environment...
For production installs, especially at scale, either we can provide some PS to help or the customer needs to do research on how it behaves and create performance requirements for the infra.

So I would say this is a bit out of scope for the documentation. Maybe qualify it like this:

"Minimum of **10 GB free disk space** and **4 GB RAM** available for Mender and its services. This heavily depends on your scale and environment, the supported [Mender Enterprise](https://mender.io/product/mender-enterprise) edition is recommended for larger-scale environments."

0lmi:
This is good practice to popularize paid offerings in such way. Good point.

mirzak:
My intention is mostly to avoid people failing to set up an initial server because they have not enough RAM, but if we can get an link to Mender Enterprise in there all the better :).

I have updated based on your suggestion @estenberg

PR #691, opened 2019-05-10
Add Command options' meaning

ajithpv:
It is always better to explicitly mention the meaning of each option(s) provided with mender artifact command.
Added the meaning of provided options for update module example.
Changelog: None

Signed-off-by: Ajith P V <ajithpv@outlook.com>

PR #692, opened 2019-05-10
Add Command options' meaning

lluiscampos:
It is always better to explicitly mention the meaning of each option(s) provided with mender artifact command.
Added the meaning of provided options for update module example.
Changelog: None

Signed-off-by: Ajith P V <ajithpv@outlook.com>
(cherry picked from commit d838fe41cf110166f3f9a05ea121bdf3ff5fa7d8)


lluiscampos:
Cherry-picking #691

PR #696, opened 2019-05-15
Changelogs backport meta mender v2019.05

lluiscampos:


lluiscampos:
@kacf In addition to cherry-picking for mender-docs 2.0.x, should we also cherry-pick it for 1.7.x?

PR #705, opened 2019-06-03
Update download links for mender-artifact

lluiscampos:
@mirzak @michaelatmender Please comment on this one.

Maybe it is clearer if we move it to a new page with a table, where we will eventually add Windows platform as well. In this case, do you have a good suggestion for which section to place the new page?

Thanks


mirzak:
Agree that it would make more sense to create a separate "Download" page with a table with the different platforms. It probably also makes sense to move the ".deb" package of the Mender client to the same location, as I assume that we will have multiple platforms there as well over time.

I would suggest to create a new top level section e.g `08.Downloads"

lluiscampos:
@mirzak Great, thanks. I think it is a good idea so I'll go for it.

What about the demo images download links, should we also move them to the new section oe leave them in `01.Getting-started/05.Donwload-tset-images`?

mirzak:
I think it is ok to leave the demo images at the current location, because this is only related to "demo environment" and that is what the `01.Getting-started` section covers.

lluiscampos:
Created a new section and updated the references accordingly.

For the mender (client) documentation I decided not to refer to this new section. As it is now, if follows better the flow of working remotely in a device (i.e. wget + dpkg). Whenever we have pkgs for several architectures, we can re-work it.

08.Downloads/docs.md
```diff
@@ -0,0 +1,40 @@
...
+[x.x.x_mender-artifact-darwin]:
+https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/master/mender-artifact-darwin
+
+
+## Mender Client
+
+A Debian package (`.deb`) is provided for convenience to install on e.g Debian, Ubuntu or Raspbian. We provide packages for the following architectures:
+
+- armhf (ARM-v6)
+    - Raspberry Pi, BeagleBone and other ARM based devices.
```
mirzak: Why not a table similar to "Mender Artifact tool"?

08.Downloads/docs.md
```diff
@@ -0,0 +1,40 @@
+---
+title: Downloads
+taxonomy:
+    category: docs
+---
+
+## Mender Artifact tool
```
mirzak: Unknown about taxonomy here, referring to "tool", but I see that this copied from previous version of the docs so not making it worse :)

We do also referrer to this as Mender Artifact CLI

08.Downloads/docs.md
```diff
@@ -0,0 +1,40 @@
...
+Remember to add execute permission (e.g. with `chmod +x mender-artifact`).
+
+!!! If you need to build `mender-artifact` from source, please see [Compiling
+mender-artifact](#compiling-mender-artifact).
+
+ <!--AUTOVERSION: "mender-artifact/%/"/mender-artifact -->
+[x.x.x_mender-artifact-linux]:
+https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/master/mender-artifact-linux
+[x.x.x_mender-artifact-darwin]:
+https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/master/mender-artifact-darwin
```
mirzak: The binaries that are downloaded are now called `mender-artifact-linux/mender-artifact-darwin`, but all our docs are still referring to `mender-artifact` when executing examples. I probably missed the discussion regarding this.

I would at least add a note that one probably needs to rename the downloaded binary to `mender-artifact`.

08.Downloads/docs.md
```diff
@@ -0,0 +1,40 @@
...
+
+
+## Mender Client
+
+A Debian package (`.deb`) is provided for convenience to install on e.g Debian, Ubuntu or Raspbian. We provide packages for the following architectures:
+
+- armhf (ARM-v6)
+    - Raspberry Pi, BeagleBone and other ARM based devices.
+
+Download link: [mender-client_x.x.x-1_armhf.deb]
```
mirzak: Why the `x.x.x-1` in the name instead of having the actual name?

08.Downloads/docs.md
lluiscampos: Yes, I am a bit unsure here. Last time I cleaned up some parts of this, we agreed in that we should use only "Mender Artifact". I'll fix

08.Downloads/docs.md
lluiscampos: There has not been a proper discussion, thanks that you pointed it out!

@kacf and me we talked about it but from a different angle, and I didn't pay attention in that we would need to modify docs and so on. Now that I think about it, `mender-artifact` is well established in the docs, in Mender Hub, in our tests...

I am changing my mind with this and instead of uploading them as mender-artifact-linux/darwin, what if we just have a linux/darwin/(windows) folder and the binary itself is still named `mender-artifact`.

I then have to rework the automated upload but I think it is worth. @kacf opinions?

08.Downloads/docs.md
lluiscampos: Typo (not adding the link), I'll fix

08.Downloads/docs.md
lluiscampos: I thought that for a single link it could be okey with the text. But when I now think about it, we have a "list" of one item, so we might very well have a "table" of one row. I'll fix

08.Downloads/docs.md
kacf: Fully support that. Manuel and I was just discussing now the effects on update modules as well. It will essentially break all update modules to have os specific suffixes on mender-artifact. So we were a bit too quick to decide this, and I agree with your proposal.

lluiscampos: :+1:

lluiscampos:
Changes pushed.

The binaries are now uploaded to the new locations

08.Downloads/docs.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Mender Artifact
+
+To download `mender-artifact` as an standalone tool, follow the correct link
+according to your host platform:
+
+| Platform | Download link                                                |
+|----------|--------------------------------------------------------------|
+| Linux    | [mender-artifact for Linux][x.x.x_mender-artifact-linux]     |
+| Mac OS X | [mender-artifact for Mac OS X][x.x.x_mender-artifact-darwin] |
```
mirzak: The "for Linux/for Mac OS X" strings seem superfluous to me, and maybe add the version string here instead?

lluiscampos: :+1:

lluiscampos:
New change set applying the last comment of @mirzak . Merging

08.Downloads/docs.md
```diff
@@ -0,0 +1,41 @@
...
+
+<!--AUTOVERSION: "mender-artifact %"/mender-artifact -->
+| Platform | Download link                                                |
+|----------|--------------------------------------------------------------|
+| Linux    | [mender-artifact master][x.x.x_mender-artifact-linux]     |
+| Mac OS X | [mender-artifact master][x.x.x_mender-artifact-darwin] |
+
+Remember to add execute permission (e.g. with `chmod +x mender-artifact`).
+
+!!! If you need to build `mender-artifact` from source, please see [Compiling
```
estenberg: These will (unfortunately) not render correctly if split into multiple lines. Need to put on one (long) line. See: https://docs.mender.io/development/downloads

08.Downloads/docs.md
```diff
@@ -0,0 +1,41 @@
...
+mender-artifact](#compiling-mender-artifact).
+
+<!--AUTOVERSION: "mender-artifact/%/"/mender-artifact -->
```
estenberg: The links don't seem to be substituted -- but maybe it is coming? Please see: https://docs.mender.io/development/downloads

estenberg: Maybe it's an issue with formatting of the reference from the table?

lluiscampos: It is unfortunate that we don't have an easy way to generate locally the docs site, so that I am blind when it comes to these changes.

Thanks for reporting, I'll try to fix ASAP

kacf: Actually, there is a way, not completely streamlined, but not too much trouble either. Check out https://github.com/mendersoftware/mender-docs-site and its README.

lluiscampos: @kacf Thanks!

08.Downloads/docs.md
lluiscampos: :+1:

08.Downloads/docs.md
lluiscampos: Fixed now in PR #708

PR #711, opened 2019-06-10
Update Update-Modules Example Instructions

reachBAM:
I have added a note about how to resolve a potential failure when following the example.  

I also added a new section describing how to run the example in standalone mode.

### DIFF
--- a/03.Devices/10.Update-Modules/docs.md
+++ b/03.Devices/10.Update-Modules/docs.md
@@ -41,7 +41,7 @@ Refer to [further reading](#further-reading) for more details.
 
 ### Mender server
 
-You will need a Mender server.
+You will need a Mender server if you are using [managed mode](https://docs.mender.io/2.0/architecture/overview#modes-of-operation).
 
 For easy setup, use [Hosted Mender](https://hosted.mender.io?target=_blank) or the [on-premise demo server](../../getting-started/create-a-test-environment).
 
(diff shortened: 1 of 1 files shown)


03.Devices/10.Update-Modules/docs.md
```diff
@@ -132,12 +137,37 @@ The command line options are detailed below:
...
+scp web-file-1.mender <username>@<deviceip>:/home/<username>
+```
+
+Then deploy the artifact from the target device:
+
+```bash
+mender -install web-file-1.mender
+```
+
+__NOTE:__ If the following error is reported:
```
lluiscampos: This part shouldn't be necessary if the reader has followed installed the client thought the deb package, as pointed out in the Prerequisites section.

I think it is better to remove it as it is a detail only if you have a custom installation of the client.

03.Devices/10.Update-Modules/docs.md
```diff
@@ -64,6 +64,11 @@ If you use Linux, [download the prebuilt mender-artifact binary][x.x.x_mender-ar
 
 In this first basic example we will create an Update Module that copies files into `/var/www` directory on the device.
 
+__Note:__ If the folder `/var/www` doesn't exist on the device it must be created first:  
+```bash
+mkdir /var/www
+```
```
lluiscampos: Good catch. But instead of having some like "if such folder does not exist, do this", I think it is better to just ask the user to create the directory regardless of if it exists or not.

So please move this to the next subsection, Create the Update Module script, adding something like:

First create the directory where the update files will be installed:
mkdir -p /var/www

03.Devices/10.Update-Modules/docs.md
```diff
@@ -132,12 +137,37 @@ The command line options are detailed below:
...
 Go to Artifacts in the Mender server UI and upload your newly generated Mender Artifact.
 
 Now go to Deployments and deploy the Artifact to All devices. It should finish within a minute or so.
 
+### Upload and deploy your Artifact in Standalone Mode
+
+Copy your newly generated Mender Artifact to the target device.  SSH can be used for this, i.e. scp.
+
+```bash
+scp web-file-1.mender <username>@<deviceip>:/home/<username>
```
lluiscampos: We should use the same name as above, `~/Downloads/web-file-1.mender`.

Also, if the destination directory is home, I think it is better to omit it and just write `scp ... <username>@<deviceip>:`

lluiscampos:
One more detail, the PR needs to be pointing to master. Please change that.

This is our common developement procedure. Then we will port the changes to all relevant stable branches (in this case, 2.0.x).

reachBAM:
Closing this pull request so that it can be moved to master branch.

PR #712, opened 2019-06-11
Update update modules instructions

reachBAM:
This is the same pull request as #711 but based on master branch instead of 2.0.X branch.

Added instruction to create directory /var/www prior to installing artifact.

Added description for deploying an update module in standalone mode.

### DIFF
--- a/03.Devices/10.Update-Modules/docs.md
+++ b/03.Devices/10.Update-Modules/docs.md
@@ -41,7 +41,7 @@ Refer to [further reading](#further-reading) for more details.
 
 ### Mender server
 
-You will need a Mender server.
+You will need a Mender server if you are using [managed mode](../../architecture/overview#modes-of-operation).
 
 For easy setup, use [Hosted Mender](https://hosted.mender.io?target=_blank) or the [on-premise demo server](../../getting-started/create-a-test-environment).
 
@@ -96,6 +96,11 @@ Then add execute permission to the script:
 chmod +x web-file
 ```
 
+Then create the directory where the update files will be installed:
+```bash
+mkdir -p /var/www
+```
+
 Your Mender client is now able to handle any Artifact containing a `web-file` Update Module Artifact.
 
 ### Create an Artifact with a payload for the new Update Module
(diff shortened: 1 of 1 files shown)


03.Devices/10.Update-Modules/docs.md
```diff
@@ -41,7 +41,7 @@ Refer to [further reading](#further-reading) for more details.
 
 ### Mender server
 
-You will need a Mender server.
+You will need a Mender server if you are using [managed mode](https://docs.mender.io/2.0/architecture/overview#modes-of-operation).
```
lluiscampos: This should be a relative link: `../../architecture/overview#modes-of-operation`

03.Devices/10.Update-Modules/docs.md
```diff
@@ -96,6 +96,12 @@ Then add execute permission to the script:
 chmod +x web-file
 ```
 
+Then create the directory where the update files will be installed:
+```bash
+cd /var
+mkdir www
```
lluiscampos: This will not work if the directory already exists. And I think it is cleaner if we don't modify the user's working directory.

Please replace the cd + mkdir with: `mkdir -p /var/www`

03.Devices/10.Update-Modules/docs.md
```diff
@@ -130,12 +136,26 @@ The command line options are detailed below:
...
 
+### Upload and deploy your Artifact in Standalone Mode
+
+Copy your newly generated Mender Artifact to the target device.  SSH can be used for this, i.e. scp.
+
+```bash
+scp ~/Downloads/web-file-1.mender <username>@<deviceip>:
+```
+
+Then deploy the artifact from the target device:
```
lluiscampos: We always use Artifact (capitalized).

Also, I think it is better to use "install" rather than "deploy", as here there is no actual deployment but only installation.

03.Devices/10.Update-Modules/docs.md
```diff
@@ -130,12 +136,26 @@ The command line options are detailed below:
...
+
+```bash
+mender -install web-file-1.mender
+```
```
lluiscampos: We are missing a `-commit` call after successful installation.

reachBAM: When I tested this, because the example does not have rollback enabled, mender will do an automatic commit.  Would you still like me to make it explicit?  Or I could add a note?

03.Devices/10.Update-Modules/docs.md
reachBAM: After reviewing this sentence, I would make a couple more changes actually, how about this?

"Then install the Mender Artifact on the target device:"

lluiscampos: :+1:

03.Devices/10.Update-Modules/docs.md
lluiscampos: You are right, of course. Disregard my comment, then.

03.Devices/10.Update-Modules/docs.md
reachBAM: Done.

03.Devices/10.Update-Modules/docs.md
reachBAM: Done.

03.Devices/10.Update-Modules/docs.md
reachBAM: Done.

lluiscampos:
Thanks. The changes are good to go now. 

lease squash your changes into a single commit and I'll merge it.

PR #723, opened 2019-07-05
Add sizelimit to mount for image modifying

mkozjak:
`mount` didn't work for me on Ubuntu 18.04 LTS and was complaining about overlapping loop device. Added `sizelimit` to mitigate the issue.


mkozjak:
Maybe add `sizelimit` as an optional note. On some images it worked and on some it didn't.

mirzak:
I think we need to confirm the problem before updating the docs, could you share some logs of when you are experiencing this error? e.g output of mount with `-v`, `fdisk -l` on you image and possibly `dmesg`.

PR #724, opened 2019-07-10
Fix outdated state script/standalone mode info box.

kacf:


kacf:
Cherry-picked to 2.0.x as well.

PR #725, opened 2019-07-10
docs/preauth: privkey device install

mkozjak:
To reduce the confusion about privkey device installation further, add a note for golden image editing, which can be used for automated builds that are basing their configuration on golden images.


oleorhagen:
sdimg's can be modified with the `mender-artifact cp/install` functionality ;)

PR #729, opened 2019-07-31
MENDER_DISK_IMAGE should not be in the input/ directory

drewmoseley:


drewmoseley:
No.  Just general confusion.

PR #731, opened 2019-08-01
update container name in troubleshooting section

mirzak:
mender-device-auth -> menderproduction_mender-device-auth_1.

Changelog: None


oleorhagen:
Would it not be better to add a step for looking up the name, in case someone is running a custom setup?

mirzak:
Actually, this command work for me when tried this (in demo environment),

```
docker-compose exec mender-device-auth /usr/bin/deviceauth maintenance --decommissioning-cleanup
```

Would be better if this worked, it seems that docker is able to look up the names based on something :).

Would need a comment from @kjaskiewiczz here, the problem was initially reported [here](https://hub.mender.io/t/cannot-cleanup-the-deviceauth-database/877/4)

PR #737, opened 2019-08-16
Reorganize sections to funnel users through application based

estenberg:
onboarding first, before doing system update demo.


estenberg:
Cherry-picked to 2.1.x also, after a few link fixes

PR #761, opened 2019-09-20
MEN-2739, MEN-2740: Document Enterprise installation procedure.

kacf:
Multi tenancy is also explained along the way.


kacf:
This is missing the testing parts of the documentation, which is why the tests are failing. The content can be reviewed, however.

kacf:
@estenberg: Also the use of "tenant" and "organization" is a little bit mixed. We cannot avoid the term "tenant token" for historical reasons. I've tried to use it as consistently as possible though, sticking to organization as much as possible, except when talking about "multi tenancy" and "tenant tokens".

07.Administration/02.Production-installation/01.Enterprise/docs.md
```diff
@@ -0,0 +1,308 @@
...
+### Basic configuration
+
+In this initial step we will use the Open Source guide to set up the main
+configuration. Once the Open Source parts of the server are set up, we will
+return to the Enterprise guide to enable and set up the Enterprise features of
+the server.
+
+Please proceed to [the Basic Preparation Setup of the Open Source
+guide](..#basic-preparation-setup) to start the setup. There will be a
+notification when it is time to go back to the Enterprise guide.
```
estenberg: Perhaps "note" instead of "notification"?
Also, this is potentially a bit tricky to follow (jump back and forth) but I understand the reasoning for it.

To make it more explicit, maybe something like "Please proceed to [the Basic Preparation Setup of the Open Source
guide](..#basic-preparation-setup) and come back here when you reach the [Bring it all up section in the Open Source guide](..#basic-preparation-setup); there is a note when to come back to this tutorial."

07.Administration/02.Production-installation/01.Enterprise/docs.md
```diff
@@ -0,0 +1,308 @@
+---
+title: Enterprise
+taxonomy:
+    category: docs
+---
+
+This is a step by step guide for deploying the Mender Enterprise Server for
+production environments, and will focus on the Enterprise specific parts of the
+installation.
+
```
estenberg: Would add a note: "!! This Mender Enterprise installation guide relies on `docker-compose` which is easy to set up on a single host and capable of scaling to thousands of Mender client devices. If you have higher scale and uptime requirements, please email [contact@mender.io] for assistance with a Kubernetes-based setup which does not have any scale limitations."

07.Administration/02.Production-installation/01.Enterprise/docs.md
```diff
@@ -0,0 +1,308 @@
...
+## Prerequisites
+
+- All [prerequisites for the Open Source edition](..#prerequisites) apply to the
+  Enterprise edition as well
```
estenberg: Also add that you need an enterprise account to proceed (it is implicit here, but a bit hard to know how to get one).

Perhaps something like: "You need an account with Mender in order to evaluate and use the commercial features in Mender Enterprise. Please email [contact@mender.io] to receive your credentials to log in with docker below before proceeding."

07.Administration/02.Production-installation/01.Enterprise/docs.md
```diff
@@ -0,0 +1,308 @@
...
+> ```
+
+Note that even though this launches the backend services, the Mender Enterprise
+server is not fully functional yet. Running the server is necessary for some of
+the upcoming steps.
+
+
+### Multi tenancy
+
+Multi tenancy is a feature of Mender where users can be divided into groups
```
estenberg: would remove "groups called". Groups is a different user-management concept so may be confused. :)

07.Administration/02.Production-installation/01.Enterprise/docs.md
```diff
@@ -0,0 +1,308 @@
...
+called "organizations" (also sometimes called "tenants"). Each organization is
+confined and can not see or influence the data of other organizations. This can
+be used for example to give two different product teams access to the Mender
+service, without them seeing each others data.
```
estenberg: ...data, as well as isolating devices and users in test and production environments.

07.Administration/02.Production-installation/01.Enterprise/docs.md
```diff
@@ -0,0 +1,308 @@
...
+
+When using Mender Enterprise, multi tenancy is automatically enabled, and it
+cannot be turned off. However, since multi tenancy is not desirable in all
+situations, a default organization can be specified, which makes using this one
+organization transparent. See the section on [using a default
```
estenberg: ...transparent. Use this approach if you do not plan to create multiple organizations and see the section...

07.Administration/02.Production-installation/01.Enterprise/docs.md
```diff
@@ -0,0 +1,308 @@
...
+organization](#option-1-using-a-default-organization) for more information.
+
+Below follows a guide for setting up a basic, working multi tenancy
```
estenberg: Below follows a guide for setting up a single organization. For additional ...

07.Administration/02.Production-installation/01.Enterprise/docs.md
```diff
@@ -0,0 +1,308 @@
...
+configuration. For additional information on administering organizations, see
+the [API for the tenantadm service (TODO!)](TODO), as well as the help screen
+from:
+
+```bash
+./run exec mender-tenantadm /usr/bin/tenantadm --help`
+```
+
+#### Creating the first organization and user
```
estenberg: This area becomes a bit complicated because there are two paths (single organization with default tenant OR multiple organizations).

I would recommend to restructure it to just go with the single default tenant approach first, and then have a section at the end where it is explained how to add more tenants and get rid of the default tenant.

07.Administration/02.Production-installation/01.Enterprise/docs.md
```diff
@@ -0,0 +1,308 @@
...
+service, without them seeing each others data.
+
+When using Mender Enterprise, multi tenancy is automatically enabled, and it
+cannot be turned off. However, since multi tenancy is not desirable in all
+situations, a default organization can be specified, which makes using this one
+organization transparent. See the section on [using a default
+organization](#option-1-using-a-default-organization) for more information.
+
+Below follows a guide for setting up a basic, working multi tenancy
+configuration. For additional information on administering organizations, see
```
kjaskiewiczz: @estenberg When it comes to administrating organizations only possible operations are create and suspend;
there is no way to remove the organization (HM team have some untested scripts)
should we create a task or tasks for extending the administration of organizations?

07.Administration/02.Production-installation/01.Enterprise/docs.md
kacf: The proposed solution is not so good I think. With this approach (only option 1) you need to add configuration which you later need to remove again. I think this will be perceived as strange. On top of this we actually *want* people to use option 2, both because it is more future proof and more secure, and if people do option 1 first, there is a much higher risk that they leave the default tenant token in there by mistake.

I agree it's complicated though. I propose instead to describe only option 2, with a special "Go here if you're upgrading from Open Source" link. Is that ok?

kacf:
> Aside from the comments I think we should make the Enterprise installation the default (top level). One of the tutorials (open source or enterprise) need to reference the other and this makes one of them slightly more tricky to follow. I think the Enterprise one should be the easiest to follow as it is the supported one for customers, but there are obviously tradeoffs here.

Indeed, good thinking. I actually went even further, and merged the two guides together completely, which I think made it simpler. It also means that your comments about jumping around are now moot. There are still jumps, but all of them are the skip-forward type.

I fixed the other small issues as well, and added the changes as new commits, so you can see the difference.

07.Administration/02.Production-installation/01.Enterprise/docs.md
estenberg: Ah, I recall this now. :) It would be great if you could create a delete org ticket for future reference, but I think we will have some time to fix this (probably not going to be a very popular feature in the beginning).

07.Administration/02.Production-installation/01.Enterprise/docs.md
estenberg: Yes, that makes sense! In fact, perhaps we should drop the "open source upgrade" and cover it in a different document? It seems like it is a quite different case?

Just keep in mind that the more we "redirect" the user to different parts in the tutorial or specific parts of other tutorials, the more difficult and error-prone it will be to follow. It is probably ok to do once, but if it happens several times I think most users will fail to follow it..

07.Administration/02.Production-installation/docs.md
```diff
@@ -98,45 +112,51 @@ Prepare a branch where all deployment related changes will be kept:
...
+  integration repository)
 
-```bash
-sed -i -e 's#/template/#/production/#g' prod.yml
-```
+- `enterprise.yml.template` - configuration for running an Enterprise instance
+  of the Mender server. This topic is covered separately in [the Enterprise
+  part](enterprise) of the Production installation guide
+
+!!! If an `enterprise.yml` file exists in the `config` directory, this will automatically turn on Enterprise features in the backend service. Make sure that `enterprise.yml` does not exist unless you are planning on using the Mender Enterprise server.
```
estenberg: Nitpicking but would rephrase the last sentence ("Make sure that...") to not use double negation.

E.g. If you want to use the open source edition (not Enterprise), please remove `enterprise.yml` from this directory.

07.Administration/02.Production-installation/docs.md
```diff
@@ -584,47 +605,299 @@ Bring up all services up in detached mode with the following command:
...
+subcommand of docker-compose, e.g.:
 
+<!--AUTOMATION: ignore -->
 ```bash
-./run ps
+./run exec mender-useradm /usr/bin/useradm create-user --username=myusername@host.com --password=mysecretpassword
+```
+
+! Keep in mind that above is executed in a command-line interpreter meaning that certain characters might need to be escaped, e.g if you are using the `$` character in your password, this could interpret as a variable name unless it is escaped.
+
```
estenberg: What is the next step for open source user after this? Perhaps link?

07.Administration/02.Production-installation/docs.md
```diff
@@ -584,47 +605,299 @@ Bring up all services up in detached mode with the following command:
...
+```
+
+Replace the organization name, username and password with desired values. Make
+sure to save the **tenant ID** that appears after calling the command; this will
+be the identifier for the first organization. Creating additional organizations
+works exactly the same way, so the above step may be repeated multiple times.
+
+Now fetch the **tenant token** for the new organization by calling:
+
+```
```
estenberg: I think I would remove this and only do the version with jq/python to parse it. It is more convenient and we use it other places in the docs (or the python -m json.tool).

In any case I think we should just do it once (not one without parsing and one with parsing) so it does not look like you have to do this twice initially.

07.Administration/02.Production-installation/docs.md
```diff
@@ -584,47 +605,299 @@ Bring up all services up in detached mode with the following command:
...
+tFY0kJnhGtN07CD3XXxcz0BpWbevDYOPOEUusGd2KpLK2Y4QU8RdngqNtRe7SppG0fn6m6tKiifrPDAv_THCEG6dvpMHyIM77vGI<br>
+PwvV4ABGhZKRAlDe1R4csJQIbhVcTWMGcoZ4bKH-zDK0900_wWM53i9rkgNFDM470i6-d1oqfaCPcbiniKsq1HcJRZsIVNJ1edDo<br>
+vhQ6IbffPRCw-Au_GlnPTn_czovJqSa3bgwrJguYRIKJGWhHgx0e3j795oJ08ks2Mp3Rshbv4da
+
+!!! On Debian based distributions you can install `jq` with the command `apt-get
+!!! install jq`.
+
+The tenant token will be used the following sections.
+
+#### Choosing organization configuration
```
estenberg: Nitpick, but perhaps we should swap what we call option 1 and 2 here.
I would expect that "option 1" in general would be the "most common" choice. If that choice is to configure tenants, then I think we should call that "option 1".

07.Administration/02.Production-installation/docs.md
```diff
@@ -28,6 +28,20 @@ repository.
   own the domains `mender.example.com` and `s3.example.com`) that resolve to the public
   IP of current host on the devices.
 
+If you are setting up a Mender Enterprise server, you will also need:
+
+- An account with Mender in order to evaluate and use the commercial features in
+  Mender Enterprise. Please email [contact@mender.io](mailto:contact@mender.io)
```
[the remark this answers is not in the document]
kacf: I tested it, works! :-)

AFAICS there are no other email links in the entire documentation, only in the README.md.

07.Administration/02.Production-installation/docs.md
kacf: Yes, makes sense. The upgrade instructions aren't completely clear to me, and are anyway handled separately in [MEN-2744](https://tracker.mender.io/browse/MEN-2744), so I think I will just remove the entire optional part for now, and tackle that later.

07.Administration/02.Production-installation/01.Enterprise/docs.md
kacf: Yes, now covered in [MEN-2744](https://tracker.mender.io/browse/MEN-2744).

kacf:
Updated and finally passed all tests, including Enterprise tests!

Leaving it open, since we need to merge it together with the feature. Creating dedicated merging task in [MEN-2804](https://tracker.mender.io/browse/MEN-2804) so that we can close this.

PR #772, opened 2019-10-08
Update compatibility table for Mender 2.2.x.

kacf:


02.Architecture/03.Compatibility/docs.md
```diff
@@ -90,6 +92,7 @@ The [Mender Artifact format](../mender-artifacts) is managed by the [Mender Arti
 | Mender 1.7.x / mender-artifact 2.4.x | yes         | yes         | no          |
 | Mender 2.0.x / mender-artifact 3.0.x | no          | yes         | yes         |
 | Mender 2.1.x / mender-artifact 3.1.x | no          | yes         | yes         |
+| Mender 2.1.x / mender-artifact 3.2.x | no          | yes         | yes         |
```
lluiscampos: From your change I assume that the table lists mender client versions, however the title seems to indicate that this is mender server version: `Mender client/server and Artifact format`. Should we change the section title? Or should we actually have three versions listed in the table?

kacf: You're right, my mistake. I don't think we should have three versions listed.

PR #776, opened 2019-10-10
Add upgrading yocto notes

lluiscampos:


lluiscampos:
Hi @drewmoseley and @estenberg,

This PR brings documentation on the know issues when upgrading from yocto thud to yocto warrior. Please have a look.

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
```diff
@@ -0,0 +1,47 @@
+---
+title: Upgrading
+taxonomy:
+    category: docs
+---
+
+This section describes the cases where upgrading from one Yocto release to the
+next requires special attention
```
mirzak: Missing `.`

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
```diff
@@ -0,0 +1,47 @@
...
+
+Note that `mender-migrate-configuration` recipe uses an State Script, and it
+might be needed to clean the yocto build after removing it. See note at [State
+Scripts](../../artifacts/state-scripts#including-state-scripts-in-artifacts-and-disk-images)
+
+* Permanently disable this feature.
+
+If it is not desired to use this feature, it can be disabled by adding
+`PACKAGECONFIG_remove = "split-mender-config"` to local.conf. Note that with
+this option the feature will need to be disable for all future builds.
```
mirzak: I would drop this. This could mess up things down the line if we implement changes that rely on that `split-mender-config` is used.

lluiscampos:
Thanks for your comments @mirzak. I applied most of your suggestions and pushed. There is a pending commit that I leave open for other to give their opinion.

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
drewmoseley: Yes, let's document it as an interim workaround and indicate that they should re-enable after migrating.  Eventually, we can remove the option entirely.

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
```diff
@@ -0,0 +1,54 @@
...
+This section describes the cases where upgrading from one Yocto release to the
+next requires special attention.
+
+## Upgrades with incompatible changes
+
+<!--AUTOVERSION: "From % to %"/ignore-->
+### From thud to warrior
+
+<!--AUTOVERSION: "meta-mender % branch"/ignore-->
+meta-mender warrior branch introduces a change for the configuration of Mender.
```
drewmoseley: I would say "The meta-mender warrior branch..."

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
```diff
@@ -0,0 +1,54 @@
...
+
+A device running on a single configuration file cannot upgrade to an image built
+with two configuration files feature.
+
+There are two possibilities:
+
+* Create an update with migration steps.
+
+This method is preferred as it will let the device(s) take advantage of this
+future in following updates.
```
drewmoseley: take advantage of this _feature_ in _future_ updates.

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
```diff
@@ -0,0 +1,54 @@
...
+
+```bash
+IMAGE_INSTALL_append = " mender-migrate-configuration"
+PACKAGECONFIG_remove = "split-mender-config"
+MENDER_PERSISTENT_CONFIGURATION_VARS = "RootfsPartA RootfsPartB"
+MENDER_ARTIFACT_EXTRA_ARGS_append = " -v 2"
+```
+
+Build an image with above configuration and deploy it your device fleet. Once
+all of you devices in the fleet have been updated with the migration script
```
drewmoseley: once all devices in the fleet...

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
```diff
@@ -0,0 +1,54 @@
...
+enabled you can remove these changes and return to the normal workflow of
+generating update Artifacts.
+
+Note that `mender-migrate-configuration` recipe uses an State Script, and it
```
drewmoseley: recipe uses _a_ State Script.

although I'm not sure if we should capitalize State Script; do we do that elsewhere?

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
lluiscampos: Good catch

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
lluiscampos: You are right, we don't capitalize them elsewhere.

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
lluiscampos: I don't understand your commend, @drewmoseley . Are you saying that sooner or later they have to migrate? Then I think the best is to just remove this text.

lluiscampos:
Modified with all your wording changes suggestions, @drewmoseley

estenberg:
I can't comment much on the content, except I would include any error messages seen so it is easy to find if you google the errors (which I think is the first thing people do).

Agree we should document this somewhere, but I do not think this is the right place to put it. The "Artifact" section is just describing general purpose features, not diving into specific version of OSes/build systems/Mender.

I think we should rather put it in the troubleshooting section as this is an "ephemeral" problem: https://docs.mender.io/2.1/troubleshooting, concretely "Upgrading Yocto Project releases" or something like that. In there we can have sections from thud -> warrior, etc.

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
drewmoseley: Yes, I think we should remove this text.  Eventually we may want to remove the ability to remove split-mender-config to avoid having yet-another-variable.  But that will obviously need to wait a while.

kacf:
Agree with both the last comments: We should remove the permanent "remove split-mender-config" solution, as well as move the instructions into the troubleshooting section. I think we should still link to it though. Since Lluis is on PTO I updated and resubmitted in #777. See if you agree.

04.Artifacts/01.Yocto-project/04.Upgrading/docs.md
lluiscampos: Understood! Fixed in #777

lluiscampos:
Superseded by #777

PR #784, opened 2019-10-11
Prod enterprise install simplifications

estenberg:
Simplified prod install instructions.

A bit unsure if the git log changes will make tests fail..


estenberg:
Passed :)

A bit tight on time so will merge. Any review would be welcome after the fact though.

PR #811, opened 2019-11-21
Fix typo - escaped '%' character.

alfrunes:
changelog: none


alfrunes:
I added an explicit check for python version to handle the changes in escaped characters for `re.escape` for python >= 3.7. Once we upgrade the python version in Travis this check may be removed.

PR #786, opened 2019-10-11
Make TENANT_ID go directly into a shell variable to avoid copy/paste.

estenberg:

PR #813, opened 2019-11-21
For production installation, instead of ignoring test automation

estenberg:
and doing it with different commands, integrate it into the doc.

Ref comments by @kacf in https://github.com/mendersoftware/mender-docs/pull/786


estenberg:
Holy moly, build passed.
Should the ENT build run as well? It has some extra steps.

kacf:
> Should the ENT build run as well? It has some extra steps.

For future reference, the Enterprise test will run on master, or if you submit a pull request using a branch in this repository, and not your private repository. Master passed the last build, so this should be good.

PR #823, opened 2019-12-06
Add changelog for mender-binary-delta 1.0.1

lluiscampos:


lluiscampos:
@oleorhagen @kack see below:
- Does it make sense to have statistics on a closed source repo?
- There is no mender-binary-delta 1.0.0 in the release notes changelogs. Should we add it with an empty changelog?
- @oleorhagen I generated the changelog manually as the original commit had no Changelog tag. Please double check the text.

202.Release-information/01.Release-notes-changelog/docs.md
```diff
@@ -4,6 +4,48 @@ taxonomy:
...
+
+#### mender-binary-delta (1.0.1)
+
+New changes in mender-binary-delta since 1.0.0:
+
+* MEN-2928: Fix: Enable file-systems larger than approx 2-GiGs
+  ([MEN-2928](https://tracker.mender.io/browse/MEN-2928))
+
+
 ## Mender 2.2.1
```
lluiscampos: Missing date here! I will fix.

oleorhagen:
My 2 cents, but just letting my thoughts flow here, so take it at face value.

Originally I figured this did not need a changelog, with the closed nature of the tool. However, we have been discussing open-sourcing it at some point in the future (maybe/maybe not). However, if we are going to, then a changelog might be appropriate.

Text looks fine

lluiscampos:
> My 2 cents, but just letting my thoughts flow here, so take it at face value.
> 
> Originally I figured this did not need a changelog, with the closed nature of the tool. However, we have been discussing open-sourcing it at some point in the future (maybe/maybe not). However, if we are going to, then a changelog might be appropriate.
> 
> Text looks fine

Well, I think we need a changelog regardless of the license. If any, in a closed source deliverable is more important to have a changelog as you have no ways to tell what the changes are :)

lluiscampos:
>     * Does it make sense to have statistics on a closed source repo?

Discussed privately with @kacf. It does not make much sense so I will remove it

kacf:
> There is no mender-binary-delta 1.0.0 in the release notes changelogs. Should we add it with an empty changelog?

I suppose you could add it with a "First revision" comment.

lluiscampos:
> > There is no mender-binary-delta 1.0.0 in the release notes changelogs. Should we add it with an empty changelog?
> 
> I suppose you could add it with a "First revision" comment.

Thanks. Added

PR #836, opened 2019-12-19
Fix broken link to mender-convert config file

michaelatmender:
Changelog: None

Signed-off-by: Michael Clelland <michael.clelland@mender.io>

### DIFF
--- a/04.Artifacts/02.Debian-family/02.image-configuration/docs.md
+++ b/04.Artifacts/02.Debian-family/02.image-configuration/docs.md
@@ -14,7 +14,7 @@ which add support for board-specific configurations. A run of `mender-convert`
 can include multiple configuration files, each one added with the `--config`
 command-line option. The standard configuration which will always be included
 can be found in the
-[configs/mender_convert_config](https://github.com/mendersoftware/mender-convert/blob/next/configs/mender_convert_config)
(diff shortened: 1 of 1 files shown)


michaelatmender:
Autoversion test is failing, I'm unsure what to add to fix it. I don't see autoversion documented anywhere either.

04.Artifacts/02.Debian-family/02.image-configuration/docs.md
```diff
@@ -14,7 +14,7 @@ which add support for board-specific configurations. A run of `mender-convert`
 can include multiple configuration files, each one added with the `--config`
 command-line option. The standard configuration which will always be included
 can be found in the
-[configs/mender_convert_config](https://github.com/mendersoftware/mender-convert/blob/next/configs/mender_convert_config)
+[configs/mender_convert_config](https://github.com/mendersoftware/mender-convert/blob/master/configs/mender_convert_config)
```
lluiscampos: You need to add something like:
```
<!--AUTOVERSION: "blob/%/config"/ignore-->
```

And you can always check if it is correct by locally executing: `./autoversion.py --check`

lluiscampos: Wait, not really "igore" but actually mender-convert. So it should be:
```
<!--AUTOVERSION: "blob/%/config"/mender-convert-->
```

PR #839, opened 2020-01-07
GitLab Pipeline beautification: indentation and order

lluiscampos:


oleorhagen:
Mby we should consider an editorconf like configuration setup?

PR #843, opened 2020-01-09
QA-139 debug logs to see which container does not start

merlin-northern:
ChangeLog:title


lluiscampos:
@merlin-northern This PR revealed an error which we fixed then. From my point of view, there is nothing else to do here. Can we close it?

merlin-northern:
of course, this was debugging. closing.

PR #849, opened 2020-01-22
yocto: Added info about MENDER_EXTRA_PARTS variable

nandra:
Signed-off-by: Marek Belisko <marek.belisko@open-nandra.com>


kacf:
Thanks @nandra! I made a couple of cleanups and resubmitted in #851.

PR #850, opened 2020-01-22
MEN-3060 Logging improvements

merlin-northern:
* adds logs when there is a problem with startup of docker containers in AUTOMATION blocks
 * for "Open Source instance" case we have 14 services now

ChangeLog:none


merlin-northern:
@lluiscampos @tranchitella I cant reproduce MEN-3060 locally in any way, could we try to merge it, so we get the logs in case of error? also I think we now have 14 services in case of "Open Source instance"

merlin-northern:
for detail see MEN-3060 closing this one.

PR #851, opened 2020-01-23
Resubmission of #849

kacf:
Resubmission of #849


kacf:
The failure is due to a known and unrelated problem, and this PR is not touching that path. Merging.

PR #857, opened 2020-01-24
Documentation for creating artifact from snapshot

alfrunes:
Describe how to generate a snapshot, both on the client and using `mender-artifact`.


lluiscampos:
Is this ready for review now? Could you please fix the conflicts first?

I want to start with another doc task, and it will be the easiest to start on top of your work (due to the multiple files renamed)

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Artifact from system snapshot
```
oleorhagen: Should some part of it, or at least a link, be provided in the getting-started section also?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,87 @@
...
+---
+
+After provisioning the first batch of devices it is sometimes useful to have a
+device to tinker with - e.g reconfigure, test and install new packages, before
+deploying to the rest of the fleet. When satisfied with the rootfs, the mender
+binary provides an easy way to create a  snapshot of the root filesystem. The
+snapshot feature is also available within the 
+[mender-artifact](../../08.Downloads/docs.md#mender-artifact) CLI tool.
+
+! This feature is available from version 2.3 and fully supported only by the
```
oleorhagen: I think I would skip the `only` part

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,87 @@
...
+! following filesystems: ext4, ext3, XFS, JFS, btrfs, f2fs, and ReiserFS. If
+! your device runs on a different filesystem a workaround is described 
+! [below](#snapshot-a-specific-partition). 
+
+## Creating a snapshot on the device
+The *mender* binary has a command `mender snapshot dump` for dumping a frozen
```
oleorhagen: s/mender/Mender

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,87 @@
...
+copy of the root filesystem to standard error. In the following, two approaches
```
oleorhagen: The following what?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,87 @@
...
+for using this command is described.
+
+!!! While a snapshot is in progress, all processes that writes to the filesystem
+!!! being backed up will be blocked until the snapshot is complete. If the output
+!!! of `mender snapshot dump` is redirected to a process that writes to the
```
oleorhagen: Nice!

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,87 @@
...
+mounted device.
+```bash
+mount /dev/(...) /mnt
+mender snapshot dump > /mnt/root-part.fs
+```
+
+!!! The snapshot command also accepts a `--compression` parameter to save
+!!! storage/bandwidth for the target output.
+
+In this case, `root-part.fs` can be passed to the
```
oleorhagen: `mender-artifact tool` mby?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,87 @@
...
+                                   -t device-type
+```
+Will yield the exact same artifact as above.
+
+## Snapshot a specific partition
+It is also possible to specify the partition to capture using the `--fs-path` <!--- FIXME: or source? -->
+flag. The path must point to either a file/directory on the target filesystem or
+to the device file (other special filetypes will throw an error). To illustrate,
+consider a device with a filesystem which does not implement the ioctl freeze 
+operation (see supported filesystem [above](#artifact-from-system-snapshot)). As
```
oleorhagen: s/filesystem/filesystems/

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,87 @@
...
+inactive partition and perform the snapshot from there. That is,
+```bash
+# Ready to snapshot partition
+mender rollback
+# <rebooting...>
+mender snapshot dump --fs-path /dev/<insert inactive partition device here> | \
+                ssh user@host /bin/sh -c 'cat > snapshot.fs'
+
+```
+!!! Make sure the inactive partition is running Mender version >= 2.3
```
oleorhagen: Suggestion: 
This will only work on an inactive partition with Mender version >= 2.3

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,87 @@
...
+device to tinker with - e.g reconfigure, test and install new packages, before
+deploying to the rest of the fleet. When satisfied with the rootfs, the mender
+binary provides an easy way to create a  snapshot of the root filesystem. The
+snapshot feature is also available within the 
+[mender-artifact](../../08.Downloads/docs.md#mender-artifact) CLI tool.
+
+! This feature is available from version 2.3 and fully supported only by the
+! following filesystems: ext4, ext3, XFS, JFS, btrfs, f2fs, and ReiserFS. If
+! your device runs on a different filesystem a workaround is described 
+! [below](#snapshot-a-specific-partition).
```
oleorhagen: Suggestion: Name the section, instead of the below link

04.Artifacts/22.Snapshots/docs.md
alfrunes: Yeah, good suggestion! I added a reference in the "Deploy a system update demo" section.

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,89 @@
...
+device to tinker with - e.g reconfigure, test and install new packages, before
+deploying to the rest of the fleet. When satisfied with the rootfs, the mender
+binary provides an easy way to create a  snapshot of the root filesystem. The
+snapshot feature is also available within the 
+[mender-artifact](../../08.Downloads/docs.md#mender-artifact) CLI tool.
+
+!!!!!! This feature is available from version 2.3 and fully supported by the
+!!!!!! following filesystems: ext4, ext3, XFS, JFS, btrfs, f2fs, and ReiserFS.
+!!!!!! If your device runs on a different filesystem a workaround is described 
+!!!!!! in the [Snapshot a partition](#snapshot-a-partition) section below.
```
kacf: I haven't seen the six-exclamation-points notation before, but maybe I'm just unaware. What does this notation produce?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,89 @@
...
+## Creating a snapshot on the device
+The *Mender* binary has a command `mender snapshot dump` for dumping a frozen
+copy of the root filesystem to standard error. In this section, two approaches
+are described: dumping to a remote host or to external storage.
+
+!!! While a snapshot is in progress, all processes that writes to the filesystem
+!!! being backed up will be blocked until the snapshot is complete. If the output
+!!! of `mender snapshot dump` is redirected to a process that writes to the
+!!! filesystem, the process will freeze for a short while until an internal timer
+!!! expires and unfreezes the filesystem.
```
kacf: The implementation to detect this is impressive, but I think in the docs it is too implementation focused. I would rephrase this to:

> Redirecting the output from `mender snapshot dump` to a process that writes to the **same** filesystem is not supported, and will terminate the snapshot early.

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,89 @@
...
+
+If the device has `ssh` installed and a host is reachable and has an ssh server 
+running, the output from the command can be redirected to standard an ssh 
+session as follows:
+```bash
+mender snapshot dump | ssh user@host-ip /bin/sh -c 'cat > $HOME/root-part.fs`
+```
+
+If a USB stick is available, simply redirecting the output to a file on the
+mounted device.
```
kacf: Small grammar issue: If a USB stick is available, **one can** simply **redirect** the output to a file on the mounted device.

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,89 @@
...
+```bash
+mount /dev/(...) /mnt
+mender snapshot dump > /mnt/root-part.fs
+```
+
+!!! The snapshot command also accepts a `--compression` parameter to save
+!!! storage/bandwidth for the target output.
```
kacf: Maybe this should be a normal paragraph with an example?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,89 @@
...
+mender-artifact write rootfs-image -f /mnt/root-part.fs \
+                                   -n artifact-name \
+                                   -o snapshot-release.1.0.mender \
+                                   -t device-type
+```
+Uploading this artifact to the mender server and creating a deployment ensures
+that all devices runs an identical rootfs version as the snapshot host.
+
+## Creating a snapshot from a host machine
+As already mentioned there is support to create a snapshot artifact directly
```
kacf: "...there is support **for creating**..."

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,89 @@
...
+is reachable and has *ssh* and *sudo* installed. *mender-artifact* accepts a
+file URL with ssh schema and will automatically try to run the commands above
+on the remote device. For example, the command:
+```bash
+mender-artifact write rootfs-image -f ssh://user@device-ip:port \
+                                   -n artifact-name \
+                                   -o snapshot-release.1.0.mender \
+                                   -t device-type
+```
+Will yield the exact same artifact as above.
```
kacf: Nice sentence, this gives confidence!

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,89 @@
...
+It is also possible to specify the partition to capture using the `--fs-path` <!--- FIXME: or source? -->
+flag. The path must point to either a file/directory on the target filesystem or
+to the device file (other special filetypes will throw an error). To illustrate,
+consider a device with a filesystem which does not implement the ioctl freeze 
+operation (see supported filesystems [above](#artifact-from-system-snapshot)). As
+an alternative to running snapshot from the active partition, rollback to the
+inactive partition and perform the snapshot from there. That is,
+```bash
+# Ready to snapshot partition
+mender rollback
```
kacf: I'm a bit concerned about this step: What will it roll back to? If you just execute this command out of nowhere, you will get an error. We discussed this a bit on Slack, IMHO I think we can just remove the docs for the `fs-path` flag. It's not something most users will use, only advanced users, and they will probably be checking the help screen.

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,89 @@
...
+---
+
+After provisioning the first batch of devices it is sometimes useful to have a
+device to tinker with - e.g reconfigure, test and install new packages, before
+deploying to the rest of the fleet. When satisfied with the rootfs, the mender
+binary provides an easy way to create a  snapshot of the root filesystem. The
+snapshot feature is also available within the 
+[mender-artifact](../../08.Downloads/docs.md#mender-artifact) CLI tool.
+
+!!!!!! This feature is available from version 2.3 and fully supported by the
```
lluiscampos: This 2.3 reference will require an AUTOREV tag. It seems like the Pipeline did not run for this PR, so let's make sure it runs before merging.

04.Artifacts/22.Snapshots/docs.md
alfrunes: A "mender-purple" text box ;)
I actually think it goes up to 8 different colors.

04.Artifacts/22.Snapshots/docs.md
alfrunes: Sorry, I'm not sure I'm following. What AUTOREV tag?

04.Artifacts/22.Snapshots/docs.md
alfrunes: On a second thought... I don't think it contributes to the on-boarding process yet. We'll have to create an implementation for generating a full image from partition snapshots first.

04.Artifacts/22.Snapshots/docs.md
alfrunes: I just removed it - it doesn't really contribute anything either. Besides, you can achieve the exact same result using other utilities that are most likely available: `cat /dev/<part> [| gzip] | ssh ...`.

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Artifact from system snapshot
+taxonomy:
+    category: docs
+---
+
+After provisioning the first batch of devices it is sometimes useful to have a
```
estenberg: I would start the intro a little differently, as this relates to a established workflow often referred to as the "Golden image".

Suggestion: "To support an easy golden image workflow, Mender supports creating a snapshot of the currently running system. In this workflow a "golden image" is maintained on a single device (or SD card) by making run-time modifications on the device, such as installing packages, changing configurations and updating the operating system. When this device has been tested and the environment should be replicated, a snapshot can be taken with Mender. This results in a file system image and Mender Artifact that can be deployed en masse in the field.

The Mender binary provides ..."

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+deploying to the rest of the fleet. When satisfied with the rootfs, the mender
+binary provides an easy way to create a  snapshot of the root filesystem. The
+snapshot feature is also available within the 
+[mender-artifact](../../08.Downloads/docs.md#mender-artifact) CLI tool.
+
+!!!!!! This feature is available from version 2.3 and fully supported by the
+!!!!!! following filesystems: ext[234], XFS, JFS, btrfs, f2fs, and ReiserFS.
+
+## Creating a snapshot on the device
+The *Mender* binary has a command `mender snapshot dump` for dumping a frozen
```
estenberg: Mender binary -> Mender client ?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+device to tinker with - e.g reconfigure, test and install new packages, before
+deploying to the rest of the fleet. When satisfied with the rootfs, the mender
+binary provides an easy way to create a  snapshot of the root filesystem. The
+snapshot feature is also available within the 
+[mender-artifact](../../08.Downloads/docs.md#mender-artifact) CLI tool.
+
+!!!!!! This feature is available from version 2.3 and fully supported by the
+!!!!!! following filesystems: ext[234], XFS, JFS, btrfs, f2fs, and ReiserFS.
+
+## Creating a snapshot on the device
```
estenberg: "on" -> "from" ?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+If a USB stick is available, one can simply redirect the output to a file on the
+mounted device.
+```bash
+mount /dev/(...) /mnt
+mender snapshot dump > /mnt/root-part.fs
+```
+
+A `--compression` flag is also available to save storage space and/or bandwidth
```
estenberg: "and/or" -> "and"

estenberg: (this is pretty nice, btw! did anyone test performance of this, if the device is fast enough to do the compression to not slow down the overall operation?)

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+In this case, `root-part.fs` can be passed as the file parameter to
+[mender-artifact](../../08.Downloads/docs.md#mender-artifact):
+```bash
+mender-artifact write rootfs-image -f /mnt/root-part.fs \
+                                   -n artifact-name \
+                                   -o snapshot-release.1.0.mender \
+                                   -t device-type
+```
+Uploading this artifact to the mender server and creating a deployment ensures
+that all devices runs an identical rootfs version as the snapshot host.
```
estenberg: "snaphost host" -> "golden device"

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Creating a snapshot from a host machine
+As already mentioned there is support for creating a snapshot artifact directly
```
estenberg: Would remove "As already mentioned"

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+from a host machine. This approach requires that the snapshot target device
```
estenberg: "host machine" -> "workstation, such as a laptop".

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+```bash
+mender-artifact write rootfs-image -f /mnt/root-part.fs \
+                                   -n artifact-name \
+                                   -o snapshot-release.1.0.mender \
+                                   -t device-type
+```
+Uploading this artifact to the mender server and creating a deployment ensures
+that all devices runs an identical rootfs version as the snapshot host.
+
+## Creating a snapshot from a host machine
```
estenberg: Perhaps "Creating a snapshot from a workstation"

04.Artifacts/22.Snapshots/docs.md
estenberg: "snapshot target device" -> "golden device"

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+As already mentioned there is support for creating a snapshot artifact directly
+from a host machine. This approach requires that the snapshot target device 
+is reachable and has *ssh* and *sudo* installed. *mender-artifact* accepts a
```
estenberg: *mender-artifact* -> `mender-artifact`

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+file URL with ssh schema and will automatically try to run the commands above
+on the remote device. For example, the command:
```
estenberg: "remote device" -> "golden device"

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+                                   -t device-type
+```
+Uploading this artifact to the mender server and creating a deployment ensures
+that all devices runs an identical rootfs version as the snapshot host.
+
+## Creating a snapshot from a host machine
+As already mentioned there is support for creating a snapshot artifact directly
+from a host machine. This approach requires that the snapshot target device 
+is reachable and has *ssh* and *sudo* installed. *mender-artifact* accepts a
+file URL with ssh schema and will automatically try to run the commands above
```
estenberg: would remove "try to"

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+on the remote device. For example, the command:
+```bash
+mender-artifact write rootfs-image -f ssh://user@device-ip:port \
```
estenberg: Would make this a bit more copy-paste friendly. In other parts of the docs we do something like thie following.

To snapshot a golden device from your workstation, run the following commands:
```bash
SSH_USER="user" # change to device user name
SSH_ADDRESS="device-ip:port"


mender-artifact write rootfs-image -f ssh://user@device-ip:port \
                                   -n artifact-name \
                                   -o snapshot-release.1.0.mender \
                                   -t device-type
```

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+of the filesystem can be produced by
+```bash
+mount /dev/(...) /mnt
+mender snapshot dump --compression gzip > /mnt/root-part.fs.gz
+```
+
+In this case, `root-part.fs` can be passed as the file parameter to
+[mender-artifact](../../08.Downloads/docs.md#mender-artifact):
+```bash
+mender-artifact write rootfs-image -f /mnt/root-part.fs \
```
estenberg: Also see note about copy-past-ability below. :)

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,72 @@
...
+
+A `--compression` flag is also available to save storage space and/or bandwidth
+on the target storage device. For the example above, a gziped compressed version
+of the filesystem can be produced by
+```bash
+mount /dev/(...) /mnt
+mender snapshot dump --compression gzip > /mnt/root-part.fs.gz
+```
+
+In this case, `root-part.fs` can be passed as the file parameter to
```
estenberg: I strongly recommend active voice in general to make docs more readable and clear. Take a look at the difference here https://www.linuxjournal.com/author/authguide - see the section "Passive Voice Should Never Be Used by You".

04.Artifacts/22.Snapshots/docs.md
estenberg: I would move this first, this will be the easiest and main way to do this, I think.

estenberg:
Nice @alfrunes ! I commented on quite a few small things, to align with the style of other parts of the docs. Hope it helps and it's not too much. :)

04.Artifacts/22.Snapshots/docs.md
lluiscampos: The most important is to run the Pipeline. I don't know really why is not running: can it be due to GitHub detecting a conflict? Can you fix that first? If this does not run then, you we'll need to open a new PR. I have seen this unexpected behavior in Travis before, sorry for the inconvenience.

With regards to the AUTOREV. All version-like strings require an AUTOREV tag. We have logic in the docs to detect these so that we we do a release we can for example replace "master" with "1.2.3". Then, there are cases when we want a release like string to stay the same (like this one, as we want to point to a specific version when this was releases), and you need to instruct the tool to ignore it. Learn more about it in the docs [here](https://github.com/mendersoftware/mender-docs/blob/master/README-autoversion.markdown).

lluiscampos: > ... can it be due to GitHub detecting a conflict? Can you fix that first? 

I think this is the problem, Travis reports `GitHub payload is missing` in all build attempts for this PR.

04.Artifacts/22.Snapshots/docs.md
alfrunes: I compared compression vs raw for ssh and using a USB stick on a Nano Pi. For ssh using compression actually speeds up the performance by 30-40% (i didn't bother quantifying transfer rates, packet loss etc.). However, using a USB stick the compression overhead was much more noticeable, slowing down the transfer by 4-5x.

04.Artifacts/22.Snapshots/docs.md
alfrunes: I changed the heading to "Creating a snapshot using mender-artifact CLI" as it's more direct. I'll keep the order as is for now, if that's alright?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+If `ssh` is not available, you can attach a removable storage device (e.g.
+USB stick) and redirect the output to a file on the mounted device.
+```bash
+mount /dev/(...) /mnt
+mender snapshot dump > /mnt/root-part.fs
+```
+
+! Make sure you have enough storage space on the removable device creating the
```
estenberg: How much is needed? Max size of rootfs? How can I tell what the size of rootfs is?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,86 @@
...
+frozen copy of the root filesystem to standard error. In this section, we go
+through two common approaches for using the snapshot feature from the device:
+dumping a snapshot to a remote host or to a storage device.
+
+!!! While a snapshot is in progress, all processes that writes to the filesystem
+!!! being backed up will be blocked until the snapshot is complete. Redirecting
+!!! the output to the same filesystem will cause the process to freeze and
+!!! terminate after a short while.
+
+If you've installed `ssh` on the device and have set up a computer with the ssh
```
estenberg: This is a long sentence and I'm not sure if the use of ";" and "then," is grammatically correct , split into two? Perhaps run a spell/grammar check on the entire doc to catch any issues.

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,86 @@
...
+USB stick) and redirect the output to a file on the mounted device.
+```bash
+mount /dev/(...) /mnt
+mender snapshot dump > /mnt/root-part.fs
+```
+
+! Make sure you have enough storage space on the removable device creating the
+! snapshot.
+
+To help save storage space and bandwidth for the target storage, a built-in
```
estenberg: Remove " for the target storage" ?

04.Artifacts/22.Snapshots/docs.md
```diff
@@ -0,0 +1,86 @@
...
+`--compression` option is available. For the example above, a gzip-compressed
+version of the filesystem is produced by passing gzip to the `--compression` 
+flag.
+```bash
+mount /dev/(...) /mnt
+mender snapshot dump --compression gzip > /mnt/root-part.fs.gz
```
estenberg: Is ".fs" a common suffix? If not, perhaps use ".ext4" or something that is more common?

04.Artifacts/22.Snapshots/docs.md
estenberg: Sure!

04.Artifacts/22.Snapshots/docs.md
kacf: Actually, in this particular case, the `AUTOVERSION` should be set to `ignore`. This is because we don't want to update this once we release the next version. It should be stable at this specific version forever.

Additionally, this is available since **2.2**, not 2.3. This is because the client is one minor number behind the main Mender release. Confusing? Yes, a bit, but that's a separate discussion.

lluiscampos: @alfrunes You were right: a `1.2` string (with no bugfix number) is not recognized as a version for by the tool, so as it is now you don't need the AUTOREV tag. I will try to fix this independently of this PR.

Anyway @kacf point stands, it should be `2.2` in this case.

kacf: > You were right: a 1.2 string (with no bugfix number) is not recognized as a version for by the tool, so as it is now you don't need the AUTOREV tag. I will try to fix this independently of this PR.

This is intentional. Only three-component version strings identify unique versions, so the tool only deals with those. I can see uses for dealing with two-component version strings as well, but then we would also have to replace those as two components, which is an additional complication to the syntax. My vote would be to keep ignoring these, unless you think there are good reasons not to.

lluiscampos: I actually made a simple change in the regex to chase these, and there are like 100+ two-component "versions". I don't think we should go that way and I vote to leave it as is.

Conclusion: @alfrunes just use **2.2** and remove the AUTOREV tag. Then we can merge!

alfrunes: Done! :+1:

lluiscampos: It still says 2.3 there.

lluiscampos: Good to go now :+1:

lluiscampos:
Good work. Merging!

PR #864, opened 2020-02-11
Use prebuilt image in quickstart with Mender 2.3.

estenberg:
We need to highlight:
* Prebuilt Raspbian image
* Snapshot feature


estenberg:
FYI @michaelatmender @mzedel @lluiscampos @alfrunes - we seem to have "forgotten" to include the Raspbian image and snapshot feature in the quickstart.

This is what everyone reads to evaluate Mender so this omission would render much of the improvements hidden.

I think there in general is a problem that nobody is taking care of the documentation at a higher level, thinks about the user journey and improves it, but this is a bit bigger discussion.

PR #866, opened 2020-02-11
Spelling updates

drewmoseley:

### DIFF
--- a/.ispell_english
+++ b/.ispell_english
@@ -77,6 +77,7 @@ Alf
 ali
 alitaheri
 Alman
+alt
 altbootcmd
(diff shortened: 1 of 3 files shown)


.ispell_english
```diff
@@ -1269,6 +1272,7 @@ loadaddr
 loadimage
 localconf
 localhost
+localhost�
```
kacf: I see several of these, so I guess it's correct, but what is the last character?

drewmoseley: It's actually 'localhost<E2>'.  github must be interpreting it as markdown.

PR #868, opened 2020-02-12
Fix wrong device type in command (our image ships with "raspberrypi3")

kacf:


kacf:
Already reviewed in #859.

PR #875, opened 2020-02-18
Fix strange rendering bug because of a single space.

kacf:
One space at the end of the line completely screwed up the remainder
of the page.


kacf:
Verified working and cherry-picked to 2.3.x.

PR #891, opened 2020-03-02
Add troubleshooting guide for when U-Boot has known broken UEFI support.

kacf:


kacf:
This goes together with https://github.com/mendersoftware/meta-mender/pull/931.

PR #892, opened 2020-03-02
Increase service count after mender-workflows-worker was added.

kacf:


kacf:
Beat you to it! ;-)

Lucky Luke

PR #897, opened 2020-03-05
Release 2.2.2 into 2.2.x

lluiscampos:


202.Release-information/02.Open-source-licenses/docs.md
```diff
@@ -775,37 +775,6 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
---------------------------------------------------------------------------------
```
lluiscampos: ping @mzedel 

Shouldn't the licenses have been the exact same?

I added here what the tool produced, but checking at the `gui` git history for this bugfix release there were not dependencies updates. This made me think that the gui is using the "latest" of some of the tools. Is this something expected?

kacf: We should perhaps make sure that two runs produce the same result, and isn't affected by dictionary order or something like that.

mzedel: in the gui builds we use the `npm ci` command to install dependencies wherever possible, this should be using fixed dependencies - since it uses the locked version information submitted to the repo (from their docs: "installs are essentially frozen"). However, the license generation tool seems to disagree... I guess I'll have to dig into that to give a better answer

lluiscampos: Yep, we'll need to look at this closer. This is not high priority at the moment, so I suggest to create a JIRA ticket for it with fixVersion 2.3.1.

PR #901, opened 2020-03-06
[autoversion] Add new parameter mender-convert-client-version

lluiscampos:
To let the release responsible update this independently of mender
client tagging.


lluiscampos:
Updated mender-convert release checklist accordingly.

PR #912, opened 2020-03-12
Add AUTOVERSION complain tag to Development title in cover

lluiscampos:
Add AUTOVERSION complain tag to Development title in cover so that we don't forget again to update this in future releases.


lluiscampos:
See also #913, #914, #915 and #916.

PR #923, opened 2020-03-27
migrated travis steps to gitlab

mzedel:
this has to go together with https://github.com/mendersoftware/mender-docs-site/pull/199

a docs-site pipeline trigger can be seen here: https://gitlab.com/Northern.tech/Mender/mender-docs/-/jobs/490608950


.gitlab-ci.yml
```diff
@@ -1,9 +1,42 @@
...
+    - apk add --no-cache bash git openssl pwgen python3
+    - git config --global user.name "user"
+    - git config --global user.email "user@example.com"
+  script:
+    - ./test_autoversion.py
+    - ./autoversion.py --check
+    - env TEST_OPEN_SOURCE=1 ./test_docs.py 07.Administration/02.Production-installation/docs.md
+    - if [ -n "$REGISTRY_MENDER_IO_PASSWORD" ]; then docker login -u ntadm_menderci -p "$REGISTRY_MENDER_IO_PASSWORD" registry.mender.io; fi
+    - if [ -n "$REGISTRY_MENDER_IO_PASSWORD" ]; then env TEST_ENTERPRISE=1 ./test_docs.py 07.Administration/02.Production-installation/docs.md; fi
+    - if [ -n "$REGISTRY_MENDER_IO_PASSWORD" ]; then env TEST_ENTERPRISE=1 ./test_docs.py 07.Administration/02.Production-installation/01.Upgrading-from-OS-to-Enterprise/docs.md; fi
```
lluiscampos: Nitpick: these would look more readable if you use a multiline syntax:
```
    - if [ -n "$REGISTRY_MENDER_IO_PASSWORD" ]; then
        docker login -u ntadm_menderci -p "$REGISTRY_MENDER_IO_PASSWORD" registry.mender.io;
      fi
...
```

.gitlab-ci.yml
```diff
@@ -1,9 +1,42 @@
...
+trigger:mender-docs-site:
+  image: alpine
+  stage: trigger
+  before_script:
+    - apk add --no-cache curl
+  script:
+    - curl -v -f -X POST
+      -F token=$MENDER_DOCS_SITE_TRIGGER_TOKEN
+      -F ref=$CI_COMMIT_REF_NAME
+      https://gitlab.com/api/v4/projects/16842368/trigger/pipeline
```
lluiscampos: So I guess we didn't manage neither to use the built-in trigger functionality of GitLab... :(

mzedel: unfortunately those seem to be limited to Gitlab Silver and above 🙁 - the documentation is a bit unclear there, as the limitation is signaled for the visualisation [here](https://docs.gitlab.com/ee/ci/multi_project_pipelines.html) but I guess that applies for the entire feature

.gitlab-ci.yml
mzedel: 👌 should now be more readable

.gitlab-ci.yml
lluiscampos: Thanks for looking into it

.gitlab-ci.yml
lluiscampos: :+1:

.gitlab-ci.yml
```diff
@@ -1,9 +1,48 @@
...
+  only:
+    refs:
+      - master
+      - production
```
lluiscampos: Wait, I got confused here. We don't have a "production" branch here. But we need to trigger this for release branches (2.3.x, etc). We should then use the same regex as for regular repos:
```
  only:
    - /^(master|[0-9]+\.[0-9]+\.x)$/
```

mzedel: should now be fixed 👌

PR #933, opened 2020-04-15
Fix and beautify pipeline

lluiscampos:
- [test_docs] Changes in production docs commands to fix CI build
Ignore the tenantadm --help command (not very important) and remove -it
flags from docker run commands.
These changes fix "the input device is not a TTY" errors seen in the CI
pipeline.

- [pipeline] Move test job to test stage
Minor nitpick, but let's explicitly state that this repo does not
"build" anything.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,6 +1,5 @@
 stages:
   - test
-  - build
   - trigger
 
 include:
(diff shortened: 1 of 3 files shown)


lluiscampos:
@mzedel FYI this was broken since we ported the tests to GitLab. The reason for going unnoticed is the same problem we had already in Travis: the protected variables only available for protected branches makes the enterprise test not to run in PRs.

lluiscampos:
To verify the fix, I protected this branch in GitLab and let the pipeline run. The logs are here: https://gitlab.com/Northern.tech/Mender/mender-docs/-/jobs/511868641

lluiscampos:
> So I guess this then broke `master` due to the env variables inheritance?

We didn't break it because `master` was never green after the full migration to GitLab. It is not related to the env variables inheritance (group) but just that the variables are not present on PRs so we didn't see this failing in the final work for porting it to GitLab.

PR #942, opened 2020-04-22
Add note block on mender-convert transitions

HerrMuellerluedenscheid:
As this can influence the development road map right from the start I think it is a good idea to emphasize this point. 
https://hub.mender.io/t/migrating-from-mender-converted-debian-to-yocto/506/2

### DIFF
--- a/04.Artifacts/15.Debian-family/01.building-a-mender-debian-image/docs.md
+++ b/04.Artifacts/15.Debian-family/01.building-a-mender-debian-image/docs.md
@@ -15,6 +15,10 @@ The output of `mender-convert` is:
   Mender through the Mender Server.
 * An ext4 image of the modified filesystem for advanced users.
 
+**Note:** Provisioning a device with an image created using `mender-convert` and
+transitioning to a `yocto` image thereafter can be tedious, error prone and is,
+thus, discouraged as discussed [here](https://hub.mender.io/t/migrating-from-mender-converted-debian-to-yocto/506/2).
+
 ### Recommended workflow
 
 The recommended workflow for using `mender-convert` is to rely on a *golden disk


HerrMuellerluedenscheid:
Had a type in the PR title. Maybe, that's why the CI test fails.

estenberg:
Thanks for your contribution @HerrMuellerluedenscheid !
This would be a good workflow to support as many use the stock (Debian-family) image during prototyping/piloting while switching to Yocto in production/scale.

Are these issues board-specific, e.g. only about boot partition structure Raspberry Pi? We try to keep the docs board-agnostic, and the board specific information in Mender Hub articles/posts.

04.Artifacts/15.Debian-family/01.building-a-mender-debian-image/docs.md
```diff
@@ -15,6 +15,10 @@ The output of `mender-convert` is:
   Mender through the Mender Server.
 * An ext4 image of the modified filesystem for advanced users.
 
+**Note:** Provisioning a device with an image created using `mender-convert` and
+transitioning to a `yocto` image thereafter can be tedious, error prone and is,
+thus, discouraged as discussed [here](https://hub.mender.io/t/migrating-from-mender-converted-debian-to-yocto/506/2).
```
kacf, line 20: I think this problem is specific to Raspberry Pi, with its complicated and non-standard boot procedure. AFAIK, with Beaglebone and other GRUB based boards, there is no such problem. I'm not sure if this belongs here, I'm thinking it should rather go into the [Known Issues section for Raspberry Pi](https://hub.mender.io/t/raspberry-pi-3-model-b-b/57).

HerrMuellerluedenscheid:
Thanks for raising that question @estenberg! The added note was solely based on the linked discussion. Further, I was not aware of the complexity of the boot process that only applies to RPis. Thanks for drawing my attention to that. You are right, this should be kept SoC agnostic.

04.Artifacts/15.Debian-family/01.building-a-mender-debian-image/docs.md
HerrMuellerluedenscheid: Thanks a lot for the clarification!

PR #943, opened 2020-04-23
Remove section describing device limit configuration

alfrunes:
The configuration has been removed as it becomes ineffective in
multi-tenant setups, and has no value in single-tenant setups.
Instead, the enterprise device limit per/plan is made configurable and
is documented in the production template in mendersoftware/integration.


drewmoseley:
@alfrunes I'll leave it to you to merge when all the code changes are in.

PR #947, opened 2020-04-27
update hosted Mender plan reference in quick start

mirzak:
This was missed when cherry-picking, https://github.com/mendersoftware/mender-docs/commit/b8ca6326d9ce49060ecc2b169d6038c0fcbab589

### DIFF
--- a/01.Getting-started/01.Quickstart-with-raspberry-pi/docs.md
+++ b/01.Getting-started/01.Quickstart-with-raspberry-pi/docs.md
@@ -22,7 +22,7 @@ To follow this guide, you will need the following:
 
 Get a hosted Mender account by [signing up here](https://mender.io/signup?target=_blank).
 
-!!! We provide $120 free credit for you to use for evaluation. You can cancel at any time without incurring a cost while your usage remains under $120.
+!!! We provide three months free usage on the Mender Starter plan for you to use for evaluation. You can cancel at any time without incurring a cost.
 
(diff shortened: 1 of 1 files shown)


01.Getting-started/01.Quickstart-with-raspberry-pi/docs.md
```diff
@@ -22,7 +22,7 @@ To follow this guide, you will need the following:
 
 Get a hosted Mender account by [signing up here](https://mender.io/signup?target=_blank).
 
-!!! We provide $120 free credit for you to use for evaluation. You can cancel at any time without incurring a cost while your usage remains under $120.
+!!! We provide three months free usage on the Mender Starter plan for you to use for evaluation. You can cancel at any time without incurring a cost.
```
estenberg, line 25: This is correct and specific, we just need to update it again if it changes in the future. :)

Did you do a quick grep for 120 or free to make sure it's the only reference?

estenberg: Can also leave it a bit more vague, like "We provide a free trial period for you to use during evaluation...". That would ensure we don't have to update it again :)

mirzak: > Did you do a quick grep for 120 or free to make sure it's the only reference?

Indeed. This should be the only reference.

PR #951, opened 2020-05-04
Clarify some upgrade/downgrade information.

kacf:
Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>

PR #954, opened 2020-05-04
Remove foot note from column where it makes no sense (2.0 not supported)

kacf:


kacf:
A small followup to #951.

PR #955, opened 2020-05-04
Clarify some upgrade/downgrade information.

kacf:
(cherry picked from commit 4588cc6e4c3a7e16f83babefb805048a5e4c0bcf)


lluiscampos:
For these old branches, the active CI is Jenkins. The GitLab expected check can be ignored.

PR #961, opened 2020-05-05
MEN-4366: Update Docker image tags for the new versionig

lluiscampos:
From 2.4.x on, the backend services will use tags like mender-1.2.3,
where 1.2.3 is the integration version (not the git version of the
component).


lluiscampos:
Relates to https://github.com/mendersoftware/mendertesting/pull/64

PR #966, opened 2020-05-14
Switch from docker-compose to docker for deviceauth maintenance

alfrunes:
Use docker exec instead of the docker-compose command in order to avoid the error:
ERROR: Service 'mender-deployments' depends on service 'storage-proxy' which is undefined.

Thank you @koptermax


alfrunes:
@koptermax I created a new PR to target master instead of the `2.3.x` tag and signed off your commit. Everything should be set now :)

PR #976, opened 2020-05-20
Migrate System-Updates/{Yocto,Debian}/Overview sections

oleorhagen:
This migrates both sections, as they are pretty much the same. 

That means:
* MEN-3205
* MEN-3211
both

### DIFF
--- a/04.System-updates/01.Debian-family/01.Overview/docs.md
+++ b/04.System-updates/01.Debian-family/01.Overview/docs.md
@@ -5,48 +5,56 @@ taxonomy:
     label: reference
 ---
 
-##Device capacity
+Mender has official support for binary-OS images based on Debian. The general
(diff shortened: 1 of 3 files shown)


04.System-updates/01.Debian-family/01.Overview/docs.md
```diff
@@ -5,48 +5,56 @@ taxonomy:
     label: reference
 ---
 
-##Device capacity
+Mender has official support for binary-OS images based on Debian. The general
```
drewmoseley, line 8: Do we want to drop the "based on Debian" bit here?  That seems overly restrictive in the docs.

drewmoseley: Never mind. I think this is clear after reading the whole thing.

PR #978, opened 2020-05-22
Add on-prem as an optional prerequisite for quickstart guide.

kacf:


kacf:
Ping @estenberg.

estenberg:
Sorry for the delay and thanks for the proposed improvement.

However in this case I do not think this is a good idea because we are working to create the lowest-possible friction for onboarding by creating a single, default path with the reference environment, using hosted Mender and RPi. We are optimizing this going forward as you know as well (CC requirement lifted, etc.).

This tutorial will not work 100% with on-premise (it needs some "if statements"), so in order to ensure it is simple I would just keep this narrow to hosted & RPi.

Does that make sense?

kacf:
Ok, makes sense.

PR #982, opened 2020-05-27
Add: Documentation style enforcement

oleorhagen:
Changelog: None


lluiscampos:
Looks promising... let me know when you want a formal review ;)

oleorhagen:
@lluiscampos feel free to have a look now :)

.gitlab-ci.yml
```diff
@@ -34,6 +35,27 @@ test:
...
+  stage: styleenforce
+  image: ubuntu
+  script:
+    - find . -type f -name docs.md | xargs -i bash scripts/weasel.sh {}
+
+style:passive-voice:
+  stage: styleenforce
+  image: ubuntu
+  script:
+    - find . -type f -name docs.md | xargs -i bash scripts/passive.sh {}
```
lluiscampos, line 57: I see the point on having all of these in different jobs, but think that we are going to spin up a new container for each which is an overkill in resources and takes longer time. I suggest we move all to a single job.

.gitlab-ci.yml
```diff
@@ -34,6 +35,27 @@ test:
     #     env TEST_ENTERPRISE=1 ./test_docs.py 07.Administration/02.Production-installation/01.Upgrading-from-OS-to-Enterprise/docs.md;
     #   fi
 
+style:readability:
+  stage: styleenforce
```
lluiscampos, line 39: I don't think we need a new stage. It can be run in "test" in parallel to the rest of the tests.

Generally speaking, we should aim of having the same standard stages in all pipelines so that they have some coherence.

scripts/gunning-fog.py
```diff
@@ -0,0 +1,30 @@
...
+for file in sys.argv[1:]:
+    with open(file) as f:
+        text = f.read()
+        r = Readability(text)
+        try:
+            gf = r.gunning_fog()
+        except ReadabilityException as e:
+            sys.exit(0)
+
+        if not 9 <= int(gf.score) <= 12:
```
lluiscampos, line 20: Let's add a link to what this numbers mean: https://en.wikipedia.org/wiki/Gunning_fog_index

scripts/passive.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+#
+# Simple passive voice checker from "My Ph.D advisor rewrote himself in Bash"
```
lluiscampos, line 3: Where are these coming from, really?

.gitlab-ci.yml
drewmoseley: Yeah, we may want to consider output from this as warnings rather than failures.  I do like the idea of at least not making it worse.

Once we have a list of the issues, then perhaps we can make a better choice how to handle them.

oleorhagen: I too think these should be soft-failures.

Another idea I had was to have this only check the actual PR diff, but be strict about it, which I think should be an acceptable goal.

This could possibly be done from the bot, which then leaves a comment instead of a gitlab runner failure. This would ensure explicit notification of failure, a history left in the PR. And the ability to ignore it at your own leisure.

lluiscampos:
Discussed in person with @oleorhagen. Unless others in the team strongly oppose, the conclusions are:
- Doing the documentation restructuring is a golden opportunity to introduce these kind of checks. We really want get this PR merged! It also resonates well in process and principles to what we do with our code.
- We will enforce 2 of the checks: passive voice and weasel. @oleorhagen takes responsibility in fixing the already merged pages that fail on these.
- For the Gunning Fog index check: we will relax the criteria. Still to be decided if we will enforce it (i.e. if it is simple and intuitive how to improve the index) or just warn about it (if it turns out difficult for a typical doc writer to improve).

Again, thanks @oleorhagen for a very valuable initiative :rocket:

oleorhagen:
@lluiscampos FUCK YEAH!

Merge?

oleorhagen:
I'm merging for now, so that this is present on monday when I am not here. So that more catch-up won't be required from my side.

PR #983, opened 2020-05-27
Adjust Quickstart steps for on-prem installation.

kacf:
The on-prem Quickstart is telling you to start by SSH'ing to the
device, but you need to start by flashing it, as for Hosted Mender.

### DIFF
--- a/01.Getting-started/01.Quickstart-with-raspberry-pi/docs.md
+++ b/01.Getting-started/01.Quickstart-with-raspberry-pi/docs.md
@@ -182,7 +182,7 @@ And finally fire up the demo server environment with:
 
 Note that the demo up script starts the Mender services, adds a demo user with the username mender-demo@example.com, and assigns a random password in which you can change after you log in to the Mender web UI. The Mender UI can be found on [https://localhost](https://localhost?target=_blank).
 
-After you log into the UI on the localhost you can follow steps 1 through 6 listed above.
+After you log into the UI on the localhost you can follow the steps above, starting with ["Prepare your Raspberry Pi"](#prepare-your-raspberry-pi).
 
 ## Have any questions?
 


01.Getting-started/01.Quickstart-with-raspberry-pi/docs.md
```diff
@@ -182,7 +182,7 @@ And finally fire up the demo server environment with:
 
 Note that the demo up script starts the Mender services, adds a demo user with the username mender-demo@example.com, and assigns a random password in which you can change after you log in to the Mender web UI. The Mender UI can be found on [https://localhost](https://localhost?target=_blank).
 
-After you log into the UI on the localhost you can follow steps 1 through 6 listed above.
+After you log into the UI on the localhost you can follow the steps above, starting with [preparing your Raspberry Pi](#prepare-your-raspberry-pi).
```
lluiscampos, line 185: Nitpick: the text in the link should say "prepare" instead of "preparing".

kacf: Hmm, I think that would be grammatically wrong though. The intention was not to recreate the header exactly, but to use a link in a naturally formed sentence. Maybe @drewmoseley can provide some writing advice?

drewmoseley: I think since "Prepare your Raspberry Pi" is the title of the step being referred to that is correct in this case. I would quote it and capitalize it the same.

kacf: Ok, will do that, thanks!

kacf:
Cherry-picked to 2.4.x as well.

PR #987, opened 2020-05-28
use 2.0.x branch for mender-convert git clone

mirzak:
The 2.0.0 tag is broken because the Docker base
image we used got removed, meaning that './docker-build'
fails with errors similar to this:

    > Err:4 http://archive.ubuntu.com/ubuntu disco Release
    >   404  Not Found [IP: 91.189.88.152 80]

### DIFF
--- a/04.Artifacts/15.Debian-family/01.building-a-mender-debian-image/docs.md
+++ b/04.Artifacts/15.Debian-family/01.building-a-mender-debian-image/docs.md
@@ -93,9 +93,8 @@ documentation for your host OS if you encounter connection issues with docker.
 
 Clone `mender-convert` from the official repository:
 
-<!--AUTOVERSION: "-b % https://github.com/mendersoftware/mender-convert"/mender-convert-->
 ```bash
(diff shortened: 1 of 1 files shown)


kacf:
I was instead going to suggest releasing a quick 2.0.1. It's just one cherry-pick missing. I don't think we should refer to ".x" versions in our documentation.

mirzak:
> I was instead going to suggest releasing a quick 2.0.1. It's just one cherry-pick missing

Sure, if we can make it happen this is definitely preferred.

> I don't think we should refer to ".x" versions in our documentation.

I do not think so either, but a temporary solution referencing `.x` is better then not working :)

oleorhagen:
Huh? ubuntu:19 got removed?

kacf:
Release has been initiated.

lluiscampos:
> Huh? ubuntu:19 got removed?

@oleorhagen Yes, Ubuntu has modify their policy so that the non-TLS versions are supported for only one year. In the repos that we find, would be good to base our builds on Ubuntu 20.04 (TLS)

PR #1003, opened 2020-06-10
Add the passive voice and weasel word checker scripts

oleorhagen:
There is currently no CI along with this.

So the scripts are only intended to be run locally on the files as they are written and merged into the new documentation structure.


mirzak:
Good stuff. Please target `master` branch

PR #1005, opened 2020-06-10
Checklinks

oleorhagen:
Make the checklinks script more user friendly.

Plus add a utility script for changing links automatically.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -39,7 +39,7 @@ test:internal-links:
   script:
     - apk add --no-cache git
     - npm install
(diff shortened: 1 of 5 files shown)


oleorhagen:
> Hard to comment on `.js` stuff, but overall changes look good to me.
> 
> What I think we are lacking right now is documentation, both for this and the passive/weasel scripts.
> 
> I think we can followup on this, and we should probably discuss this during our call, to find a good common location for topics like these.

Yes, in general, documenting how you're supposed to write documentation, which means that this should be a part of it, will have to be a next step soon I think.

PR #1009, opened 2020-06-11
New 'Get started" section

mirzak:
Rendering available here:

http://13.53.78.221:8080/foo

PR #1012, opened 2020-06-15
Switch the link-checker to use Grav-slug when checking sections

oleorhagen:
This will fix up errors which came from using the Github-slugger, as they are
not 100% compatible with Grav.

Changelog: None


oleorhagen:
@mirzak this did the trick locally for me on your PR #1009 
Thus, if you're happy, rebase on top of this, and it should be good to go (hopefully :crossed_fingers: )

mirzak:
Testing :)

PR #1018, opened 2020-06-17
Add a bash style-guide readme

oleorhagen:
Changelog: None

### DIFF
--- a/README-bash-styleguide.markdown
+++ b/README-bash-styleguide.markdown
@@ -1,0 +1,119 @@
+# Bash style guide
+
+This is a short guide on how the Bash (CLI) examples in this documenation is to
+be structured.
+
+## Code blocks
+
+### Highlighting
+
+#### CLI commands
+
+Code blocks with shell commands should always have the `bash` tag, for syntax highlighting.
+
+i.e.
+
+```````
+ ```bash
+ echo foobar
+ ```
+```````
+
+Not
+
+``````
+```console
+echo foobar
+```
+``````
+
+#### Scripts
+
+Scripts should always have the `bash` tag, for syntax highlighting. That is:
+
+```````
+```bash
+#! /bin/bash
+echo foobar
+```
+```````
+
+### Remove the prompt string
+
+__PS__, or the _prompt string_ should always be removed from the CLI examples.
+This makes copy-pasting the examples straight forward. That is:
+
+```````
+```bash
+echo foobar
+```
+```````
+
+Not
+
+```````
+```bash
+$ echo foobar
+```
+```````
+
+### Code output
+
+Output from code examples should follow all the rules above, except that the
+prompt string is encouraged. The syntax highlighting should be `bash`.
+Output should also be formatted in a quote block. That is:
+
+```````
+> ```bash
+> $ echo foobar
+> foobar
+> ```
+```````
+
+Not
+
+```````
+```bash
+echo foobar
+foobar
+```
+```````
+
+
+## Variables
+
+Variables should always use the `${VARIABLE}` syntax. That is:
+
+```````
+```bash
+cd "${VAR}"
+```
+```````
+
+Not
+
+```````
+```bash
+cd $VAR
+```
+```````
+
+## Long lines
+
+If commands grow too long, wrap them over multiple lines through escaping the line. That is:
+
+```````
+```bash
+SOME_ENV_VAR=foobar bash-script --flag1 arg1 \
+                                --flag2 arg2 \
+                                ...
+```
+```````
+
+Not
+
+```````
+```bash
+SOME_ENV_VAR=foobar bash-script --flag1 arg1 --flag2 arg2 ...
+```
+```````


oleorhagen:
@mirzak did you forget about this one?

mirzak:
> @mirzak did you forget about this one?

Not forget, but has been pushed down my list.

Before I do an deeper review, I have a question regarding `shell-session`. This tag was new to me. Where does this come from? Is it Grav that supports this? What is the difference between this and e.g `bash`? Reason I have so many questions is that we do not have any occurrences of `shell-session` in the current docs while the style guide would say that this is the prefererad way :)

oleorhagen:
> > @mirzak did you forget about this one?
> 
> Not forget, but has been pushed down my list.
> 
> Before I do an deeper review, I have a question regarding `shell-session`. This tag was new to me. Where does this come from? Is it Grav that supports this? What is the difference between this and e.g `bash`? Reason I have so many questions is that we do not have any occurrences of `shell-session` in the current docs while the style guide would say that this is the prefererad way :)

Yes, if you agree to the `shell-session` markdown, then I will go ahead, and change them pretty much throughout the documentation.

The reason being:
`bash` highlighting is really meant for `bash-scripts`, and not for shell commands, like we are mostly doing. i.e., run these commands in your terminal. If you highlight these with `bash`, then it will highlight bash keywords, and I think it looks a little odd.

If you highlight it with `shell-session`, it knows that this is an interactive shell session, and highlights it accordingly.

See: http://13.53.78.221:8090/foo/getting-started/quickstart-with-raspberry-pi for an example.

mirzak:
Even though I agree with the reasoning, I actually prefer the way "with highlight bash" looks. It just looks more familiar and based on a quick look around it seems the "common" way (looked at Docker, Kubernetes and jfrog docs)

But lets get some more input, tagging stakeholder.

@michaelatmender (UX)
@estenberg

oleorhagen:
> Even though I agree with the reasoning, I actually prefer the way "with highlight bash" looks. It just looks more familiar and based on a quick look around it seems the "common" way (looked at Docker, Kubernetes and jfrog docs)
> 
> But lets get some more input, tagging stakeholder.
> 
> @michaelatmender (UX)
> @estenberg

Yes, sounds good. Note that the bash highlighting is a little bad, and only does the most common keywords. Thus, 

```bash
echo foobar
```

Highlights echo. Whilst:

```bash
ls -la .
```

Does not highlight `ls`

mirzak:
Note that there are differences in github and Grav highlights `bash`. What I like in Grav is that Variables and strings are highlighted when using `bash`. 

Example from: https://docs.mender.io/development/get-started/deploy-a-system-update

estenberg:
Personally I prefer the way bash highlighting is in Grav. Scrolling through these snippets, the coloring makes it easier to read (e.g. string, command like export, variable, etc. https://docs.mender.io/development/get-started/deploy-a-system-update

Also a plus that other sites like Docker docs does the same way.

So my vote would be to keep it as-is (bash).

oleorhagen:
Alright, majority rules. Changing it.

oleorhagen:
Alright, @mirzak demolish me with that final review

README-bash-styleguide.markdown
```diff
@@ -0,0 +1,83 @@
...
+foobar
+```
+
+
+### Variables
+
+Variables should always use the `${VARIABLE}` syntax. That is:
+
+```bash
+echo ${VAR}
```
mirzak, line 76: We might as well require:

```bash
echo "${VAR}"
```

A better example is:

```bash
cd "${VAR}"
```

We do use plenty of variables for paths, and this should safe-guard us from spaces in paths etc.

README-bash-styleguide.markdown
```diff
@@ -0,0 +1,83 @@
...
+
+#### CLI commands
+
+Code blocks with shell commands should always have the `bash` tag, for syntax highlighting.
+
+i.e.
+
+```bash
+echo foobar
+```
```
mirzak, line 18: When viewing this document on e.g github which renders the Markdown, the actual examples on how to do it are hidden and you just see how it should be rendered. (this applies to all examples)

I would assume most people will read this on github? Or?  What do you think?

README-bash-styleguide.markdown
```diff
@@ -0,0 +1,83 @@
...
+
+```bash
+echo ${VAR}
+```
+
+Not
+
+```bash
+echo $VAR
+```
```
mirzak, line 83: Would add something about multi-line vs long lines (that is split long lines in multiple line expressions :))

README-bash-styleguide.markdown
mirzak: To clarify,

if you go [here](https://github.com/mendersoftware/mender-docs/blob/127fc3f326acd83a47b9cae3e15bc76c8343f53b/README-bash-styleguide.markdown).

You see:

```bash
echo foobar
```

and not:

    ```bash
    echo foobar
    ```

oleorhagen: ahh, haha. Missed it a whole two times in a row

README-bash-styleguide.markdown
```diff
@@ -0,0 +1,121 @@
...
+### Highlighting
+
+#### CLI commands
+
+Code blocks with shell commands should always have the `bash` tag, for syntax highlighting.
+
+i.e.
+
+```````
+
```
mirzak, line 17: Remove extra line here. There are several throughout the document, same comment applies to them.

oleorhagen: What other newlines are you referring to? I only see this one. Are you thinking of between paragraphs?

PR #1017, opened 2020-06-16
[RFC]: Overview/Delta-udpates section

oleorhagen:
This is the current `delta-updates` section.

It is sort of short, and no sections or images.

For the former, I'm not quite sure it is necessary here, with so little text.

For the latter, I did not find any good ones, which I think did fit. Suggestions are more than welcome though.

- O

PR #1027, opened 2020-06-25
add Overview/Delta updates

mirzak:
Based on work in #1017 

Rendering: http://13.53.78.221:8080/foo/overview/delta-updates


mirzak:
@drewmoseley @estenberg @oleorhagen, final review?

PR #1030, opened 2020-06-25
Hosted branch preparations

lluiscampos:
* [hosted] Remove Administration and Troubleshooting/Mender-Server
These sections are not relevant for hosted Mender. Redirect users to
doc page for latest stable release instead.
There is a technical reason to do this: we don't have "hosted" versions
of the microservices that we can use to replace "master" versions in
these pages.

* [hosted] Set branch version to Hosted

* [hosted] Update all versions to current production tags
Versions updated with the following command:
```
./autoversion.py --update --integration-dir <local-dir> \
  --integration-version staging \
  --mender-convert-version 2.0.1 \
  --mender-convert-client-version 2.2.0 \
  --meta-mender-version zeus \
  --poky-version zeus
```


lluiscampos:
This is my proposal for the "hosted" branch of the docs: based on current master, removed the Server maintenance parts, and set the client related components versions to latest production tags.

07.Administration/chapter.md
```diff
@@ -8,5 +8,7 @@ taxonomy:
 
 # Administration
 
-Configuration and administration of Mender services.
-Use [hosted Mender](https://hosted.mender.io?target=_blank) to save time setting up and maintaining the Mender server.
+[Hosted Mender](https://hosted.mender.io?target=_blank) is configured, managed
+and upgraded by the Mender team. To read documentation for self-managed
+installations of the Mender server, visit
+[the documentation page of the latest stable release](https://docs.mender.io/2.3/administration).
```
kacf, line 14: This won't be covered by any `AUTOVERSION` tag. Unfortunately I think the `complain` tag won't work here either. I'm not sure what the best approach here is. Maybe add an item to Hosted Mender's release checklist?

chapter.md
```diff
@@ -10,9 +10,9 @@ For page headers the tag may come after due to misrendering if it is above.
...
 
 <!--AUTOVERSION: "### Development"/integration/complain-->
-### Development
+### Hosted
 
 # Documentation
 
 <!--AUTOVERSION: "bleeding-edge % branch"/integration/complain-->
-Documentation for Mender's bleeding-edge master branch.
+Documentation for hosted Mender.
```
kacf, line 18: Nitpick, should say "**Hosted** Mender".

07.Administration/chapter.md
lluiscampos: Yep. My plan was to add an item to the **on-prem** release checklist (not Hosted checklist, though).

chapter.md
lluiscampos: Hehe, no. The correct term nowadays is "hosted Mender"; it is everywhere else in the docs. There were few PRs fixing this some time ago.

estenberg, line 18: Would add a URL and a short sentence.

Suggesting: Documentation for [hosted Mender](https://hosted.mender.io?target=_blank), the secure Mender server hosted by the team behind Mender. If you are running Mender on-premise, rather use the documentation for the Mender version you are running.

lluiscampos: > Btw, are all the client/standalone tools (e.g. mender-artifact) from the "latest stable" release of on-premise, or how is that decided?

@estenberg  Your question is a bit confusing to me. The standalone tools (client, mender-artifact, mender-convert) are from the latest "on-premise" release, yes. But we don't release these tools without an on-prem release, so there are no "hosted" versions of these...

lluiscampos:
> Btw, are all the client/standalone tools (e.g. mender-artifact) from the "latest stable" release of on-premise, or how is that decided?

@estenberg Your question is a bit confusing to me. The standalone tools (client, mender-artifact, mender-convert) are from the latest "on-premise" release, yes. But we don't release these tools without an on-prem release, so there are no "hosted" versions of these...

estenberg:
> > Btw, are all the client/standalone tools (e.g. mender-artifact) from the "latest stable" release of on-premise, or how is that decided?
> 
> @estenberg Your question is a bit confusing to me. The standalone tools (client, mender-artifact, mender-convert) are from the latest "on-premise" release, yes. But we don't release these tools without an on-prem release, so there are no "hosted" versions of these...

OK! That was what I was wondering and as expected - thank you.

mirzak:
Looks good. I do not have much to add beside what has already been mentioned.

Though before merging this it would be nice if we can sync. Reason being is that there are a couple of PRs pending that would have been "nice" to get in before we branch out to "hosted".

chapter.md
kacf: Aha, I guess I'm out of the loop!

07.Administration/chapter.md
kacf: Ah yes, of course, this will be on every on-prem release, not hosted. Then I say :+1:

lluiscampos:
> Though before merging this it would be nice if we can sync. Reason being is that there are a couple of PRs pending that would have been "nice" to get in before we branch out to "hosted".

We can merge to "hosted" now, it won't go live or anything. Then we can launch more PRs merging master into hosted branch every time we need to update the content here.

mirzak:
> @mirzak We can merge to "hosted" now, it won't go live or anything. Then we can launch more PRs merging master into hosted branch every time we need to update the content here.

Sure, it is fairly standalone now.

Something unrelated, that I noticed while rendering this PR locally. We have this header that says:

> You are browsing documentation for a version other than the latest stable release. Switch to the latest stable release, 1.7.

I am not sure how this will work when we have "Hosted" section which will be default and a user is browsing on of the versioned sections. It will probably not be relevant to keep reminding them to go to "Hosted" version of the docs if they are interested in the latest stable release.

lluiscampos:
> Something unrelated, that I noticed while rendering this PR locally. We have this header that says:
> 
> > You are browsing documentation for a version other than the latest stable release. Switch to the latest stable release, 1.7.
> 
> I am not sure how this will work when we have "Hosted" section which will be default and a user is browsing on of the versioned sections. It will probably not be relevant to keep reminding them to go to "Hosted" version of the docs if they are interested in the latest stable release.

That is a good point. We'll need to dig into `mender-docs-site` code to fix this. My suggestion is to accept hosted and latest stable as "current", and only show this reminder on old versions or beta code. But I am not sure how difficult is this. I'll look at it.

lluiscampos:
With regards to the still broken links,

They are links to existing pages but without the expected headers (because the content has been replaced).

I had a chat with @oleorhagen and we decided that the best is to have a convenient way to "whitelist" missing links, so that this branch could have the current missing links and still pass CI.

If this is something we cannot fix in one day, we will merge this PR to "hosted" branch (breaking CI) and create a ticket to follow up ASAP. I don't like accepting red builds but we have a tight schedule for this one.

mirzak:
> I had a chat with @oleorhagen and we decided that the best is to have a convenient way to "whitelist" missing links, so that this branch could have the current missing links and still pass CI.

Do the links still work, and it is only the CI that complains?

lluiscampos:
> > I had a chat with @oleorhagen and we decided that the best is to have a convenient way to "whitelist" missing links, so that this branch could have the current missing links and still pass CI.
> 
> Do the links still work, and it is only the CI that complains?

@mirzak So the link will "work" in the sense that you get into the page, but not into the HTML anchor. For example like going to https://docs.mender.io/2.3/getting-started#inexisting-header

mirzak:
> @mirzak So the link will "work" in the sense that you get into the page, but not into the HTML anchor. For example like going to https://docs.mender.io/2.3/getting-started#inexisting-header

That was my definition of "works" as well. This is acceptable I believe. 

Over time I think we should clean up these links (remove and maybe replace with content instead). I will take a note on that.

lluiscampos:
Related PR: https://github.com/mendersoftware/mender-docs-site/pull/214

mirzak:
Please rebase on top of master before this is merged.

lluiscampos:
> Please rebase on top of master before this is merged.

@mirzak I'll merge this and launch a follow-up PR straight away.

PR #1034, opened 2020-06-26
Minor fixes in some internal docs links

lluiscampos:


lluiscampos:
> I've been shown a fair bit of proof today, that my migration script was not perfect see_no_evil

@oleorhagen Still was pretty good, I think!

PR #1036, opened 2020-06-26
update order based on new architecture proposal

mirzak:
As following (there are gaps for missing sections):

01.Introduction
02.Device-Group
03.Artifact
04.Deployment
06.Delta-update
07.Identity
08.Inventory
10.State-script
11.Security
12.Role.Based.Access.Control
13.Device-authentication
14.Compatibility

Also renamed:

02.Device-Groups > 02.Device-Group
04.Deployments -> 04.Deployment
06.Delta-updates -> 06.Delta-update
10.State-scripts > 10.State-script


mirzak:
> Now this is coming together real nice

Indeed.

PR #1040, opened 2020-06-29
Update hosted from master

lluiscampos:


lluiscampos:
Something very strange happened with GitHub here... I think this PR is actually merged into `hosted` branch, but there is some inconsistency with the service and the PR "thinks" it is open and with conflicts.

lluiscampos:
Closing; it is actually merged.

PR #1044, opened 2020-06-29
[hosted] Add server related pages to checklinks-whitelist

lluiscampos:
So that CI accepts these missing "pages".

Note that the pages are actually there. What is missing is the headers
(as the content is replaced) but the chechlinks tool does not support
white listing only these atm.


.checklinks-whitelist
```diff
@@ -0,0 +1,9 @@
+{
+  "paths": [
+    "200.APIs",
+    "201.Troubleshooting/04.Mender-Server/docs.md#cleaning-up-the-deviceauth-database-after-device-decommissioning",
```
oleorhagen, line 4: Currently, this won't work. It will only ignore all headers in one file for now. 

As a follow up though, would you prefer it like this? That not listing a section is like a wildcard, and all sections are ignored, and listing the section only ignores that one specifically?

lluiscampos: It's okey: I used the page link after all. See #1046 1046

lluiscampos:
Re-submitted at #1046

PR #1053, opened 2020-07-01
updates to Get Started

mirzak:
Align screenshots with UI


mirzak:
@lluiscampos would need to have this in the hosted branch before we switch to hosted to default.

PR #1055, opened 2020-07-01
[hosted] Remove APIs/Open-source section and adjust links

lluiscampos:
Version "hosted" does not have OS APIs.


lluiscampos:
FYI, I needed to adjust also all the links that were pointing to the OS APIs.

PR #1056, opened 2020-07-01
Remove inconsistent reference to `--client` argument.

kacf:
It is not specified anywhere else that you should use client. Rely on
the onboarding to tell you this instead.

### DIFF
--- a/07.Administration/02.Demo-installation/docs.md
+++ b/07.Administration/02.Demo-installation/docs.md
@@ -127,5 +127,5 @@ run the following commands in the `integration-master` directory:
 ```
 
 ```bash
-./demo --client up
+./demo up
 ```


mirzak:
It has been a while since I tested the on-boarding for a local server, so not really sure what this change is about or the implications but I trust you :)

kacf:
These instructions are brand new, we have never used them before (and they will not be used in 2.4, 2.5 will be the first). This is just something I spotted while making the test spreadsheet.

The reason I am suggesting it is because the instructions higher up tell you to run `./demo up` without the `--client` argument, and in fact there is no mention of a client anywhere on that page.. And running the script with that argument is wrong if you are using RPi. So we should just let the onboarding in the UI tell you to do it instead of saying it here.

PR #1064, opened 2020-07-06
[tests] Fix "the input device is not a TTY" error in CI tests

lluiscampos:
* [tests] Add tests for all TENANT_ID/TENANT_TOKEN variables
All these variables are actually empty at the moment. We'll fix it, but
these tests will catch future regressions.

* [tests] Production installation: add -T flag to docker-compose exec
Fixes "the input device is not a TTY" error in CI tests.


lluiscampos:
This should fix the [issue in master](https://gitlab.com/Northern.tech/Mender/mender-docs/-/pipelines/163423581).

PR #1066, opened 2020-07-07
Taxonomy: Adding the Mender Client definition

merlin-northern:
ChangeLog:none


merlin-northern:
> This looks ok to me but I wonder if we should add a term for "Mender Client Device" as well since it is sometimes confusing when people refer to the entire device as a client.

https://github.com/mendersoftware/mender-docs/pull/1068

PR #1094, opened 2020-07-21
Signing System taxonomy entry

merlin-northern:
ChangeLog:none


merlin-northern:
> I don't see a problem adding this however since it is used only one place and somewhat self-explanatory I think we could leave it out as well.

creating PRs for closing thats my modus operandi.

merlin-northern:
@estenberg let make a deal, if @drewmoseley accepts, we merge, if not, you close.

drewmoseley:
Certainly no harm in it.  Let's merge.

PR #1099, opened 2020-07-27
Doc restructuring: Write System-updates-Debian-family chapter

kacf:
Also add Modify-an-Artifact section to Artifact-creation.

**This should not be reviewed by itself, but should be reviewed together with the restructuring plan in Google Docs**


kacf:
The latest update should address all review comments. If I hear nothing more I will merge this by the end of today.

Because this chapter essentially is half of the old chapter under "03.Devices", I'm not going to fix broken links now. Instead I'll merge this into the `system_update_rewrite` topic branch, and then I will start working on the other half, "System updates: Yocto Project". After that it will be much easier to fix the broken links.

.ispell_english
```diff
@@ -2192,6 +2238,7 @@ tmpdir
 tmpf
 tmpfiles
 Tobias
+TODO
```
kacf, line 2241: @drewmoseley: Btw, should this be an allowed word? I would argue it's the ultimate **dis**allowed word! :smile:

kacf:
Merged this into the topic branch. I know Mirza has some outstanding comments in the Google Document, but I will address those in a separate PR.

PR #1104, opened 2020-07-29
Missing chapters for Client installation and Artifact creation

merlin-northern:
Fix error on both
* https://docs.mender.io/development/artifact-creation
and
* https://docs.mender.io/development/client-installation

ChangeLog:none

### DIFF
--- a/06.Artifact-creation/chapter.md
+++ b/06.Artifact-creation/chapter.md
@@ -1,0 +1,11 @@
+---
+title: Artifact creation
+taxonomy:
+    category: docs
+---
+
+### Chapter 7
+
+# Artifact creation
+
+Package your software for deployment with Mender.
--- a/03.Client-installation/chapter.md
+++ b/03.Client-installation/chapter.md
@@ -1,0 +1,11 @@
+---
+title: Client installation
+taxonomy:
+    category: docs
+---
+
+### Chapter 3
+
+# Client installation
+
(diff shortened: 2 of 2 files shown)


06.Artifact-creation/chapter.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Artifact creation
+taxonomy:
+    category: docs
+---
+
+### Chapter 7
```
kacf, line 7: I think this actually is chapter 6, as the folder name suggests. The reason it is number seven on the web site is that "Creating artifacts" is a duplicate category now, which we have not fully emptied.

merlin-northern: as it is now, it is 7, I believe. I will renumber them in a separate ticket, since we have to remove 04.Artifact (see https://github.com/mendersoftware/mender-docs/pull/1105)

kacf: Ok, if you are keeping track of it, that's fine with me.

06.Artifact-creation/chapter.md
```diff
@@ -0,0 +1,11 @@
...
+
+# Artifact creation
+
+Artifact creation: Package your software for deployment with Mender.
```
mirzak, line 11: Would drop  `Artifact creation:`

merlin-northern: right! copy/paste mistake. thanks.

PR #1113, opened 2020-07-30
doc restructure: System-updates-Yocto-Project

kacf:

PR #1116, opened 2020-08-06
System updates Debian & Yocto Project

mirzak:
Supersedes #1113 & #1099.

This has now been reviewed by @estenberg & @drewmoseley and I think it should go in to master now. Maybe one final check.

This is now rendered at: http://13.53.78.221:8080/foo

There is one problem with the system update sections and that is the heavy usage of passive voice trough out both System update sections. But I think we should do a follow up with this to keep the size down of this PR


mirzak:
@lluiscampos any idea what the QA error means:

```
$ env /tmp/mendertesting/check_python_code_format
Using configuration from /builds/Northern.tech/Mender/mender-docs/pyproject.toml.
.git ignored: matches the --exclude regular expression
would reformat /builds/Northern.tech/Mender/mender-docs/autoversion.py
/builds/Northern.tech/Mender/mender-docs/test_autoversion.py already well formatted, good job.
/builds/Northern.tech/Mender/mender-docs/test_docs.py already well formatted, good job.
Oh no! 💥 💔 💥
1 file would be reformatted, 2 files would be left unchanged.
Please fix your code locally with: black .
This will automatically format the code as needed.
Note that the Black version used is:  19.10b0.
ERROR: Job failed: exit code 1
```

PR #1122, opened 2020-08-12
Hosted update 2020.08.12

lluiscampos:
* Merge branch 'master' into hosted-update-2020.08.12
Resolved Conflicts (removed):
  03.Devices/03.Debian-family/docs.md
  03.Devices/10.Update-Modules/docs.md
  04.Artifacts/25.Modifying-a-Mender-Artifact/docs.md
  05.Client-configuration/02.Cross-compiling/docs.md

* [hosted] Remove server admin content after merge
Namely content in sections 07.Server-installation and
201.Troubleshoot/04.Mender-Server, same as we did in commit 8d297dc.

* [hosted] Update all versions to current production tags
Versions updated with the following command:
./autoversion.py --update --integration-dir <local-dir> \\
      --integration-version staging \\
      --mender-convert-version 2.1.0 \\
      --mender-convert-client-version 2.3.0 \\
      --meta-mender-version dunfell \\
      --poky-version dunfell \\
      --mender-binary-delta-version 1.1.0

PR #1123, opened 2020-08-12
Change API link to Enterprise and fix anchor

lluiscampos:
In the context of this page, it is irrelevant if we point to OS or
Enterprise (and it will help on the hosted branch). Also fix the anchor
in the link that was broken.


lluiscampos:
Indirectly approved at #1122

PR #1137, opened 2020-08-13
[autoversion.py] Exclude 200.API dir from version checks

lluiscampos:
This directory contain the checked-in API specs on old branches, which
might contain version strings that we want to ignore.

Add also "node_modules" to the ignore list, so that we can run the
script locally together with the checklinks tool installed.

### DIFF
--- a/autoversion.py
+++ b/autoversion.py
@@ -69,16 +69,15 @@ def get_version_of(repo):
 
 
 def walk_tree():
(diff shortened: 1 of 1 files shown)


lluiscampos:
@merlin-northern Thanks!

(Last force-push is only style change to make `black` happy)

PR #1146, opened 2020-08-20
MEN-3814: Format variable values correctly

alfrunes:
Use HTML `&lt;` and `&gt;` to represent < and >, otherwise the angled
brackets are interpreted as invalid HTML tags.


lluiscampos (approved):
Thanks

PR #1148, opened 2020-08-24
Add Client Certificates tutorial.

kacf:


08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
+---
+title: Client certificates
+taxonomy:
+    category: docs
+---
+
+Mender supports setting up an ambassador server which can authenticate devices using TLS client certificates. Each client is equipped with a certificate signed by a CA certificate (Certificate Authority), and the ambassador authenticates devices by verifying this signature. Devices that are authenticated are automatically authorized in the Mender backend, and do not need to be manually approved.
```
tranchitella, line 7: `ambassador server` sounds strange; we call it ambassador because that's the name of the pattern in distributed architectures, but I'm not 100% sure this is understandable by people outside of the project. The technical name is edge proxy, which stands for "reverse proxy at the edge of (your) network".

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+organizationName=My Organization
+organizationalUnitName=My Unit
+emailAddress=myusername@example.com
+countryName=NO
+localityName=Oslo
+stateOrProvinceName=Oslo
+EOF
+```
+
+The fields should be filled with information about your organization, locality and contact information. In particular, make sure `commonName` matches the domain name of the server which will serve as the mTLS ambassador. <!-- TODO, link to ambassador setup -->
```
tranchitella, line 84: reword:
```matches the domain name of the edge proxy which will serve as the mTLS ambassador.```

File: 08.Server-integration/03.Client-certificates/docs.md [hunk 3]
```diff
@@ -0,0 +1,289 @@
...
+```bash
+openssl x509 -req -CA ca-cert.pem -CAkey ca-private.key -CAcreateserial -in device-cert.req -out device-cert.pem -days $((365*10))
+```
+
+! The `-days` argument specifies how long the certificate is valid, and can be adjusted if needed. The example expression gives a certificate which is valid for approximately 10 years. Since the certificate will only be used by the server to authenticate devices, it is usually not desirable that it expires after a short time, since this requires certificate rotation on the devices. To manage compromised devices, it is often better to maintain a certificate blacklist on the server.
+
+The generation and signing of the client certificate needs to be repeated for each device, so these are natural steps to automate in your device provisioning workflow.
+
+
+<!-- TODO: EVERYTHING FROM HERE AND DOWN TO "END_OF_SERVER_PART" NEEDS TO BE REWRITTEN. Originally copied from the Pre-Authorizing document. -->
```
tranchitella, line 150: Not sure I understand: can't we simply delete this section?

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+    "Certificate": "/data/mender/mender-cert.pem",
+    "Key": "/data/mender/mender-cert.pem"
+  }
+}
+```
+
+Then copy the modified file back into the disk image:
+
+```bash
+mender-artifact cp mender.conf mender-disk-image.sdimg:/etc/mender/mender.conf
```
tranchitella, line 268: One thing that is missing here is adding the CA cert as trusted, uploading it into `/etc/ssl/certs` and running `update-ca-certificates`.

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+
+Now that we have both a CA certificate, and a certificate request for the device, we need to sign the latter with the former. This produces a signed certificate which the server will recognize when the client connects.
+
+```bash
+openssl x509 -req -CA ca-cert.pem -CAkey ca-private.key -CAcreateserial -in device-cert.req -out device-cert.pem -days $((365*10))
+```
+
+! The `-days` argument specifies how long the certificate is valid, and can be adjusted if needed. The example expression gives a certificate which is valid for approximately 10 years. Since the certificate will only be used by the server to authenticate devices, it is usually not desirable that it expires after a short time, since this requires certificate rotation on the devices. To manage compromised devices, it is often better to maintain a certificate blacklist on the server.
+
+The generation and signing of the client certificate needs to be repeated for each device, so these are natural steps to automate in your device provisioning workflow.
```
oleorhagen, line 147: :+1:

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Set up Mender configuration for client certificate
+
+First, copy the existing `mender.conf` out of the disk image, so that we can edit it.
+
+```bash
+mender-artifact cp mender-disk-image.sdimg:/etc/mender/mender.conf mender.conf
+```
+
+Next, open `mender.conf` in any text editor, and add the following content:
```
oleorhagen, line 243:
```suggestion
Next, open `mender.conf` in a text editor, and add the following content:
```

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+
+```json
+  "HttpsClient": {
+    "Certificate": "/data/mender/mender-cert.pem",
+    "Key": "/data/mender/mender-cert.pem"
+  }
+```
+
+Make sure that the result is valid JSON, in particular that commas appear on every line except the last in a block. The snippet should be added inside the first set of curly braces in the file. For example, it might look like this in a typical `mender.conf` file:
```
oleorhagen, line 252: The last comma thing is such a retarded part of JSON

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+Now provision the storage with this new disk image, just like you have done in the past. If you are using a SD card, insert it into your workstation and use a command similar to the following:
+
+```bash
+sudo dd if=<PATH-TO-IMAGE>.sdimg of=<DEVICE> bs=1M && sudo sync
+```
+
+Then insert the SD card back into your device and boot it.
+
+
+## Verify the device is accepted
```
oleorhagen, line 285:
```suggestion
## Verify that the device is accepted
```

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+
+If everything went as intended, your device will get the `accepted` status in the Mender server. You can log in to the Mender UI to ensure your device is listed and reports inventory.
```
oleorhagen, line 287:
```suggestion
If everything went as intended, your device shows up as `accepted` in the Mender server. You can log in to the Mender UI to ensure your device is listed and reports inventory.
```

oleorhagen (commented):
Good job! Feel free to cherry any of the following, or ignore at will. :)

Also, ran the passive voice checker on it, since this is new material:
```console
mender-docs git:(7080681) ✗ bash ./scripts/passive.sh 08.Server-integration/03.Client-certificates/docs.md
08.Server-integration/03.Client-certificates/docs.md:7:is equipped
08.Server-integration/03.Client-certificates/docs.md:7:are authenticated
08.Server-integration/03.Client-certificates/docs.md:9:are manufactured
08.Server-integration/03.Client-certificates/docs.md:19:been integrated
08.Server-integration/03.Client-certificates/docs.md:35:is booted
08.Server-integration/03.Client-certificates/docs.md:57:be used
08.Server-integration/03.Client-certificates/docs.md:63:be switched
08.Server-integration/03.Client-certificates/docs.md:84:be filled
08.Server-integration/03.Client-certificates/docs.md:92:be adjusted
08.Server-integration/03.Client-certificates/docs.md:92:be used
08.Server-integration/03.Client-certificates/docs.md:97:be generated
08.Server-integration/03.Client-certificates/docs.md:107:be switched
08.Server-integration/03.Client-certificates/docs.md:128:be changed
08.Server-integration/03.Client-certificates/docs.md:128:be filled
08.Server-integration/03.Client-certificates/docs.md:145:be adjusted
08.Server-integration/03.Client-certificates/docs.md:145:be used
08.Server-integration/03.Client-certificates/docs.md:147:be repeated
08.Server-integration/03.Client-certificates/docs.md:154:are documented
08.Server-integration/03.Client-certificates/docs.md:183:is done
08.Server-integration/03.Client-certificates/docs.md:215:be preauthorized
08.Server-integration/03.Client-certificates/docs.md:252:be added
08.Server-integration/03.Client-certificates/docs.md:285:is accepted
08.Server-integration/03.Client-certificates/docs.md:287:is listed
08.Server-integration/03.Client-certificates/docs.md:289:are installed
[FAIL] The passive rate for 08.Server-integration/03.Client-certificates/docs.md:
	passive-words/total-lines = 20/289 ~= 6 - and thus above the cut-off rate: 3
```

08.Server-integration/03.Client-certificates/docs.md
kacf: Perhaps I should be more clear: What I meant that was a section needs to be written to set up the ambassador part with the certificate. I'll update the comment to reflect this.

08.Server-integration/03.Client-certificates/docs.md
kacf: But does the client care whether it's trusted? If I've understood correctly, all trust checking happens on the server.

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+
+- [Client installation](../../03.Client-installation/chapter.md)
+- [System updates: Debian family](../../04.System-updates-Debian-family/chapter.md)
+- [System updates: Yocto Project](../../05.System-updates-Yocto-Project/chapter.md)
+
+### Mender client and server connectivity
+
+<!-- TODO, this section may need to be rewritten slightly. It's not really incorrect, but it doesn't fit perfectly either, since we will set up the ambassador. -->
+
+Once your device boots with a newly provisioned disk image, it should already be correctly connecting to the Mender server. After booting the device you see it pending authorization in the Mender server UI, similar to the following.
```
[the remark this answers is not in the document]
kacf: I think this doesn't carry the same meaning. In fact, I would go in the opposite direction, and change

* After booting the device you see it pending authorization"

-into-

* After booting the device you **should** see it pending authorization

The important point here is that these are things you need to check, because there are no preceding instructions, so we can't automatically expect it to be true.

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+
+![Mender UI - device pending authorization](device-pending-authorization.png)
+
+If your device does not show as pending authorization in the Mender server once it is booted with the disk image, you need to diagnose this issue before continuing. See the [troubleshooting section on connecting devices](../../201.Troubleshoot/05.Device-Runtime/docs.md#mender-server-connection-issues) in this case.
+
+
+### A CLI environment for your server
+
+In order to access the the Mender server API with the commands below, you need to set up some shell variables in the terminal you will be using.
```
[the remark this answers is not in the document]
kacf: Yep, both shorter and better!

08.Server-integration/03.Client-certificates/docs.md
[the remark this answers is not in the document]
oleorhagen: Makes sense :)

08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -0,0 +1,289 @@
...
+```bash
+openssl req -new -x509 -key ca-private.key -out ca-cert.pem -config ca-cert.conf -days $((365*10))
+```
+
+! The `-days` argument specifies how long the certificate is valid, and can be adjusted if needed. The example expression gives a certificate which is valid for approximately 10 years. Since the CA certificate will only be used on the Mender server, it is usually not important that it expires, and it's better to have a long expiry time to avoid having to rotate certificates on the devices.
+
+
+### Generate a client certificate
+
+When preparing client certificate for a device, the certificate key will be generated on a separate system (not on the device), and then provisioned into the device storage. This way we can keep records of the public key of the device and ensure sufficient entropy during key generation, so the resulting keys are secure random.
```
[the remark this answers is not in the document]
kacf: In this case, I think we should use "can" or some equivalent, because we are not instructing the user, just vaguely mentioning something they can and possibly should do.

So I think "can keep" is better, but is using "we" bad? Should that be written differently?

08.Server-integration/03.Client-certificates/docs.md
kacf: Omg, I know! This is like newline endings all over again. One design flaw, and it will haunt developers for decades.

08.Server-integration/03.Client-certificates/docs.md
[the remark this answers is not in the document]
oleorhagen: Not necessarily. [This](https://github.com/mendersoftware/mender-docs/blob/master/README-styleguide.markdown#use-second-person-you-when-speaking-to-or-about-the-reader) is what our styleguide says.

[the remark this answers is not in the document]
kacf: Thanks, will use "you" form.

kacf:
Updated!

kacf:
Ok, only one failure left, which I don't care about, since it is in the server section, which needs to be rewritten. Merging.

08.Server-integration/03.Client-certificates/docs.md
kacf: @tranchitella: I did not get an answer to this. If you still think I should change something here, let me know, and I will follow up with a new PR.

PR #1149, opened 2020-08-25
Remove "./" prefix for relative links in the directory

lluiscampos:
As these seem to not render correctly with the new version of Grav


mirzak:
Nice, thanks for taking care of this.

mirzak (approved):

PR #1157, opened 2020-08-28
Move root document to "home" directory

lluiscampos:
This is what the new Mender Docs Site theme expects. Currently there is
a workaround on build time but has inconvenient side effects for local
development.


lluiscampos:
Depends on https://github.com/mendersoftware/mender-docs-site/pull/270. Do not merge before such dependency is merged and ported to production.

mzedel (approved):

PR #1158, opened 2020-08-28
Redirect old url to new api site

michaelatmender:
Changelog: none

Signed-off-by: Michael Clelland <michael.clelland@mender.io>

### Description

Thank you!

### DIFF
--- a/apis/docs.md
+++ b/apis/docs.md
@@ -1,0 +1,9 @@
+---
+title: API
+redirect: /api/
+---
+<!--AUTOVERSION: "redirect: /../api/development/"/integration/complain-->
+<!--
+Exception to the rule about AUTOVERSION tags coming before their affected block:
+For page headers the tag may come after due to misrendering if it is above.
+-->


apis/docs.md
```diff
@@ -0,0 +1,9 @@
+---
+title: API
+redirect: /api/
```
lluiscampos, line 3: If any, this needs to be a relative path like `../api` to work.

But anyway, my understanding is that we need to fix this at the gateway level, as a request to `/apis` will still be redirected to the `mender-api-docs` container from there.

michaelatmender: > If any, this needs to be a relative path like ../api to work.
Paths like this work fine in Grav – writing it this way just means it will redirect to `{ home url }/api`

I think I see what you mean about the second part though, I take it that anything beginning with "api" is not even reaching Grav, so the change in this PR would be redundant?

lluiscampos: > Paths like this work fine in Grav – writing it this way just means it will redirect to `{ home url }/api`

That used to work, but with the new theme (or new Grav version) does not work any more.

> [...] so the change in this PR would be redundant?

Yes, I think so. We need to follow-up with SRE team, adding such redirect in `sre-tools` repo.

lluiscampos:
Fixed from the gateway side

PR #1160, opened 2020-08-31
fix link in 06.Artifact-creation/04.State-scripts/docs.md

mirzak:


mirzak:
Based on the following report: https://www.deadlinkchecker.com/reports/docs-mender-io-20200831052911.html

lluiscampos (approved):

PR #1166, opened 2020-09-08
Include extra OpenSSL error messages in the trouble shooting section

merlin-northern:
ChangeLog:none

### DIFF
--- a/201.Troubleshoot/03.Mender-Client/docs.md
+++ b/201.Troubleshoot/03.Mender-Client/docs.md
@@ -121,6 +121,38 @@ If these mismatch, then you need to update `/etc/mender/server.crt` on your clie
 You can do this manually for testing purposes, and you should
 [include the certificates in your Yocto Project build](../../05.System-updates-Yocto-Project/06.Build-for-production/docs.md#including-the-client-certificates).
 
+## Depth zero self-signed certificate, openssl verify rc: 18
+
+The Mender Client detected a self-signed certificate that is the only one in the chain
+and the same certificate can't be located in the trusted store. That means
+the OpenSSL is unable to verify the server identity. You have to either use
+another (not self-signed) certificate or include the certificate in the local trust store.
+
+## End entity key too short, openssl verify rc: 66
+
+The key length for the end entity in the certificate chain is too short.
+This can happen in conjunction with the security level setting in the OpenSSL
+configuration file, and the actual key length. First check if the security level
+setting is what you want it to be. Start by locating the `openssl.cnf` file,
+by running:
+
+```bash
+# openssl version -d
+OPENSSLDIR: "/opt/local/etc/openssl"
+```
+
+In the above example you can find the configuration file
+at `/opt/local/etc/openssl/openssl.cnf` and check the security level:
+
+```bash
+# cat `openssl version -d | sed -e 's/.*"\([^"]*\)".*/\1/'`/openssl.cnf | grep SECLEVEL
+CipherString = DEFAULT@SECLEVEL=2
+```
+A security level of `2` is the default one on many installations nowadays.
+<!--AUTOVERSION: "Starting with the Mender Client %,"/ignore-->
+Starting with the Mender Client 2.4.0, the client uses OpenSSL and it is possible
+that you see this error with shorter keys and certain values of security level.
+You have two choices: make the keys longer, or decrease the security level.
 
 ## The Current Software installed on my device has `_INCONSISTENT` appended to it
 


merlin-northern:
@kacf this is the one we discussed back in early August, now I think it should be in, as we are with OpenSSL full time.

merlin-northern:
@lluiscampos this is something that would be nice to have somewhere around the release. it is not urgent, but we should mention the new errors that can appear which are related to the port to OpenSSL and the security level.

201.Troubleshoot/03.Mender-Client/docs.md
```diff
@@ -121,6 +121,36 @@ If these mismatch, then you need to update `/etc/mender/server.crt` on your clie
 You can do this manually for testing purposes, and you should
 [include the certificates in your Yocto Project build](../../05.System-updates-Yocto-Project/06.Build-for-production/docs.md#including-the-client-certificates).
 
+## Depth zero self-signed certificate, openssl verify rc: 18
+
+The Mender Client detected a zero depth self-signed certificate, and
+the OpenSSL code cant verify the server identity. You have to either use
```
kacf, line 127: `s/cant/can't/`

201.Troubleshoot/03.Mender-Client/docs.md
```diff
@@ -121,6 +121,36 @@ If these mismatch, then you need to update `/etc/mender/server.crt` on your clie
...
+a valid certificate or include the certificate in the local trust store.
```
kacf, line 128: I'm not really clear on why a zero depth certificate is not a valid certificate. Can you explain a bit more why it isn't valid, and what needs to change for it to become valid?

201.Troubleshoot/03.Mender-Client/docs.md
```diff
@@ -121,6 +121,36 @@ If these mismatch, then you need to update `/etc/mender/server.crt` on your clie
...
+
+In the above example you can find the configuration file
+at `/opt/local/etc/openssl/openssl.cnf` and check the security level:
+
+```bash
+# cat `openssl version -d | sed -e 's/.*"\([^"]*\)".*/\1/'`/openssl.cnf | grep SECLEVEL
+CipherString = DEFAULT@SECLEVEL=2
+```
+The security level of `2` is the default one on many installations nowadays.
+With the Mender Client ported to OpenSSL it is possible that you see this error
```
kacf, line 151: Suggestion: "Starting with Mender client 2.4.0, the client uses OpenSSL and it is possible..."

kacf (changes requested):
Good to have this covered!

201.Troubleshoot/03.Mender-Client/docs.md
```diff
@@ -121,6 +121,36 @@ If these mismatch, then you need to update `/etc/mender/server.crt` on your clie
...
+```
+
+In the above example you can find the configuration file
+at `/opt/local/etc/openssl/openssl.cnf` and check the security level:
+
+```bash
+# cat `openssl version -d | sed -e 's/.*"\([^"]*\)".*/\1/'`/openssl.cnf | grep SECLEVEL
+CipherString = DEFAULT@SECLEVEL=2
+```
+The security level of `2` is the default one on many installations nowadays.
```
lluiscampos, line 150:
```suggestion
The security level `2` is the default one on many installations nowadays.
```
or
```suggestion
A security level of `2` is the default one on many installations nowadays.
```

lluiscampos (changes requested):
Thanks. I will bring this to 2.5.x release branch

201.Troubleshoot/03.Mender-Client/docs.md
merlin-northern: fixed.

201.Troubleshoot/03.Mender-Client/docs.md
merlin-northern: neither am I. my wife tells me that when I try to explain something no one understands.
fixed. is it better now?

201.Troubleshoot/03.Mender-Client/docs.md
merlin-northern: sure.

201.Troubleshoot/03.Mender-Client/docs.md
merlin-northern: chose second one.

201.Troubleshoot/03.Mender-Client/docs.md
kacf: Yes, that's better, thanks!

kacf (approved):

lluiscampos (approved):

PR #1167, opened 2020-09-09
fix: Wrong name for the server key and certificate generation

oleorhagen:
I expect this is just a typo?


oleorhagen:
I will cherry-pick to `2.5` if this is correct.

tranchitella (approved):
well spotted!

PR #1175, opened 2020-09-16
MEN-3927 MEN-3955 Document HttpsClient directive in mender.conf

merlin-northern:
ChangeLog:none

### DIFF
--- a/03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
+++ b/03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
@@ -13,6 +13,61 @@ enables signed-updates-only mode when it is set. If set the client will reject
 incorrectly signed updates, or updates without a signature. See also the section
 about [signing and verification](../../../06.Artifact-creation/07.Sign-and-verify/docs.md).
 
+#### HttpsClient
+
+Allows you to configure the certificate, private key and, SSL Engine id to use
+during the SSL handshake. If you provide the certificate and private key
+as locally accessible files you don't have to specify SSLEngine.
+If you want to use a Hardware Security Module (HSM) you can provide the private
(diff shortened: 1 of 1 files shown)


mchalski (approved):
lgtm. 

I could have some style nitpicks (active/passive etc.) if you're interested:
`In case both the certificate and the private key are
￼stored in locally accessible files there is no need to provide SSLEngine.` 
could become
`If you provide the certificate and private key as locally accessible files you don't have to specify SSLEngine`

`If you want to use a Hardware Security Module (HSM) you can provide the private
key as a [PKCS#11 URI](https://tools.ietf.org/html/rfc7512) and in that case
you must specify the SSLEngine`
could become

`The above can have the following example OpenSSL configuration present:`
could become
`The above can be accompanied by the following OpenSSL configuration:`

but take with a grain of salt and perhaps wait for someone else's corrections.

merlin-northern:
> lgtm.
> 
> I could have some style nitpicks (active/passive etc.) if you're interested:
> `In case both the certificate and the private key are ￼stored in locally accessible files there is no need to provide SSLEngine.`
> could become
> `If you provide the certificate and private key as locally accessible files you don't have to specify SSLEngine`

even better!

> `If you want to use a Hardware Security Module (HSM) you can provide the private key as a [PKCS#11 URI](https://tools.ietf.org/html/rfc7512) and in that case you must specify the SSLEngine`
> could become

thanks!

> 
> `The above can have the following example OpenSSL configuration present:`
> could become
> `The above can be accompanied by the following OpenSSL configuration:`

reworded to avoid the passive :)

> but take with a grain of salt and perhaps wait for someone else's corrections.

makes sense! corrected!

tranchitella (approved):

PR #1196, opened 2020-10-01
MEN-3953: update instuctions to download `mender-binary-delta`

tranchitella:
Changelog: none

### DIFF
--- a/05.System-updates-Yocto-Project/05.Customize-Mender/01.Delta-update-support/docs.md
+++ b/05.System-updates-Yocto-Project/05.Customize-Mender/01.Delta-update-support/docs.md
@@ -29,6 +29,9 @@ HOSTED_MENDER_EMAIL="myusername@example.com"
 curl -u $HOSTED_MENDER_EMAIL -O https://download.mender.io/hosted/content/mender-binary-delta/1.1.0/mender-binary-delta-1.1.0.tar.xz
 ```
 
+Replace the value of `HOSTED_MENDER_EMAIL` with the email address you used to sign up on *Hosted Mender*, then enter your Hosted Mender password when prompted to proceed.
+**NOTE**: if you signed up using your Google or GitHub login, use the email address linked to that account and enter `x` as the password.
+
 On the other hand, if you are using *on-premise Mender Enterprise*, download using the following
 command:
 


05.System-updates-Yocto-Project/05.Customize-Mender/01.Delta-update-support/docs.md
```diff
@@ -20,24 +20,15 @@ filesystem](../../04.Image-customization/02.Read-only-root-filesystem/) section.
...
+Download the `mender-binary-delta` archive with the following command:
 
 <!--AUTOVERSION: "mender-binary-delta/%/mender-binary-delta-%.tar"/mender-binary-delta-->
 ```bash
 HOSTED_MENDER_EMAIL="myusername@example.com"
 curl -u $HOSTED_MENDER_EMAIL -O https://download.mender.io/hosted/content/mender-binary-delta/1.1.0/mender-binary-delta-1.1.0.tar.xz
 ```
 
-On the other hand, if you are using *on-premise Mender Enterprise*, download using the following
-command:
```
lluiscampos, line 33: So what happens now with the on-prem Enterprise installs?

lluiscampos (commented):
A question below.

Also, we have to update [this article at Mender Hub](https://hub.mender.io/t/robust-delta-update-rootfs/1144) once this change goes live.

05.System-updates-Yocto-Project/05.Customize-Mender/01.Delta-update-support/docs.md
tranchitella: The documentation is misleading, because as of today, if you are on-prem user, you cannot download the tool using these instructions because the authentication is pointing to Hosted Mender only. No idea why we've added that paragraph, it is completely misleading. To answer your question: you sign up on Hosted Mender.

mirzak: > you cannot download the tool using these instructions because the authentication is pointing to Hosted Mender only

Did we removed on-premise Enterprise evaluation?

tranchitella: I don't know if we removed on-premise Enterprise evaluation, but what I stated is that you cannot download the binary without a Hosted mender user, as of today.

mirzak: > I don't know if we removed on-premise Enterprise evaluation, but what I stated is that you cannot download the binary without a Hosted mender user, as of today.

Might not be a big problem, but I find this workflow to be a bit strange for all our on-premise users. 

As far as I understand all on-premise Enterprise users, get an account to login to the private docker registry. This same account they previously could use to download the mender-binary-delta package. Instead now they need to create a trial account on hosted Mender to download this. 

To be honest, I am pondering if we should just make the link public and distribute an EULA together with the binaries. It is very easy to get access to anyways.

Input from @estenberg would be valuable.

tranchitella: @mirzak the account used to log in to the private docker registry cannot be used to download the mender-binary-delta package. This is not implemented, as of today. That's why I suggest to remove the paragraph, it is simply wrong.

mirzak: > @mirzak the account used to log in to the private docker registry cannot be used to download the mender-binary-delta package. This is not implemented, as of today.

Yes, I understand that, but my questions is why we have changed this?

In the end I do not want to block this PR if this the state of things today. But it might be worth a follow up discussion.

tranchitella: I have the impression it never worked as described, this is not something we changed.
The only way you could download the binary, now and in the past, is using a Hosted Mender user.

estenberg: I would double-check this; I am pretty sure that we had a way to grant access to on-premise Enterprise users. You can see the URLs are different for hosted and on-premise, and I believe the on-premise one was managed manually.

We still have to manage manually the access to Enterprise docker registries for on-premise? If it's just about adding access one more place when creating an on-premise Enterprise account, I think it's worth keeping that. But I would not spend a lot of time on this.

This might seem like a longer discussion, I would suggest:
* Merge this
* I will follow up to see how we create on-premise Enterprise accounts. What is the process? Add a step to create credentials & send link to this.

estenberg (approved):

05.System-updates-Yocto-Project/05.Customize-Mender/01.Delta-update-support/docs.md
tranchitella: I'm really sorry, but it was my mistake since the beginning: I did not notice the two URLs are slightly different, the second one is using /content instead of hosted/content, and it makes use of a password file. Updating such a file for ME on-prem users is a manual process, indeed. I've updated my PR to restore the paragraph about on-prem, and limited it to the paragraph which suggests using `x` as the password if logging in using GitHub/Google.

I'm sorry for the noise.

tranchitella:
@mirzak @lluiscampos can you please review this new version? See my comment above.

lluiscampos (approved):

PR #1201, opened 2020-10-06
Fix typos

tranchitella:
Changelog: none


lluiscampos (approved):
Thanks

PR #1204, opened 2020-10-08
MEN-4020: Enable the passive-voice checker in CI

oleorhagen:
* Some minor cleanups of the `passive-voice.sh` script
* A new script to select the files to be checked using a blacklist
* A new job in Gitlab-CI to enforce this on master

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -41,6 +41,12 @@ test:internal-links:
     - npm install
     - ./checklinks --color --verbose
 
+test:passive-voice:
+    stage: test
+    image: ubuntu
+    script:
+        - bash ./scripts/run-passive.sh
+
 test:latest_versions:
   stage: test
   image: alpine
--- a/scripts/run-passive.sh
+++ b/scripts/run-passive.sh
@@ -18,7 +18,7 @@
 
 
 # List all files not in caught by the blacklist
-files=$(find . -type f -regex './[0-9]+.*\.md' | grep -vFf .passive-blacklist)
+files=$(find . -type f -regex './[0-9]+.*\.md' | grep -vFf .passive-blacklist | sort)
 
 EXIT_CODE=0
 
--- a/scripts/run-passive.sh
+++ b/scripts/run-passive.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+#
+# This script collects all markdown files in the documentation repositories, and
+# filters them through the blacklist, and then runs the `passive.sh` script on
+# them.
+#
+# Usage:
+#
+# ./scripts/run-passive.sh [--verbose]
+#
+# Note:
+#
+# The script automatically reads in the .passive-blacklist from the root
+# directory, and applies the entries as a filter to a grep expression. Have a
+# look at the `.passive-blacklist` file in the root directory for an example of
+# how it works.
+#
+
+
+# List all files not in caught by the blacklist
+files=$(find . -type f -regex './[0-9]+.*\.md' | grep -vFf .passive-blacklist)
+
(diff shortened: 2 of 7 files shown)


scripts/run-passive.sh
```diff
@@ -0,0 +1,36 @@
...
+
+if [[ "$1" == "--verbose" ]]; then
+  VERBOSE="--verbose"
+fi
+
+for file in $files; do
+  if ! ./scripts/passive.sh $VERBOSE $file; then
+    EXIT_CODE=1
+  fi
+done
```
lluiscampos, line 33: Why don't just:
```suggestion
if ! ./scripts/passive.sh $VERBOSE $files; then
  EXIT_CODE=1
fi
```

AFAICS, `passive.sh` script accepts a list of files.

scripts/run-passive.sh
```diff
@@ -0,0 +1,36 @@
...
+
+
+# List all files not in caught by the blacklist
+files=$(find . -type f -regex './[0-9]+.*\.md' | grep -vFf .passive-blacklist)
+
+EXIT_CODE=0
+
+if [[ "$1" == "--verbose" ]]; then
+  VERBOSE="--verbose"
+fi
```
lluiscampos, line 27: Also, instead of having an explicit --verbose catch here and propagating it to `passive.sh`, consider just passing `$@`. This way we don't need to touch this CI script when adding flags to the `passive.sh` one.

lluiscampos (approved):
Looks good!

Couple of suggestions, but it is up to you to follow-up or ignore.

scripts/run-passive.sh
oleorhagen: Excellent suggestion. I was thinking of doing this initially. However, the `passive.sh` script accepts a list of files, but it exits on the first file which does not pass. I was thinking about this a bit. And I do think this is fair behaviour. So therefore, we call the files individually, so that in CI, all the failures show up. When running this from the CLI locally, having the first one fail, is sort of nice, as it allows you to weed out the failures one by one.

I'm open to other approaches though

lluiscampos: That makes sense and it is a good idea!

We actually had the same situation with the AUTOVERSION tool in this repo (i.e. CI exited on first failure) and was a bad experience (you as a PR author don't get the full picture at once).

So yes, leave as it is :+1:

oleorhagen:
Also, the CI passed ,which it is not supposed to. There are some files in Overview, which fail. Leaving it for now, to get red CI before merging.

.gitlab-ci.yml
```diff
@@ -41,6 +41,13 @@ test:internal-links:
     - npm install
     - ./checklinks --color --verbose
 
+test:passive-voice:
+    stage: test
+    image: ubuntu
+    script:
+        # For the Bash image, bash does not live in '/bin'.
```
lluiscampos, line 48: This comment does not make sense now after last force-push

oleorhagen: Yup. WIP. I'll ping you when it is ready. I'll leave the errors in for now. To make sure it works. I chose ubuntu, as the stripped down images lack the CLI flags I need for fancy stuff, like `egrep --line-number` etc. 

Do you have a suggestion for having verbose as an easy to add option in the `.gitlab-ci` file. Or is this unnecesary you think?

lluiscampos: I think is unnecessary, I wouldn't bother.

oleorhagen:
@lluiscampos Ready for you to tear it apart :)

lluiscampos (approved):
@oleorhagen Nothing to tear apart, looks good.

PR #1208, opened 2020-10-12
Server-installation/Upgrading: add information about upgrading MongoDB

kjaskiewiczz:

### DIFF
--- a/07.Server-installation/07.Upgrading/docs.md
+++ b/07.Server-installation/07.Upgrading/docs.md
@@ -52,6 +52,136 @@ These can sometimes happen due to device decommissioning.
 You can find instructions on how to clean up the deviceauth database
 in the [Troubleshoot](../../201.Troubleshoot/04.Mender-Server/docs.md) chapter.
 
+## Upgrading MongoDB
+
+Mender Server 2.4.\* uses MongoDB 3.6 and Mender Server 2.5.\* uses MongoDB 4.4.
+Before upgrading Mender Server you must upgrade MongoDB from version 3.6 to version 4.4.
+It is not possible to upgrade MongoDB directly from version 3.6 to version 4.4.
+Instead, you should upgrade MongoDB from 3.6 to 4.0, then from 4.0 to 4.2
+and finally from 4.2 to 4.4.
+
+MongoDB provides instruction on how to perform the upgrades:
+* [upgrade MongoDB to 4.0](https://docs.mongodb.com/manual/release-notes/4.0/#upgrade-procedures)
+* [upgrade MongoDB to 4.2](https://docs.mongodb.com/manual/release-notes/4.2/#upgrade-procedures)
+* [upgrade MongoDB to 4.4](https://docs.mongodb.com/manual/release-notes/4.4/#upgrade-procedures)
+
+If you need more guidance please look at the description below.
+
+### Upgrading MongoDB from version 3.6 to version 4.0
+
+To upgrade MongoDB from version 3.6 to version 4.0 check the `featureCompatibilityVersion` parameter.
+The MongoDB 3.6 instance must have `featureCompatibilityVersion` set to `3.6`.
+
+To check `featureCompatibilityVersion`:
+```
+db.adminCommand( { getParameter: 1, featureCompatibilityVersion: 1 } )
+```
+
+If you are using docker-compose based setup you can check it by running docker exec command:
+```bash
+docker exec -it menderproduction_mender-mongo_1 mongo --eval "db.adminCommand( { getParameter: 1, featureCompatibilityVersion: 1 } )"
+```
+The operation should return a result that includes
+> ```
+> { "featureCompatibilityVersion" : { "version" : "3.6" }, "ok" : 1 }
+> ```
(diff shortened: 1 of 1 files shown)


07.Server-installation/07.Upgrading/docs.md
```diff
@@ -52,6 +52,19 @@ These can sometimes happen due to device decommissioning.
...
+Before upgrading Mender Server you must upgrade MongoDB from version 3.6 to version 4.4.
+It is not possible to upgrade MongoDB directly from version 3.6 to version 4.4.
+Instead, you should upgrade MongoDB from 3.6 to 4.0, then from 4.0 to 4.2
+and finally from 4.2 to 4.4.
+
+MongoDB provides instruction on how to perform the upgrades:
+* [upgrade MongoDB to 4.0](https://docs.mongodb.com/manual/release-notes/4.0/#upgrade-procedures)
+* [upgrade MongoDB to 4.2](https://docs.mongodb.com/manual/release-notes/4.2/#upgrade-procedures)
+* [upgrade MongoDB to 4.4](https://docs.mongodb.com/manual/release-notes/4.4/#upgrade-procedures)
+
```
mirzak, line 67: I, who know very little about MongoDB, have a very hard time following the linked instructions and how to translate them to a Mender server environment.

tranchitella (changes requested):
I believe we have to provide a bit of guidance regarding the upgrade of MongoDB. At minimum, we should tell the user to manually bump the image tag of mongodb to the next version, start it using docker-compose, enter the container and upgrade the compatibility version using the admin command. This has to be repeated for each major version until it reaches MongoDB 4.4.

Can you please add commands and explanations on how to perform these upgrades?

kjaskiewiczz:
@tranchitella @mirzak I'll provide a more detailed description

kjaskiewiczz:
@tranchitella @mirzak I've updated the docs; it still needs some polishing so please wait with the review, but it would be great if you could just take a look and say if I wasn't too verbose

07.Server-installation/07.Upgrading/docs.md
```diff
@@ -52,6 +52,136 @@ These can sometimes happen due to device decommissioning.
...
+The MongoDB 3.6 instance must have `featureCompatibilityVersion` set to `3.6`.
+
+To check `featureCompatibilityVersion`:
+```
+db.adminCommand( { getParameter: 1, featureCompatibilityVersion: 1 } )
+```
+
+If you are using docker-compose based setup you can check it by running docker exec command:
+```bash
+docker exec -it menderproduction_mender-mongo_1 mongo ir --eval "db.adminCommand( { getParameter: 1, featureCompatibilityVersion: 1 } )"
```
tranchitella, line 82: What is the `ir` in the command line?

mirzak:
> @tranchitella @mirzak I've updated the docs; it still needs some polishing so please wait with the review, but it would be great if you could just take a look and say if I wasn't too verbose

The approach looks good to me.

07.Server-installation/07.Upgrading/docs.md
kjaskiewiczz: name of the database on which I was experimenting with features incompatible across different versions of MongoDB;
I've already removed it; thanks for pointing this out; I was copy/pasting from the wrong terminal;

tranchitella (approved):
Great, well done!

mirzak (approved):
Looks good. I guess this needs to be cherry-picked to 2.5.x branch?

kjaskiewiczz:
@mirzak yes, this should be in the 2.5.x for sure.
I was wondering if this change should be in the master branch?
If master is development then what upgrade really means? Upgrading from last stable to development? If that's the case then the last stable is 2.5 and the section about upgrading MongoDB shouldn't be there.

07.Server-installation/07.Upgrading/docs.md
```diff
@@ -52,6 +52,136 @@ These can sometimes happen due to device decommissioning.
 You can find instructions on how to clean up the deviceauth database
 in the [Troubleshoot](../../201.Troubleshoot/04.Mender-Server/docs.md) chapter.
 
+## Upgrading MongoDB
```
estenberg, line 55: Would rephrase to "Upgrade to MongoDB 4.4"

This is section is version-specific but merged to master, so we'll need to remove it in a while (maybe a year or so from now, or when we do the next major DB upgrade).

estenberg (approved):

mirzak:
> @mirzak yes, this should be in the 2.5.x for sure.
> I was wondering if this change should be in the master branch?

I think I would also question if this really should be in master at all. Makes most sense if goes directly to 2.5.x.

kjaskiewiczz:
@mirzak @estenberg change has been merged to 2.5.x and now I need a final decision if I should merge to master
in my opinion, we don't need it in master branch

mirzak:
As this is not applicable in master, I would say that we close this one without merge

estenberg:
:+1: makes sense to me as well, it should be a temporary version-specific problem only.

PR #1209, opened 2020-10-12
MEN-3924 Allow to load private key from the Security conf section

merlin-northern:
ChangeLog:none

### DIFF
--- a/03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
+++ b/03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
@@ -20,10 +20,14 @@ verification](../../../06.Artifact-creation/07.Sign-and-verify/docs.md).
 
 Allows you to configure the certificate, private key and, SSL Engine id to use
 during the SSL handshake. If you provide the certificate and private key
-as locally accessible files you don't have to specify SSLEngine.
+as locally accessible files you don't have to specify
+<!--AUTOVERSION: "docs/man%"/ignore-->
+[SSLEngine](https://www.openssl.org/docs/man1.1.1/man1/engine.html).
 If you want to use a Hardware Security Module (HSM) you can provide the private
 key as a [PKCS#11 URI](https://tools.ietf.org/html/rfc7512) and in that case
-you must specify the `SSLEngine`.
+you must also specify the `SSLEngine`. The client will use this key for signing
+the authorization requests unless you provide `Security.AuthPrivateKey`
+(see below).
 
 ##### Certificate
 
@@ -109,6 +113,23 @@ mender-convert) sets this variable so it is rarely modified manually.
 
 The Linux device that contains root filesystem B. The build system (ie Yocto or
 mender-convert) sets this variable, so it is rarely modified manually.
+
+#### Security
+
+Allows you to specify the basic security options, `AuthPrivateKey`
+and `SSLEngine`, which you can use for signing of authentication requests.
+If you specify both `Security.AuthPrivateKey` and `HttpsClient.Key`
+the former takes precedence.
+
+##### AuthPrivateKey
+
+A path to the file in pem format holding the private key, or a
+[PKCS#11 URI](https://tools.ietf.org/html/rfc7512).
+
+##### SSLEngine
+
+<!--AUTOVERSION: "docs/man%"/ignore-->
+The [SSLEngine](https://www.openssl.org/docs/man1.1.1/man1/engine.html) to use.
 
 #### Servers
 


03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
```diff
@@ -110,6 +112,21 @@ mender-convert) sets this variable so it is rarely modified manually.
 The Linux device that contains root filesystem B. The build system (ie Yocto or
 mender-convert) sets this variable, so it is rarely modified manually.
 
+#### Security
+
+Allows you to specify the basic security options: `PKIPrivateKey`
+and `SSLEngine` which you can use for signing of authentication requests.
```
kacf, line 118: Small grammar tweak, I would use this punctuation: ``...basic security options, `PKIPrivateKey` and `SSLEngine`, which you...`` (note the two new commas and removed colon).

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
```diff
@@ -110,6 +112,21 @@ mender-convert) sets this variable so it is rarely modified manually.
...
+If you specify both `HttpsClient.Key` and `Security.PKIPrivateKey`
+the former takes precedence over for signing.
```
kacf, line 120: Actually it means that the key specified here won't be used, or have I misunderstood? If so, you can just say "...takes precedence ~over for signing~.".

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
```diff
@@ -110,6 +112,21 @@ mender-convert) sets this variable so it is rarely modified manually.
...
+
+##### PKIPrivateKey
+
+A path to the file in pem format holding the private key, or  a PKCS#11 URI.
```
kacf, line 124: Nitpick: double space.

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
```diff
@@ -110,6 +112,21 @@ mender-convert) sets this variable so it is rarely modified manually.
...
+
+##### SSLEngine
+
+The SSLEngine to use.
```
kacf, line 128: For this one, and PKCS#11, is there anything you can link to? Even Wikipedia might be better than nothing.

kacf (approved):
Looks good, consider my comments below as suggestions.

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
kacf, line 120: Actually, reading your [mender PR](https://github.com/mendersoftware/mender/pull/617), I realize that this is wrong. You're stating the opposite, it is the latter which takes precedence.

kacf (changes requested):

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
merlin-northern: wow, you are right.

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
merlin-northern: of course, I even linked that in the HttpsClient subsection.

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
merlin-northern: the HttpsClient.Key will not be used for signing if the Security.PKIPrivateKey is given. but HttpsClient.Key will still be used in the SSL handshake.

merlin-northern: so I am keeping "over for signing", ok?

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
```diff
@@ -110,6 +114,23 @@ mender-convert) sets this variable so it is rarely modified manually.
...
 mender-convert) sets this variable, so it is rarely modified manually.
 
+#### Security
+
+Allows you to specify the basic security options, `PKIPrivateKey`
+and `SSLEngine`, which you can use for signing of authentication requests.
+If you specify both `Security.PKIPrivateKey` and `HttpsClient.Key`
+the former takes precedence over for signing.
+
+##### PKIPrivateKey
```
mirzak, line 124: What is the purpose of having `PKI` prefix here?

merlin-northern: I wanted some prefix to point out that it is used in signing and in case we will introduce some other keys in that section -- I am not that bound to it -- do you think it would look better without it? if you really have a strong opinion here, then I will have to change client patch as well.

mirzak: I find it confusing,

From my understanding, PKI relates to certificates and this key will normally be used when you **do not** have an PKI (certificates)

`AuthPrivateKey` might be a better description if you want to have some kind of prefix?

merlin-northern: all right. so AuthPrivateKey it is.

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
kacf: The logic is correct now, but I would still delete "~over for signing~" because the grammar is a bit off. I think just saying "...the former takes precedence." is enough.

03.Client-installation/06.Configuration-file/50.Configuration-options/docs.md
```diff
@@ -110,6 +114,23 @@ mender-convert) sets this variable so it is rarely modified manually.
...
+
+##### AuthPrivateKey
+
+A path to the file in pem format holding the private key, or a
+[PKCS#11 URI](https://tools.ietf.org/html/rfc7512).
+
+##### SSLEngine
+
+<!--AUTOVERSION: "docs/man%"/ignore-->
+The [SSLEngine](https://www.openssl.org/docs/man1.1.1/man1/engine.html) to use.
```
kacf, line 132: :+1:

kacf (changes requested):

mirzak (approved):

kacf (approved):

merlin-northern:
this one was actaully https://tracker.mender.io/browse/MEN-3957

PR #1215, opened 2020-10-15
Update software naming and Depends/Provides info after review. 

kacf:
* Explain what Depends and Provides are in Overview/Artifact chapter.

* Adjust description of Artifact names vs software versions.

* Use different versioning schemes for rootfs and application examples
  to make it easier to visually distinguish the two.

* Provide examples for how to see the version information after the
  Artifact has been installed. This is still missing screenshots; I
  have marked them TODO for now.

* Changed "myfs" to "data-partition" to match what is in the
  preliminary UI Design (this is not enforced in any way, so it's just
  to reduce confusion, both labels are technically correct).

* Provide one example for what happens when you have an application on
  the root filesystem and you install a rootfs update.

* Remove '$' from shell examples so that the copy button works
  properly.

At the time of writing, the source for the diagrams can be found in
Northern.tech's Engineering folder:
https://drive.google.com/drive/folders/0B6EXyILUYX54TWR5THZVbUZNR3M


kacf:
If this is approved I will create a task to fix the TODOs.

drewmoseley (approved):
Looks good.

06.Artifact-creation/09.Software-versioning/docs.md
```diff
@@ -24,92 +26,155 @@ For example, to explicitly set the software version for a full system update to
...
+```console
+data-partition.myapp.version=v2020.10
+```
+
+Since we have selected `data-partition` as the filesystem which hosts the software, installing this Artifact will have no effect on the `rootfs-image.version`.
+
+![Effect on versions when installing an application Artifact](versions-and-installing-application-artifact.png)
+*Effect on versions when installing an application Artifact.*
+
+### Inspect the versions
```
estenberg, line 126: This is a bit of duplication from above.

You can assume that people know the background here, and just say something along the lines, "If we now inspect the versions installed like described above we get.. mender-artifact -show-provides.. output.."

estenberg (approved):
I think it is good as well. It contains all the information and is written in a clear way.

The only risk is that it becomes a little too much to understand for a first-time user of it. They probably just want to get to the command of how to set the version (and name) of the software and how it's reported/shown/used in Mender.

However the screenshots might help mitigate this and we can revise it too once we see it rendered, if necessary.

kacf:
> The only risk is that it becomes a little too much to understand for a first-time user of it. They probably just want to get to the command of how to set the version (and name) of the software and how it's reported/shown/used in Mender.

Yes, this is true. But it's hard to trim it down much further without omitting important information about how it works. We could perhaps move the "Advanced usage" header into its own sub-chapter though. Do you think that would help?

Additionally, this is meant as a complete description, but maybe we can add a more trimmed down version somewhere else? How about adding the three arguments to the creation commands in the "Create an Artifact" chapter. We can very briefly describe what they mean, and that they will be reported in the UI, but nothing more. No mechanics or screenshots, just links to read more in this chapter.

drewmoseley:
> We could perhaps move the "Advanced usage" header into its own sub-chapter though. Do you think that would help?

I think this is good.  I also think we should consider having some tutorials documenting specific situations that are likely to be common.  A tutorial could be less formal, ignore irrelevant details for the specific use case, and likely save on support issues.

estenberg:
> > The only risk is that it becomes a little too much to understand for a first-time user of it. They probably just want to get to the command of how to set the version (and name) of the software and how it's reported/shown/used in Mender.
> 
> Yes, this is true. But it's hard to trim it down much further without omitting important information about how it works. We could perhaps move the "Advanced usage" header into its own sub-chapter though. Do you think that would help?

I think maybe it would be trying to cover too much. After thinking a bit more, I think it is best to leave this as an "expert" reference - not for beginners.

> Additionally, this is meant as a complete description, but maybe we can add a more trimmed down version somewhere else? How about adding the three arguments to the creation commands in the "Create an Artifact" chapter. We can very briefly describe what they mean, and that they will be reported in the UI, but nothing more. No mechanics or screenshots, just links to read more in this chapter.

Yes, that's a great idea! The Create an Artifact documentation is for beginners, so it would fit very well there.

kacf:
I moved the Clears Provides documentation into a sub chapter, since it is very advanced, and provided a minimal set of options and a description in the "Create-an-Artifact" section. Only the two latest commits are new. @estenberg: Final ack?

> I think this is good. I also think we should consider having some tutorials documenting specific situations that are likely to be common. A tutorial could be less formal, ignore irrelevant details for the specific use case, and likely save on support issues.

Yep! Actually, I think that the "Create-an-Artifact" chapter now serves as the two base use cases. Take a look.

More use cases could also be added, but I think I'll stop here to avoid exploding this task too much. I guess as we see people use this, we will become more aware of what the typical use cases are and where the pain points are.

06.Artifact-creation/02.Create-an-Artifact-with-system-snapshot/docs.md
```diff
@@ -63,6 +63,7 @@ produces a deployment ready Mender Artifact:
 ```bash
 mender-artifact write rootfs-image -f /mnt/root-part.ext4 \
                                    -n artifact-name \
```
estenberg, line 65: I would use --software-name in these places if that would have the same result as -n. It goes better together with --software-version.

estenberg (approved):
Nice!

Just one minor cosmetic comment if it makes sense.

06.Artifact-creation/02.Create-an-Artifact-with-system-snapshot/docs.md
kacf: No, `-n` (`--artifact-name`) is quite different from `--software-name`. `--artifact-name` provides the name of the Artifact itself, and is required. It also provides the default value for `--software-version` if you have not specified anything, so it's better to think of those as synonymous than `--artifact-name` and `--software-name`. In this sense `--artifact-name` is a bit ambiguous in the sense that we don't quite know if it's a name or a version. But that was exactly the problem we were trying to solve when we came up with the partial naming epic, and we always knew it would not play perfectly with the existing Artifact name (in one of the early proposals, I removed `artifact_name` entirely, but this apparently would cause all kinds of hell on the server, so it was abandoned).

The best thing we can do for the future is probably to distance these two concepts as much as possible from each other. `--artifact-name` is *just* the name of the Artifact, and nothing else.

PR #1216, opened 2020-10-22
[autoversion.py] New option --mender-client-version

lluiscampos:
To be used on client only releases.

### DIFF
--- a/autoversion.py
+++ b/autoversion.py
@@ -353,6 +353,10 @@ def main():
         help="Integration version to update to. Depends on --integration-dir option",
     )
     parser.add_argument(
+        "--mender-client-version",
+        help="Mender client version for client only releases to update to",
+    )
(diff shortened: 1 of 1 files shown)


autoversion.py
```diff
@@ -384,6 +388,15 @@ def main():
             INTEGRATION_REPO = args.integration_dir
             INTEGRATION_VERSION = args.integration_version
 
+        if args.mender_client_version is not None:
```
oleorhagen, line 391: Question:
Is this not the same as:
```
if args.mender_client_version
```

kacf: I think your variant also includes:

```python
if args.mender_client_version is not None and args.mender_client_version != "" and args.mender_client_version != 0
```

kacf (approved):
Nice, thanks! It was even smaller than I expected.

PR #1217, opened 2020-10-26
passive checker: unblacklist 08.Server-integration

mchalski:
https://tracker.mender.io/browse/MEN-4026

### Description

Thank you!

### DIFF
--- a/08.Server-integration/03.Client-certificates/docs.md
+++ b/08.Server-integration/03.Client-certificates/docs.md
@@ -4,9 +4,9 @@ taxonomy:
     category: docs
 ---
 
-Mender supports setting up a reverse proxy at the edge of the network, which can authenticate devices using TLS client certificates. Each client is equipped with a certificate signed by a CA certificate (Certificate Authority), and the edge proxy authenticates devices by verifying this signature. Authenticated devices are automatically authorized in the Mender backend, and do not need manual approval.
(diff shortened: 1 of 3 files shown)


08.Server-integration/03.Client-certificates/docs.md
```diff
@@ -129,12 +129,12 @@ Now that we have both a CA certificate, and a certificate request for the server
...
 ```
 
-! The `-days` argument specifies how long the certificate is valid, and can be adjusted if needed. The example expression gives a certificate which is valid for approximately 2 years.
+! The `-days` argument specifies how long the certificate is valid, and you can adjust it as needed. The example expression gives a certificate which is valid for approximately 2 years.
 
 
 ### Generate a client certificate
 
-When preparing a client certificate for a device, the certificate key is generated on a separate system (not on the device), and then provisioned into the device storage. This way you can keep records of the public key of the device and ensure sufficient entropy during key generation, so the resulting keys are securely random.
+When preparing a client certificate for a device, you generatethe certificate key on a separate system (not on the device), and then provision it into the device storage. This way you can keep records of the public key of the device and ensure sufficient entropy during key generation, so the resulting keys are securely random.
```
drewmoseley, line 137: generate the

drewmoseley (changes requested):

08.Server-integration/03.Client-certificates/docs.md
mchalski: fixed

drewmoseley (approved):

PR #1220, opened 2020-10-28
Brief mention on reserved inventory status attribute in the identity scope

merlin-northern:
ChangeLog:none

### DIFF
--- a/02.Overview/08.Inventory/docs.md
+++ b/02.Overview/08.Inventory/docs.md
@@ -68,3 +68,8 @@ fields to create Dynamic groups (see the screenshot below).
 
 For more information on customizing the Inventory attributes,
 please refer to the [client configuration section](../../03.Client-installation/04.Inventory/docs.md).
+
+!! Currently, there is one "_reserved_" attribute in the *identity* scope: _status_.
+!! It stores the status (`accepted`, `rejected`, etc.) of a device and changes whether
+!! the device status changes. Therefore, you cannot modify it directly.
+


02.Overview/08.Inventory/docs.md
```diff
@@ -68,3 +68,8 @@ fields to create Dynamic groups (see the screenshot below).
 
 For more information on customizing the Inventory attributes,
 please refer to the [client configuration section](../../03.Client-installation/04.Inventory/docs.md).
+
+Currently, there is one "_reserved_" attribute in the *identity* scope: _status_.
+It stores the status (`accepted`, `rejected`, etc.) of a device and changes whether
+the device status changes. Therefore, you cannot modify it directly.
+
```
mirzak, line 75: Would maybe put this in a warning block so it is highlighted, using `!!`

Out of curiosity, what would happen if a device sends `status=something` to the server as part of the inventory attributes? It is just ignored?

merlin-northern: sure.

I like your curiosity. both the inventory and the identity scope attributes are presented in the UI in any case. As @kjaskiewiczz always points out, it is a theoretical surprising behaviour of the API calls that will return status attribute in the identity scope from inventory as a status of a device and not something that in principle user could want to override.

mirzak (approved):

kjaskiewiczz (approved):
looks good

drewmoseley (approved):

PR #1236, opened 2020-11-16
Update all mender bucket URLS to downloads.mender.io

lluiscampos:


mirzak (approved):
Nice :)

lluiscampos:
@mirzak The converted Raspberry Pi images are in other bucket, so that ones we cannot change for now.

estenberg (approved):
Slightly easier to remember. :+1: 

Thanks for taking care of it, a good improvement!

PR #1239, opened 2020-11-21
Update hosted branch after saas-v2020.11.19

lluiscampos:


lluiscampos:
@oleorhagen The second part of 3e064d8 should have been done as part of the client releases earlier this month (the client version itself was updated, but the pre-built images no). How did we miss this?

lluiscampos:
@alfrunes @tranchitella I realized that the [Hosted Mender release checklist](https://docs.google.com/spreadsheets/d/1DhxaSIgYDylVCdDZvNX94U3YUiHr7utHiPPgOY67XXA/edit#gid=0) was not used for this release (hence mender-docs not being updated).

Was this an oversight or rather the checklist does not serve the process? If the latter, we should then bring it back to the table (probably Mender QA meeting) to simplify the checklist or find an alternative, but IMO we should have the process documented somewhere to avoid missing steps (like this PR).

tranchitella:
@lluiscampos sorry, I believe it was an oversight at our end. The checklist is fine, we do not need to improve it.

tranchitella (approved):

mirzak (approved):

PR #1275, opened 2020-12-27
[hosted] Update mtls-ambassador tag to mender-2.5.0

lluiscampos:
This is the tag that we build/test during releases and the one that we
should document for users to pull.

### DIFF
--- a/test_autoversion.py
+++ b/test_autoversion.py
@@ -18,6 +18,16 @@ def test_version_regex():
     assert re.search(regex, "release.2_master_2") is not None
     assert re.search(regex, "morty") is not None
 
+    # Special case: mender-X.Y.Z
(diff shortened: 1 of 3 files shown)


lluiscampos:
See also https://github.com/mendersoftware/integration/pull/1128

lluiscampos:
`autoversion` fix at https://github.com/mendersoftware/mender-docs/pull/1276

test_autoversion.py
```diff
@@ -18,6 +18,16 @@ def test_version_regex():
     assert re.search(regex, "release.2_master_2") is not None
     assert re.search(regex, "morty") is not None
 
+    # Especial case: mender-X.Y.Z
```
tranchitella, line 21: Special :)

tranchitella (approved):

PR #1306, opened 2021-01-19
tests: update number of expected containers

lluiscampos:
Following recent changes in `demo` and production `run` scripts.

PR #1308, opened 2021-01-19
tests: Update (again) the expected containers for Enterprise setup

lluiscampos:
We have added three containers: deviceconnect, nats and auditlogs.

The previous update was not enough, but as this test only runs for
master it passed unnoticed the PR.


lluiscampos:
See #1306

tranchitella (approved):

PR #1315, opened 2021-01-21
tests: update number of expected containers after migrating from nginx to traefik

kjaskiewiczz:


07.Server-installation/03.Production-installation/docs.md
```diff
@@ -831,9 +831,9 @@ git log --oneline master..HEAD
 
 <!-- Verification of Enterprise instance -->
```
lluiscampos, line 833: The Open Source test needs also an update, I assume. See lines 636-645.

lluiscampos (changes requested):

lluiscampos:
Also, don't you feel is about time to set these magic numbers as env variables at the top of the page(s)? :wink:

07.Server-installation/03.Production-installation/docs.md
kjaskiewiczz: right, fixed

kjaskiewiczz:
> Also, don't you feel is about time to set these magic numbers as env variables at the top of the page(s)?

hmm, those are two different numbers in six places in two different files; I'm not sure if changing it to two different variables at the top of two different pages will help a lot; I'm not saying it will not help at all, but
IMO the main issue is that we're forgetting to update the numbers when changing the number of containers in our setup;
I don't know if that's possible, but maybe figuring out in the tests the number of expected containers from yaml files could be a solution?

lluiscampos (approved):

PR #1316, opened 2021-01-21
[pipeline] Add option to skip tests

lluiscampos:
So that we can update the site if tests are failing, when necessary.


lluiscampos:
The proof that it works is that this PR is green :)

tranchitella (approved):

PR #1317, opened 2021-01-21
Revert mender-connect rename in changelogs

lluiscampos:
This text is auto generated and should not have been renamed together
with the rest.

This commit partially reverts 48519643.


lluiscampos:
> I know it's autogenerated, but does it make sense to make an exception this time? It's a bit confusing to use mender-shell, since we have not communicated this outside.

Okey, your reasoning makes sense. I'll close this PR, then resolve conflicts in hosted accordingly, and also port to 2.5.x, 2.4.x

PR #1324, opened 2021-01-26
Add changelogs and statistics for meta-mender/dunfell-v2021.01.

kacf:


kacf:
The failure is due to known issues regarding traefik, which are being worked on. This passed in older release branches without traefik. Merging.

PR #1376, opened 2021-03-18
[Debian family/Variables] Add Mender Configure add-on variables

lluiscampos:
And correct style of existing Mender Connect ones.

### DIFF
--- a/04.System-updates-Debian-family/99.Variables/docs.md
+++ b/04.System-updates-Debian-family/99.Variables/docs.md
@@ -117,14 +117,14 @@ The size of the boot partition.
 The version of the Mender client to include in the update.
 
 
-#### MENDER_ADDON_CONNECT_INSTALL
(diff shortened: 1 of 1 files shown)


04.System-updates-Debian-family/99.Variables/docs.md
```diff
@@ -117,21 +117,36 @@ The size of the boot partition.
 The version of the Mender client to include in the update.
 
 
-#### MENDER_ADDON_CONNECT_INSTALL
+#### `MENDER_ADDON_CONNECT_INSTALL`
```
kacf, line 120: I would use a separate commit so that this can be cherry-picked to 2.6.

kacf (approved):

04.System-updates-Debian-family/99.Variables/docs.md
lluiscampos: Good thinking, thanks

PR #1382, opened 2021-03-29
MEN-4514: Include session logging in remote terminal documentation

alfrunes:
changelog: title

### DIFF
--- a/09.Add-ons/01.Remote-Terminal/40.Security/docs.md
+++ b/09.Add-ons/01.Remote-Terminal/40.Security/docs.md
@@ -50,6 +50,8 @@ Remote Terminal does not open nor listen on any port: the device initiates the c
 !!! See [the Mender features page](https://mender.io/plans/features?target=_blank)
 !!! for an overview of all Mender plans and features.
 
-Audit logs hold a list of every start and stop of a session. You can access this information
-in the Mender UI.
-
+Every remote terminal invocation is saved in the audit logs. The logs
+include the session start and stop events as well as the terminal output. The
+audit logs are available in the Mender UI "AUDIT LOG" tab. There you can also
+inspect the terminal recording by playing back the terminal session in real time
+in the browser.
--- a/09.Add-ons/01.Remote-Terminal/docs.md
+++ b/09.Add-ons/01.Remote-Terminal/docs.md
@@ -76,6 +76,15 @@ is to read messages from the websocket, route them by session id
 to the correct shell, read the output going to the terminal,
 and pass it upstream over the websocket.
 
+For Mender Enterprise customers, Mender provides audit logging of all terminal
+sessions. The audit logs include terminal creation/termination events as well as
+the terminal output which users can later inspect from the
+[AUDIT LOG](https://hosted.mender.io/ui/#/auditlog) panel in the Mender UI.
+
+!!! Audit logs is only available in the Mender Enterprise plan.
+!!! See [the Mender features page](https://mender.io/plans/features?target=_blank)
+!!! for an overview of all Mender plans and features.
+
 ## Further reading
 
 * For a detailed list of the configuration options please refer to the


09.Add-ons/01.Remote-Terminal/docs.md
```diff
@@ -76,6 +76,13 @@ is to read messages from the websocket, route them by session id
 to the correct shell, read the output going to the terminal,
 and pass it upstream over the websocket.
 
+For customers with Mender Enterprise, audit logs are collected for all terminal
```
merlin-northern, line 79: suggested change: "For customers with Mender Enterprise" -> "For _Mender Enterprise customers_"

09.Add-ons/01.Remote-Terminal/docs.md
```diff
@@ -76,6 +76,13 @@ is to read messages from the websocket, route them by session id
 to the correct shell, read the output going to the terminal,
 and pass it upstream over the websocket.
 
+For customers with Mender Enterprise, audit logs are collected for all terminal 
+sessions. The audit logs include terminal creation/termination events as well as 
+the raw terminal output, for which you can replay in the browser from the Mender UI.
```
merlin-northern, line 81: 1. saying "raw terminal output" will raise questions, I would drop "raw". Also technically speaking it is not the auditlogs that contain the session recording as it is. 
2. suggested change: "terminal output, for which you can"  -> "terminal output which you can"

09.Add-ons/01.Remote-Terminal/40.Security/docs.md
```diff
@@ -50,6 +50,6 @@ Remote Terminal does not open nor listen on any port: the device initiates the c
 !!! See [the Mender features page](https://mender.io/plans/features?target=_blank)
 !!! for an overview of all Mender plans and features.
 
-Audit logs hold a list of every start and stop of a session. You can access this information
-in the Mender UI.
-
+Mender audits every remote terminal connection by logging the start/stop events
+as well as the terminal output. The audit logs are available in the Mender UI "AUDIT LOG" tab.
+There you can replay the recorded terminal session in real time.
```
merlin-northern, line 55: five comments here:

1. I would start with "Every Remote Terminal invocation is saved in the audit logs"
2. Perhaps we shall create a link referring  to the audit logs for further reading?
3. every time I say "replay" people ask: "ok, so those commands are going to be executed on the device again?" I have a strong feeling that we have to mention the fact that it is not the case.
4. As far as I know we also provide the "download" of the recording, maybe it is worthwhile to mention that?
5. I would mention that the same holds for mender-cli

09.Add-ons/01.Remote-Terminal/docs.md
```diff
@@ -76,6 +76,13 @@ is to read messages from the websocket, route them by session id
...
+
+!!! Audit logs are available to customers with
```
merlin-northern, line 83: minor:
```suggestion
!!! Audit logs are only available in the Mender Enterprise plan.
```
(reasoning: to be consistent with what we write in RBAC "`Role Based Access Control is only available in the Mender Enterprise plan.`")

09.Add-ons/01.Remote-Terminal/40.Security/docs.md
alfrunes: > 1.   I would start with "Every Remote Terminal invocation is saved in the audit logs"
> 2.    Perhaps we shall create a link referring to the audit logs for further reading?
> 3.   every time I say "replay" people ask: "ok, so those commands are going to be executed on the device again?" I have a strong feeling that we have to mention the fact that it is not the case.

:+1:

> 4.    As far as I know we also provide the "download" of the recording, maybe it is worthwhile to mention that?
> 5.    I would mention that the same holds for mender-cli

I don't know how to accurately document this. The record you download from the UI is incompatible with mender-cli and vice-versa, and I would argue that the `mender-cli` recording is a separate feature from what is provided by the backend. Thoughts?

09.Add-ons/01.Remote-Terminal/docs.md
alfrunes: > Also technically speaking it is not the auditlogs that contain the session recording as it is.

I think the technical details does not matter in this context. The session log is so tightly coupled to audit log in the UI that I think we should treat it like so in the documentation.

alfrunes:
@merlin-northern I tried to address most of your comments.
I also did some re-phrasing -- changing "Audit logs" to singular "Audit log".

09.Add-ons/01.Remote-Terminal/40.Security/docs.md
merlin-northern: mender-cli: I meant to mention that the sessions can also be recorded in the CLI :) "The same holds" referred to the session logging in general. sorry for the confusion.

Download from the UI: Manuel worked so hard on the downloader, so I would at least mention that in the UI you can download the recording. And good point to mention that the recordings are not compatible, they are the one from the UI is a html5 recording and the one from the CLI is terminal output zipped with custom metadata header.

09.Add-ons/01.Remote-Terminal/40.Security/docs.md
```diff
@@ -50,6 +50,8 @@ Remote Terminal does not open nor listen on any port: the device initiates the c
 !!! See [the Mender features page](https://mender.io/plans/features?target=_blank)
 !!! for an overview of all Mender plans and features.
 
-Audit logs hold a list of every start and stop of a session. You can access this information
-in the Mender UI.
-
+Every remote terminal invocation is saved in the audit logs. The logs
+include the session- start and stop events as well as the terminal output. The
```
merlin-northern, line 54: is '-' on purpose in "session- "? if so then why there is space after it? (feels a bit confusing as it is now)

merlin-northern (approved):
lgtm, I leave the mention of CLI recording support at your discretion as well as the UI recording download.

09.Add-ons/01.Remote-Terminal/40.Security/docs.md
alfrunes: No, this hyphen does not belong here.

PR #1407, opened 2021-04-17
Docs exec testing: pass as soon as containers run.

merlin-northern:
ChangeLog:none

### DIFF
--- a/07.Server-installation/03.Production-installation/docs.md
+++ b/07.Server-installation/03.Production-installation/docs.md
@@ -5,6 +5,12 @@ taxonomy:
     label: tutorial
 ---
 
+<!-- The variables used in the testing of production installation instructions -->
+<!--AUTOMATION: execute=export EXPECTED_COUNT_OPENSOURCE=14 -->
+<!--AUTOMATION: execute=export EXPECTED_COUNT_ENTERPRISE=16 -->
+<!--AUTOMATION: execute=export SLEEP_INTERVAL=4 -->
+<!--AUTOMATION: execute=export MAX_INTERATIONS=11 -->
+
 <!-- AUTOMATION: execute=if [ "$TEST_OPEN_SOURCE" != 1 ] && [ "$TEST_ENTERPRISE" != 1 ]; then echo "Either TEST_OPEN_SOURCE xor TEST_ENTERPRISE must be set to 1!"; exit 1; fi -->
 <!-- AUTOMATION: execute=if [ "$TEST_OPEN_SOURCE" = 1 ] && [ "$TEST_ENTERPRISE" = 1 ]; then echo "TEST_OPEN_SOURCE and TEST_ENTERPRISE cannot both be 1!"; exit 1; fi -->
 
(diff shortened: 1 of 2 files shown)


07.Server-installation/03.Production-installation/docs.md
```diff
@@ -613,10 +613,14 @@ installing an Open Source server, please proceed to
 
 <!-- Verification of Open Source instance -->
 
-<!--AUTOMATION: test=for ((n=0;n<11;n++)); do sleep 4 && test "$(docker ps | grep menderproduction | grep -c -i 'up')" = 14  || ( echo "some containers are not 'Up'" && docker ps && ./run images && ./run logs && exit 1 ); done -->
+<!--AUTOMATION: execute=export EXPECTED_COUNT=14 -->
```
lluiscampos, line 616: Once we are reworking the test... please let's move these magic numbers to explicit variables at the beginning of the file. Something like `EXPECTED_COUNT_ENTERPRISE` and `EXPECTED_COUNT_OPENSOURCE` will be much better. I suggest we do the same in `01.Upgrading-from-OS-to-Enterprise` file, even though there we would only use one of them (enterprise)

07.Server-installation/03.Production-installation/docs.md
```diff
@@ -613,10 +613,14 @@ installing an Open Source server, please proceed to
...
+<!--AUTOMATION: execute=export SLEEP_INTERVAL=4 -->
+<!--AUTOMATION: execute=export MAX_INTERATIONS=11 -->
+<!--AUTOMATION: execute=function CONTAINERS_COUNT_TEST() { local n; [ "${EXPECTED_COUNT}" == "" ] && return 1; for ((n=0;n<${MAX_INTERATIONS:-11};n++)); do count=$(docker container ls -f "status=running" --format 'table {{.Status}}\t{{.Names}}' | grep ^Up | grep -c menderproduction); [ ${count} -eq ${EXPECTED_COUNT} ] && break; sleep ${SLEEP_INTERVAL:-4}; done; [ ${count} -eq ${EXPECTED_COUNT} ] || { echo "some containers are not 'Up'; ${count}/${EXPECTED_COUNT} running." && docker ps && ./run images && ./run logs && exit 1; }; } -->
```
lluiscampos, line 619: I don't think the variables here should have defaults as we should expect all to have been defined
```suggestion
<!--AUTOMATION: execute=function CONTAINERS_COUNT_TEST() { local n; [ "${EXPECTED_COUNT}" == "" ] && return 1; for ((n=0;n<${MAX_INTERATIONS};n++)); do count=$(docker container ls -f "status=running" --format 'table {{.Status}}\t{{.Names}}' | grep ^Up | grep -c menderproduction); [ ${count} -eq ${EXPECTED_COUNT} ] && break; sleep ${SLEEP_INTERVAL}; done; [ ${count} -eq ${EXPECTED_COUNT} ] || { echo "some containers are not 'Up'; ${count}/${EXPECTED_COUNT} running." && docker ps && ./run images && ./run logs && exit 1; }; } -->
```

I would also suggest (weak suggestion) to remove the `[ "${EXPECTED_COUNT}" == "" ] && return 1` check, what is the purpose of that?

lluiscampos (changes requested):
Great initiative, thank you!

One request: to please define the magic numbers at the beginning of the files for easier maintenance in the future. What do you think?

07.Server-installation/03.Production-installation/docs.md
merlin-northern: fixed!

> I would also suggest (weak suggestion) to remove the [ "${EXPECTED_COUNT}" == "" ] && return 1 check, what is the purpose of that?

to fail the test if the count is not set? I can remove, no worries, it made sense to me to have it.

07.Server-installation/03.Production-installation/docs.md
merlin-northern: sure.
moved and renamed, also renamed the function name to indicate the ENT/OS test scenario.

lluiscampos (approved):

PR #1436, opened 2021-04-21
2.6.1 release changelogs n stats

oleorhagen:
* 2.6.1 release logs and stats
* mender-binary-delta 1.1.2 logs

### DIFF
--- a/202.Release-information/01.Release-notes-changelog/docs.md
+++ b/202.Release-information/01.Release-notes-changelog/docs.md
@@ -1498,6 +1498,21 @@ New changes in workflows-enterprise since 1.3.0:
   * Bumps alpine from 3.13.1 to 3.13.2.
 
 
+## mender-binary-delta 1.1.2
+
+_Released 16.04.2021_
+
+#### mender-binary-delta (1.1.2)
+
+New changes in mender-binary-delta since 1.1.0:
+
+* Probe U-Boot env before use to support libubootenv fw tools
+  ([MEN-4246](https://tracker.mender.io/browse/MEN-4246))
+* Remove harmless warning message about unhandled states.
+* Fix failed rollback status when bootloader is the one to roll back.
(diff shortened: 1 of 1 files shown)


202.Release-information/01.Release-notes-changelog/docs.md
```diff
@@ -1406,6 +1406,156 @@ New changes in workflows-enterprise since 1.3.0:
...
+New changes in mender-binary-delta since 1.1.0:
+
+* Probe U-Boot env before use to support libubootenv fw tools
+  ([MEN-4246](https://tracker.mender.io/browse/MEN-4246))
+* Remove harmless warning message about unhandled states.
+* Fix failed rollback status when bootloader is the one to roll back.
+* Detect mismatches between `mender_boot_part` and `RootfsPart(A|B)` variables.
+
+
+## Mender 2.6.1
```
lluiscampos, line 1424: Shouldn't this be between 2.4.3 and 2.7.0? Alternatively just on top (as we usually do, prepend as we go), but inserting it here with no good criteria does not look right to me.

oleorhagen: Hmm, sure, but should we then place `2.7.0` at the top maybe?
Then `2.6.1`, and lastly `2.4.3`, to make it pretty

oleorhagen: @lluiscampos

lluiscampos: @oleorhagen I wouldn't change the existing ones, so I suggest either put this one between 2.7.0 and 2.4.3 or just in the first one (which is the regular process, we just prepend them as we update the docs).

lluiscampos (approved):
Either option is fine, this file won't be here for very much longer

PR #1437, opened 2021-04-21
MEN-4513: Move changelog pages out of mender-docs into an independent repo

lluiscampos:
The new repo, mender-docs-changelog, will contain individual pages for
each of the "components" that we release.

They will be added here at build time from mender-docs-site repository.


lluiscampos:
Depends on https://github.com/oleorhagen/remark-validate-links/pull/3

tranchitella (approved):

package.json
```diff
@@ -5,10 +5,9 @@
   "main": "example.js",
   "dependencies": {
     "remark-cli": "^8.0.0",
-    "remark-validate-links": "https://github.com/oleorhagen/remark-validate-links#grav-slug",
+    "remark-validate-links": "github:oleorhagen/remark-validate-links#5b203f58dc1e3b44b5c5376a3543969a0eda7f6b",
```
lluiscampos, line 8: @oleorhagen why use the commit instead of the branch?

oleorhagen: Because it refused to set it to the right sha for some reason.
I just went along with it, since it's locked to a sha even if you use the branch ref.

lluiscampos:
@oleorhagen Ouch, I force-pushed and missed your commits. Can you please fix again? (do we need to fix? Seems like now the branch ref is working). Sorry for the inconvenience.

oleorhagen:
> @oleorhagen Ouch, I force-pushed and missed your commits. Can you please fix again? (do we need to fix? Seems like now the branch ref is working). Sorry for the inconvenience.

o_O I have no idea why it is not respecting the `package-lock.json` sha.

kacf (approved):
:partying_face: Yey, finally!

PR #1449, opened 2021-05-03
MEN-4256: Mender-convert no longer installs the client by default

oleorhagen:
Changelog: None

### DIFF
--- a/04.System-updates-Debian-family/99.Variables/docs.md
+++ b/04.System-updates-Debian-family/99.Variables/docs.md
@@ -109,6 +109,12 @@ Set the default index for the boot partition.
 The size of the boot partition.
 
 
+#### `MENDER_CLIENT_INSTALL`
+
+> Value: y(default)/n
+
+Install the Mender client.
+
 #### `MENDER_CLIENT_VERSION`
 
 <!--AUTOVERSION: "/%/"/ignore-->
--- a/09.Downloads/docs.md
+++ b/09.Downloads/docs.md
@@ -11,7 +11,7 @@ process:
 ## Disk images
 
 These disk images (`*.img` or `*.sdimg`) are based on images provided by board
-manufacturers and already have Mender fully integrated. They are used to
+manufacturers and are ready to install the Mender client. They are used to
 provision the device storage for devices without Mender running already.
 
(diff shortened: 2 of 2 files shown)


lluiscampos (approved):

lluiscampos (changes requested):
Missing: to add the new variable in https://docs.mender.io/system-updates-debian-family/variables#mender_client_version

oleorhagen:
> Missing: to add the new variable in https://docs.mender.io/system-updates-debian-family/variables#mender_client_version

Oh yes, good catch! I didn't know we documented these

04.System-updates-Debian-family/99.Variables/docs.md
```diff
@@ -109,6 +109,12 @@ Set the default index for the boot partition.
 The size of the boot partition.
 
 
+#### `MENDER_CLIENT_INSTALL`
+
+> Value: y (default)
```
lluiscampos, line 114:
```suggestion
> Value: y(default)/n
```

See the syntax in, for example, `MENDER_ADDON_CONNECT_INSTALL`

lluiscampos (approved):
See comment below, though.

04.System-updates-Debian-family/99.Variables/docs.md
oleorhagen: This is a bit all over the place though:

```
#### `MENDER_DATA_PART_GROWFS`

> Values: y(default)/n

Enable/Disable automatically growing the filesystem to fill the physical storage device.
```

```
#### `MENDER_ADDON_CONFIGURE_INSTALL`

> Value: y/n(default)

Install the Mender Configure add-on.

```
```
#### `MENDER_BOOT_PART_SIZE_MB`

> Value: 40 (default)

The size of the boot partition.

``` (with air, like most of them have)

To be honest, it's the `ADDON` values standing out as the outliers

lluiscampos: They all look good to me.

Criteria seems to be: we use `/` to describe legal values with no air, adding `(default)` on the default one, and we have air on the "any-string" kind of config.

lluiscampos (approved):

PR #1452, opened 2021-05-11
Add dependabot for npm dependencies

lluiscampos:


lluiscampos:
Same here... not important but nice to avoid the security alerts

PR #1459, opened 2021-05-20
Add an example to illustrate that deployed containers are also started.

kacf:
Just a small example I wanted to add after doing Hosted onboarding testing.


estenberg (approved):
Great clarification :+1:

PR #1461, opened 2021-06-02
MEN-4107: introduce new chapter: Device-side API

tranchitella:


.passive-blacklist
```diff
@@ -1,6 +1,7 @@
 06.Artifact-creation
 07.Server-installation
 09.Downloads
-200.API
-201.Troubleshoot
-202.Release-information
+200.Server-side-API
+201.Device-side-API
+202.Troubleshoot
+203.Release-information
```
lluiscampos, line 7: Once we are renaming, could we move these to `300` and `301`? Then we have room to add more "APIs" in the future.

200.Server-side-API/chapter.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: API
+title: Server-side API
 redirect: /../api/development/
```
lluiscampos, line 3: To the best of my knowledge, thi should work, but have you built it locally and verified?

lluiscampos (approved):

200.Server-side-API/chapter.md
tranchitella: Yes, I tested it

.passive-blacklist
tranchitella: Done!

PR #1463, opened 2021-06-03
MEN-4715: Move to using /etc/apt/source.list.d/mender.list

oleorhagen:
This moves the documentation to use a custom 'mender.list' source script, as
opposed to directly adding the source to '/etc/apt/sources.list' globally, which
was done previously, and is consider poor form.

Also, add a simple command for helping the user purge any existing mender
configuration from the global source list, if the user has previously done a
mender install with the old documentation.

Changelog: None


lluiscampos (approved):
Looks good to me, but I'd like to have fresh eyes on it so I pinged @alfrunes too.

alfrunes (approved):

PR #1473, opened 2021-06-14
MEN-4561: Customize the update process with Synchronized updates

tranchitella:
Commits will be squashed before merge.


02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,45 @@ machine. Depending on the return code from the script the Mender client will:
...
+machine and can optionally have the priority defined as a positive integer value. The default
+action for each step is **continue**. To pause the update process, set **pause**.
+
+Once *paused*, the update process can either be *resumed* or *canceled* and the whole deployment marked
+as failed updating the update control map of the Mender client. You can achieve it either using the
+[Server-side API](../../200.Server-side-API), replacing the deployment's update control map,
+or the [Device-side APIs](../../201.Device-side-API), updating the Mender client's update control map.
+
+You can also set a default update control map in the Mender client running in the device, which will
+be active by default unless overridden by a deployment's update control map with a higher priority.
```
oleorhagen, line 98: Suggestion:
unless a deployment's update control map overrides it with a map with a higher priority.

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,45 @@ machine. Depending on the return code from the script the Mender client will:
...
+
+Once *paused*, the update process can either be *resumed* or *canceled* and the whole deployment marked
+as failed updating the update control map of the Mender client. You can achieve it either using the
+[Server-side API](../../200.Server-side-API), replacing the deployment's update control map,
+or the [Device-side APIs](../../201.Device-side-API), updating the Mender client's update control map.
+
+You can also set a default update control map in the Mender client running in the device, which will
+be active by default unless overridden by a deployment's update control map with a higher priority. 
+
+For more information about the different update control map's settings, please refer to the
```
oleorhagen, line 100: I think we could get away with:
the different update control map settings

oleorhagen (approved):
Nice, left some comments for thought, feel free to ignore if you disagree, most of them are suggestions/nits. Only a couple of points which needs attention.

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,45 @@ machine. Depending on the return code from the script the Mender client will:
...
+### Further reading
 
 To learn more, have a look at the [tutorial on how to implement State
 scripts](../../06.Artifact-creation/04.State-scripts/docs.md).
+
+## Synchronized updates
+
+When creating a software deployment, it is possible to customize the update process by specifying
+an update control map, which can control the update process considering **external conditions**,
+eventually including input from the user.
```
kacf, line 75: I think we might need a slightly softer introduction. I think that the sentence is difficult to understand if you don't know what an Update Control map is. How about starting with "*When creating a software deployment, it is possible to control the progress of the update. It's possible to tell the Mender client to wait at certain points in the deployment, and then either continue or roll back from that point at some later time. This is done by specifying an update control map, which can...*"

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,45 @@ machine. Depending on the return code from the script the Mender client will:
...
-## Further reading
+### Further reading
 
 To learn more, have a look at the [tutorial on how to implement State
 scripts](../../06.Artifact-creation/04.State-scripts/docs.md).
+
+## Synchronized updates
+
+When creating a software deployment, it is possible to customize the update process by specifying
+an update control map, which can control the update process considering **external conditions**,
```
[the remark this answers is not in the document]
tranchitella: It sounds good to me, to be honest.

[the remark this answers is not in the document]
oleorhagen: Then go with it :)

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,45 @@ machine. Depending on the return code from the script the Mender client will:
...
+![Update process options](update-process-options.jpg)
+
+Common use cases are:
+
+* User presses *Start update* on the device to process the software update.
+* Synchronized update, all devices are verified by QA and can commit the software, updating
+  simultaneously.
+
+Update control maps can specify an action for any transition between the Mender client's state
+machine and can optionally have the priority defined as a positive integer value. The default
```
kacf: I would omit the priority altogether. This is only important if you mix the backend and D-Bus approaches, at which point you're already very deep into this feature. So I think it's enough that this property is listed in the D-Bus API reference.

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,45 @@ machine. Depending on the return code from the script the Mender client will:
...
 logs are sent to the Mender server. The output from state scripts are included
 in these logs. This is useful when troubleshooting intermittent update failures.
 
-## Further reading
+### Further reading
 
 To learn more, have a look at the [tutorial on how to implement State
 scripts](../../06.Artifact-creation/04.State-scripts/docs.md).
+
+## Synchronized updates
```
kacf, line 71: I think the two approaches for using update control, D-Bus and backend-based, have very different audiences, and we should not treat them in the same section. I would split this up, perhaps handle the server based variant first (since it is easier to grasp), and the D-Bus information under a new header. The fact that they are based on the same underlying mechanism is more of an implementation detail, and not normally how you would approach the subject, I think.

You may have to duplicate some things then, but I think that's alright.

kacf (commented):
Maybe @estenberg have some comments too, especially regarding my last comment.

02.Overview/10.Customize-the-update-process/docs.md
tranchitella: @kacf two sections in the same page, right?

kacf: Long term I think it would be better to have a separate page, but I can't find a good place for it in our current structure. So yes, for now I think we can use the same page.

tranchitella: Done: https://github.com/mendersoftware/mender-docs/pull/1473/commits/d8692b195e61e9e5720aa31717af3ffa96fdb6ce

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,58 @@ machine. Depending on the return code from the script the Mender client will:
...
+* User presses *Start update* on the device to process the software update.
+* Synchronized updates, all devices are verified by QA and can commit the software, updating
+  simultaneously.
+
+Update control maps can specify an action for some transitions between the Mender client's
+state machine. The default action for each step is **continue**. To pause the update process,
+set **pause**.
+
+Once *paused*, the update process can either be *resumed* or *canceled* and the whole deployment marked
+as failed by updating the update control map of the Mender client.
```
lluiscampos, line 96: Should `Update Control Map` be capitalized or not? There is one iteration above with `Update Control Map` and others here and below with `update control map`.

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,58 @@ machine. Depending on the return code from the script the Mender client will:
...
+contact the Mender Server to refresh the update control map and perform the required actions.
+
+### Synchronized updates using the Device-side API
+
+You can also manage the update progress using the [Device-side APIs](../../201.Device-side-API),
+updating the Mender client's update control map using the *SetUpdateControlMap* method of the
+*Mender Update Management API* .
+
+You can also set a default update control map in the Mender client running on the device, which will
+be active by default unless a deployment's update control map overrides it.
```
lluiscampos, line 111: This is not really trivial, as to do that you need an application in the device constantly refreshing the map into the Mender client.

While it is a valid use case, the phrasing here seems to be that you can set a default map like in a config file or somewhere *static* and let it be. It is not the case so I would either remove the paragraph or be more explicit.

02.Overview/10.Customize-the-update-process/docs.md
tranchitella: No idea, @oleorhagen asked to capitalize the first occurrence. What's our policy?

02.Overview/10.Customize-the-update-process/docs.md
tranchitella: @lluiscampos what do you mean with refreshing the map? should it be done every N seconds/minutes, or after each Mender client restart, or...?

lluiscampos: Once you set a map via D-Bus you get an expire time (in seconds). The application needs to refresh it (or send a new one) before it expires, or the client will disregard this map whenever the next update arrives (or within an update in progress, for all that matters).

tranchitella: Let's give it a try: https://github.com/mendersoftware/mender-docs/pull/1473/commits/3471ed4318bd302b1c7eae5ff205c2a65e474cb7

02.Overview/10.Customize-the-update-process/docs.md
tranchitella: Let's capitalize it everywhere: https://github.com/mendersoftware/mender-docs/pull/1473/commits/3471ed4318bd302b1c7eae5ff205c2a65e474cb7

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,61 @@ machine. Depending on the return code from the script the Mender client will:
...
 
-## Further reading
+### Further reading
 
 To learn more, have a look at the [tutorial on how to implement State
 scripts](../../06.Artifact-creation/04.State-scripts/docs.md).
+
+## Synchronized updates
+
+When creating a software deployment, it is possible to control the update progress.
```
estenberg, line 73: "..update progress from an external service by calling Mender APIs."

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,61 @@ machine. Depending on the return code from the script the Mender client will:
...
+
+![Update process options](update-process-options.jpg)
+
+Common use cases are:
+
+* User presses *Start update* on the device to process the software update.
+* Synchronized updates, all devices are verified by QA and can commit the software, updating
+  simultaneously.
+
+Update control maps can specify an action for some transitions between the Mender client's
```
estenberg, line 91: I think it's too detailed for overivew. Would move to (to-be-written "Use the Device-side API" tutorial).

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,61 @@ machine. Depending on the return code from the script the Mender client will:
...
+state machine. The default action for each step is **continue**. To pause the update process,
+set **pause**.
+
+Once *paused*, the update process can either be *resumed* or *canceled* and the whole deployment marked
```
estenberg, line 95: I think it's too detailed for overivew. Would move to (to-be-written "Use the Device-side API" tutorial).

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,61 @@ machine. Depending on the return code from the script the Mender client will:
...
+replacing the deployment's Update Control Map. During pauses, the Mender client will periodically
+contact the Mender Server to refresh the Update Control Map and perform the required actions.
+
+### Synchronized updates using the Device-side API
+
+You can also manage the update progress using the [Device-side APIs](../../201.Device-side-API),
+updating the Mender client's Update Control Map using the *SetUpdateControlMap* method of the
+*Mender Update Management API* .
+
+Applications running in the device can also set a default Update Control Map in the Mender
```
estenberg, line 110: I think it's too detailed for overivew. Would move to (to-be-written "Use the Device-side API" tutorial).

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,61 @@ machine. Depending on the return code from the script the Mender client will:
...
+  simultaneously.
+
+Update control maps can specify an action for some transitions between the Mender client's
+state machine. The default action for each step is **continue**. To pause the update process,
+set **pause**.
+
+Once *paused*, the update process can either be *resumed* or *canceled* and the whole deployment marked
+as failed by updating the Update Control Map of the Mender client.
+
+### Synchronized updates using the Server-side API
```
estenberg, line 98: Would replace everything from here to the end with a simple next step (API reference is quite "rough"):

## Further reading
"For more information on getting started with synchronized updates attributes, see [Use the Device-side API][404 LINK for now], or take a look at the [Device-side APIs](../../201.Device-side-API) and [Server-side API](../../200.Server-side-API) reference directly."

estenberg (changes requested):
A few minor tweaks to simplify. I suppose the "Use the Device-side API" is coming later.

02.Overview/10.Customize-the-update-process/docs.md
tranchitella: Something like this, maybe? https://github.com/mendersoftware/mender-docs/pull/1473/commits/1bdd4470137bf5096a5c25eb689acbe4d69bf4fa

02.Overview/10.Customize-the-update-process/01.How-to-use-the-update-control-map/docs.md
```diff
@@ -0,0 +1,34 @@
...
+contact the Mender Server to refresh the Update Control Map and perform the required actions.
+
+## Synchronized updates using the Device-side API
+
+You can also manage the update progress using the [Device-side APIs](../../../201.Device-side-API),
+updating the Mender client's Update Control Map using the *SetUpdateControlMap* method of the
+*Mender Update Management API* .
+
+Applications running in the device can also set a default Update Control Map in the Mender
+client, which will be active by default unless a deployment's Update Control Map overrides it.
```
lluiscampos, line 28: Good try, but I think we are not completely there yet.

A deployment Update Control Map cannot override a D-Bus set map. All maps need a UUID and are therefore unique. (For the deployment, the UUID of the map would be the same as the deployment ID). What can happen is that the deployment map has a higher priority, then both are taking into consideration starting by the high priority one and then the other, but I think that the effect is that it will get "merged" in the attributes that are not present in both.

@kacf what is the right thing to say here?

tranchitella: Trying my best: https://github.com/mendersoftware/mender-docs/pull/1473/commits/9858c653fc6281e85688c175c259137083a3f020

estenberg (changes requested):
@tranchitella I think you achieved the task at hand if you remove "10.Customize-the-update-process/01.How-to-use-the-update-control-map/docs.md" (and the link to it).

This tutorial does not fit within Overview, which is supposed to be only what & why (use cases), not how (tutorials).

Also note that https://tracker.mender.io/browse/MEN-4630 will cover what you started here, this is why I suggest you simply drop it for now. :)

tranchitella:
> @tranchitella I think you achieved the task at hand if you remove "10.Customize-the-update-process/01.How-to-use-the-update-control-map/docs.md" (and the link to it).
> 
> This tutorial does not fit within Overview, which is supposed to be only what & why (use cases), not how (tutorials).
> 
> Also note that https://tracker.mender.io/browse/MEN-4630 will cover what you started here, this is why I suggest you simply drop it for now. :)

Done: https://github.com/mendersoftware/mender-docs/pull/1473/commits/6f355cfb58c67e588e15370504d083447893e3e0
I added the text I wrote to the ticket, in case it can be useful in the future.

02.Overview/10.Customize-the-update-process/01.How-to-use-the-update-control-map/docs.md
lluiscampos: Looks good to me :+1: Thanks

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,39 @@ machine. Depending on the return code from the script the Mender client will:
...
 To learn more, have a look at the [tutorial on how to implement State
 scripts](../../06.Artifact-creation/04.State-scripts/docs.md).
+
+## Synchronized updates
+
+When creating a software deployment, it is possible to control the update progress from an
+external service by calling Mender APIs. It's possible to tell the Mender client to wait at
+specific points in the deployment, and then either continue or roll back from that point at
+some later time. To achieve that, you specify an Update Control Map, which can control the
+update process considering **external conditions**, eventually including input from the user.
```
kacf, line 77: With the focus on high level information, I think we should just leave out API altogether and focus on usability from the UI, in which case this paragraph can be significantly simplified. This is my suggestion:

> When creating a software deployment, it is possible to control the update from the Mender UI by telling the Mender client to wait at specific points in the deployment. The deployment can then either be continued or rolled back from this point at some later time. See the screenshot below for an example of what is possible.

@estenberg, do you agree?

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,39 @@ machine. Depending on the return code from the script the Mender client will:
...
+
+When entering the Install, Reboot, or Commit states during an update, you can require the Mender
+client to pause and wait for an external event to further proceed with the processing of the
+update.
```
kacf, line 81: With the picture below, I think this paragraph is redundant and we can remove it.

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,39 @@ machine. Depending on the return code from the script the Mender client will:
...
+
+![Update process options](update-process-options.jpg)
+
+Common use cases are:
+
+* User presses *Start update* on the device to process the software update.
```
kacf, line 87: This is not a valid use case anymore after the device-side of things was removed from this section. However, I think we should have at least two use cases, so what about:

> Synchronized reboot: Devices can download whenever they have the opportunity, but all devices need to reboot during the maintenance window, at the same time.

kacf (changes requested):

02.Overview/10.Customize-the-update-process/01.How-to-use-the-update-control-map/docs.md
tranchitella: This sentence was removed, as requested by @estenberg

02.Overview/10.Customize-the-update-process/docs.md
tranchitella: https://github.com/mendersoftware/mender-docs/pull/1473/commits/3c7290ca9c253d8d88023efa662d5432526c9a22

estenberg: @tranchitella not sure if you wanted all these comments when you started asking for reviews :)

But now it omits the device-side use-cases, which leaves out a large part of the value (only talking about the UI).
I think we should have the device-side use cases there without getting into the API details. Will add some comments.

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,37 @@ machine. Depending on the return code from the script the Mender client will:
...
+the Mender UI by telling the Mender client to wait at specific points in the
+deployment. The deployment can then either continue or roll back from this
+point at some later time.
+
+See the screenshot below for an example of what is possible.
+
+![Update process options](update-process-options.jpg)
+
+Common use cases are:
+
```
estenberg, line 83: * User presses *Start update* on the device to process the software update as an update is ready.
* Device is in use, by default no updates are started. Once ready, the user can press *Enable updates* on the device.

02.Overview/10.Customize-the-update-process/docs.md
```diff
@@ -49,13 +57,37 @@ machine. Depending on the return code from the script the Mender client will:
...
-## Further reading
+### Further reading
 
 To learn more, have a look at the [tutorial on how to implement State
 scripts](../../06.Artifact-creation/04.State-scripts/docs.md).
+
+## Synchronized updates
+
+When creating a software deployment, it is possible to control the update from
+the Mender UI by telling the Mender client to wait at specific points in the
```
estenberg, line 74: "..the Mender UI or an application running locally on the device.."

estenberg (changes requested):

tranchitella:
@estenberg @kacf this is now a loop: you asked to add the device-side API, then remove it, now asking to add it again. :)
What shall we do? The content is there, more or less, can you please you two find an agreement? Based on that, I can finalize the PR.

estenberg:
@tranchitella I don't think there is a loop/conflict and that @kacf and me are saying the same thing:

* Do not cover (server or device) API details / howtos under Overview/ (so you removed that part and posted it in the ticket that will deal with it: https://tracker.mender.io/browse/MEN-4630)
* Do cover use cases (both server and device), without going into how they are implemented.

Basically Overview/ is about what & why, while the more detailed sections (e.g. Client installation) is more about how. Overview/ is used so users can see what can be done, and motivate them to move to the tutorials later.

Did it make sense? @kacf ?

tranchitella:
@estenberg @kacf @lluiscampos @oleorhagen I addressed the last two comments from Eystein, and squashed the commits. Can I kindly ask you a final round of comments?

kacf:
Yes, it looks good now! Sorry about all the back and forth.

One question for @estenberg, though it is more for the future: Should we have a dedicated page for how to use Update Control from the Mender UI? If we do, I don't think we have a ticket for this.

It's also possible it's self explanatory enough that we don't need it, but I suspect users will feel more reassured if they know exactly what the buttons do.

kacf (approved):

tranchitella:
@estenberg your call, now. :)

PR #1475, opened 2021-06-18
MEN-4785: Update Device API compatibility table for v2.

kacf:


kacf:
This is pending a decision on which versions we will bump due to the introduction of Device API v2, either client version, Mender Product version, or both.

lluiscampos (approved):
I support bumping the client. We have not bumped in the past the Mender server when some microservices have gone from v1 to v2, so with that reasoning we can continue in the 2.x series there.

Also, the new table is much more readable, regardless of the bumping.

tranchitella (approved):
I have the same observation Lluis had: I personally see no reason to bump the Server version.

kacf:
As it turns out, @estenberg immediately agreed to bump the server. I guess it makes sense, in that bumping the device API is more severe than bumping the management API. So unless there are any direct objections to that, I'll change it in this PR as well.

lluiscampos:
Okey with me - it wasn't an strong opinion in the first place.

PR #1490, opened 2021-07-06
Slight adjustment to the return value of Update Control Maps API.

kacf:
You're not meant to update the map before this time, you're meant to
update after it has elapsed (this is why it is half the expiry time).


tranchitella (approved):

kacf:
Merged this into 3.0.x as well (there were no other changes in master since the branching).

PR #1491, opened 2021-07-06
MEN-4630: restore "Update confirmation by end user" chapter

tranchitella:
Changelog: none


kacf (approved):
Remember to cherry-pick to 3.0.x also!

PR #1493, opened 2021-07-09
DBus Python example: Avoid need to visit different site for installation.

kacf:


estenberg (approved):
:+1:

tranchitella (approved):

kacf:
Cherry-picked to 3.0.x as well.

lluiscampos:
Fabio's original work was among these lines and it was me who suggested to point to the other docs instead of copy-pasting the rules. Sorry!

PR #1496, opened 2021-07-13
MEN-4881 File Transfer limits docs update

merlin-northern:
ChangeLog:title

### DIFF
--- a/09.Add-ons/03.File-Transfer/docs.md
+++ b/09.Add-ons/03.File-Transfer/docs.md
@@ -15,6 +15,15 @@ and provides data to reply to the HTTP calls. The implementation does not relay
 on any third-party software installed on a device, and in that sense has no relation
 to FTP or SFTP. Although feature-wise it provides similar functionalities.
 
+## Controlling the uploads and downloads
+
+You can use the File Transfer feature to upload and download anything to and from a device.
+In itself it gives a great power, but may require a certain degree of control.
+[The Limits](../90.Reference/docs.md#limits-configuration) configuration section allows you
+to restrict many aspects of the file transfer process. We designed it to give you full control over what
+you can do, while at the same time providing a way to introduce restrictions for environments where limiting
+is crucial.
+
 ## Limitations
 
 ### Multiple parallel file transfer sessions to the same device
--- a/09.Add-ons/90.Reference/docs.md
+++ b/09.Add-ons/90.Reference/docs.md
@@ -28,6 +28,27 @@ mender-connect along with the default values. The default configuration path is
   "Servers": null,
   "ServerURL": "",
   "SkipVerify": false,
+  "Limits": {
+    "Enabled": true,
+    "FileTransfer": {
+      "Chroot": "/var/lib/mender/filetransfer",
+      "OwnerGet": ["pi","root"],
+      "GroupGet": ["games","users"],
+      "OwnerPut": "root",
+      "GroupPut": "pi",
+      "MaxFileSize": 4,
+      "FollowSymLinks": true,
+      "AllowOverwrite": true,
+      "RegularFilesOnly": true,
+      "PreserveOwner": true,
+      "PreserveGroup": true,
+      "PreserveMode": true,
+      "Counters": {
+       "MaxBytesTxPerHour": 1048576,
+       "MaxBytesRxPerHour": 1048576
+      }
+    }
+  },
   "FileTransfer": {
     "Disable": false
   },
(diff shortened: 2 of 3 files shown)


09.Add-ons/03.File-Transfer/docs.md
```diff
@@ -15,6 +15,15 @@ and provides data to reply to the HTTP calls. The implementation does not relay
 on any third-party software installed on a device, and in that sense has no relation
 to FTP or SFTP. Although feature-wise it provides similar functionalities.
 
+## Controlling the uploads and downloads
+
+You can use the File Transfer feature to upload and download anything to and from a device.
+In itself it gives a great power, bu may require a certain degree of control.
```
kacf, line 21: `s/bu/but/`

09.Add-ons/03.File-Transfer/docs.md
```diff
@@ -15,6 +15,15 @@ and provides data to reply to the HTTP calls. The implementation does not relay
...
+[Limits](../90.Reference/docs.md#limits-configuration) configuration section allows you
```
kacf, line 22: I would say "**The** Limits configuration section...".

09.Add-ons/03.File-Transfer/docs.md
```diff
@@ -15,6 +15,15 @@ and provides data to reply to the HTTP calls. The implementation does not relay
...
+to restrict many aspects of the file transfer process. We designed it to give you full control over what
```
kacf, line 23: Is it better to say "It was designed"..? Not sure, treat it as a suggestion only.

09.Add-ons/90.Reference/docs.md
```diff
@@ -70,16 +91,45 @@ mender-connect along with the default values. The default configuration path is
 ! `Servers` and `ServerURL` are deprecated and unused since `mender-connect`
 ! version `1.0.0` - the values are automatically configured by the `mender-client`.
 
-#### File transfer configuration
+#### Limits configuration
+There are certain features that you would want to keep under finer
+control than just enable/disable. File Transfer is one example; imagine
+you would like to restrict the transfers to a certain user or a group,
+or limit the average number of bytes that a device can transfer in an hour.
+Limits section can be helpful here.
```
kacf, line 99: Again, I would say "The Limits section".

09.Add-ons/90.Reference/docs.md
```diff
@@ -70,16 +91,45 @@ mender-connect along with the default values. The default configuration path is
 ! `Servers` and `ServerURL` are deprecated and unused since `mender-connect`
 ! version `1.0.0` - the values are automatically configured by the `mender-client`.
 
-#### File transfer configuration
+#### Limits configuration
```
kacf, line 94: Here, since this is a header, indeterminate "Limits configuration" is fine.

09.Add-ons/90.Reference/docs.md
```diff
@@ -70,16 +91,45 @@ mender-connect along with the default values. The default configuration path is
...
+##### File Transfer limits configuration
+The `FileTransfer` section in the `Limits` configuration block has the following
+options available:
+
+* `Chroot`: limit the directory from which you can transfer files and to which you can upload them.
+* `OwnerGet`: you can only transfer the files owned by the users on this list.
+* `GroupGet`: you can only transfer the files that have a group from this list.
+* `OwnerPut`: all the files you upload to a device will have this username set as an owner.
+* `GroupPut`: all the files you upload to a device will have this group set.
+* `MaxFileSize`: the maximal file sie that you can download from a device.
```
kacf, line 114: `s/sie/size/`.

kacf, line 114: What about upload?

09.Add-ons/90.Reference/docs.md
```diff
@@ -70,16 +91,45 @@ mender-connect along with the default values. The default configuration path is
...
+* `FollowSymLinks`: if set to true, `mender-connect` will resolve all the links in the target or destination path and the transfer will proceed. If false, and if any part of an upload or download path is a link, `mender-connect` will refuse to carry out the request.
```
kacf, line 115: Is this actually true? I'm asking because symlink elements in the middle of the path are really tricky to avoid following, especially if you want to do it securely. I was involved in the CFEngine fixes for this, which are [huge and complicated](https://github.com/cfengine/libntech/blob/828aa81863f96f243ca3b8a824e1e154f7878df1/libutils/file_lib.c#L541). My hunch is that what this really means is that the final component will either be followed, or not followed. But can you enlighten me?

09.Add-ons/90.Reference/docs.md
```diff
@@ -70,16 +91,45 @@ mender-connect along with the default values. The default configuration path is
...
+* `AllowOverwrite`: if set to true, `mender-connect` will overwrite the target file path when processing the upload request. If set to false `mender-connect` will refuse to overwrite the file.
```
kacf, line 116: Can you delete files with this protocol?

merlin-northern: no. we do not have a notion fo delete file command. and we do not support http DELETE.

09.Add-ons/90.Reference/docs.md
merlin-northern: what we actually do is: we follow the full path, I think it resolves all the links, and then we compare with the original, if they match it means that there were no links on the way, including the target (`basename` of the path).  as far as I remember testing it, it worked for links on ever level.

09.Add-ons/90.Reference/docs.md
merlin-northern: good point, I think it also works for upload. fixed.

merlin-northern: fixed. (wow, sorry about that, why my spell check did not catch those?)

09.Add-ons/90.Reference/docs.md
merlin-northern: right. fixed.

09.Add-ons/03.File-Transfer/docs.md
merlin-northern: that, my friend, is a passive voice. you can forget about passive voice in the docs :)

09.Add-ons/03.File-Transfer/docs.md
merlin-northern: fixed.

09.Add-ons/03.File-Transfer/docs.md
merlin-northern: fixed. (wow, sorry for that)

09.Add-ons/90.Reference/docs.md
kacf: With something like `realpath()`? Ok. It's possible to defeat that since it has race conditions (ask me in person if you want to know the details 🙂), but I think we don't need fortress level security here, CFEngine's threat model is a bit different. The meaning is anyway clear, so this is good enough.

09.Add-ons/90.Reference/docs.md
kacf: 👍

kacf (approved):

09.Add-ons/90.Reference/docs.md
merlin-northern: right! you know I do want to know the details. as soon as I am done with two other things on my puny plate (more like a bowl really), I will ask. thanks.

PR #1505, opened 2021-07-15
Update "latest" references to 3.0 documentation.

kacf:


tranchitella (approved):
Sorry, I must have missed it.

PR #1506, opened 2021-07-19
Move test running into a script in order to reuse from other CI jobs.

kacf:

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -ex
+
+./test_autoversion.py
+./autoversion.py --check
+env TEST_OPEN_SOURCE=1 ./test_docs.py 07.Server-installation/03.Installation-with-docker-compose/docs.md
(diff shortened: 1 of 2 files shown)


kacf:
The motivation for this PR is to use it [here](https://github.com/mendersoftware/integration/blob/master/.gitlab-ci.yml#L83)

tranchitella (approved):

run-tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -ex
+
+./test_autoversion.py
+./autoversion.py --check
```
lluiscampos, line 6: I would have kept this part out. These two tests are only for this repository.

lluiscampos (approved):

run-tests.sh
kacf: True, but I think this script is useful to run locally as well, and then these two tests are appropriate. They take literally less than a second to run, so I think I prefer to leave them in so that everything is in one place.

PR #1513, opened 2021-07-23
Clarify compatibility explanations

TheMeaningfulEngineer:
Small improvement suggestions which make the compatibility page more unambiguous.

### DIFF
--- a/02.Overview/14.Compatibility/docs.md
+++ b/02.Overview/14.Compatibility/docs.md
@@ -67,23 +67,30 @@ the lists of specific criteria we use for our versioning policy.
 ## Mender client and Yocto Project version
 
 <!--AUTOVERSION: "% to %"/ignore-->
-In general the Mender client introduces new features in minor (e.g. 1.2.0 to 1.3.0) versions and the [meta-mender layer](https://github.com/mendersoftware/meta-mender?target=_blank) is updated accordingly to easily support these new features (e.g. by exposing new [MENDER_* variables](../../05.System-updates-Yocto-Project/99.Variables/docs.md)). The [meta-mender layer](https://github.com/mendersoftware/meta-mender?target=_blank) has branches corresponding to [versions of the Yocto Project](https://wiki.yoctoproject.org/wiki/Releases?target=_blank).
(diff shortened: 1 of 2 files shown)


tranchitella (approved):

TheMeaningfulEngineer:
Don't know the flow for the docs PR (testing pipelines and prerequisites for merging).
@kacf Let's pls shortly sync on this once you're back.

02.Overview/14.Compatibility/docs.md
```diff
@@ -128,6 +135,11 @@ The [Mender Artifact format](../03.Artifact/docs.md) is managed by the [Mender A
 
 The compatibility between the Mender server and client is managed by the Device API versions exposed by the server and used by the client. If the Mender server supports the API version of the Mender client, they are compatible.  However, please ensure that the client and server support the [Artifact format](#mender-clientserver-and-artifact-format) version you are using. Device API docs are available in the [API chapter](../../200.Server-side-API/?target=_blank#device-apis).
 
+The higher version API is a superset of the previous one. For endpoints which haven't changed in the new version the previous version ones are assumed.
```
kacf, line 138: This is not true, the two APIs are actually incompatible, which is why we bumped the version. But I get your point, it feels a bit like a hybrid because we have kept the v1 endpoints instead of bumping those also to v2.

However, to avoid confusion I would rephrase this a bit: 

> The higher version API contains a mix of old and new API endpoints. For endpoints which haven't changed in the new version the previous version ones are assumed.

Does it seem confusing? Yeah, it is. :slightly_smiling_face: I have created [this ticket](https://tracker.mender.io/browse/MEN-4927) to try to convert it to a more consistent style, where everything is moved to v2.

kacf (approved):
Pre-approving, but please take a look at the commect below.

kacf:
> Don't know the flow for the docs PR (testing pipelines and prerequisites for merging).
> @kacf Let's pls shortly sync on this once you're back.

If it has run the tests and has approval, you can go ahead and merge. The tests are supposed to run automatically, but occasionally there is a hiccup causing them not to run. Looks like this happened for you. However, if you fix the issue I mentioned above, and re-push, probably they will run this time.

02.Overview/14.Compatibility/docs.md
TheMeaningfulEngineer: Great catch, definitely not a superset and much more correct phrasing.
Changing.

PR #1517, opened 2021-08-04
Add `global.auditlogs` to Helm chart installation instructions

tranchitella:
Changelog: none


merlin-northern (approved):
lgtm, do we need to refer it somehow in the text? I will defer to Marcin.

tranchitella:
@merlin-northern I don't think we need to update the text, it's enough to provide the right example.

tranchitella:
This is needed because the released Helm Chart version 3.0.0 requires this flag.

PR #1518, opened 2021-08-05
Make it clear Raspberry Pi OS image no longer contains Mender.

kacf:

### DIFF
--- a/01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md
+++ b/01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md
@@ -29,7 +29,7 @@ To follow this tutorial, you will need the following:
 
 ## Step 1 - Flash the Raspberry Pi
 
-Download the Raspberry Pi OS image with Mender integrated:
+Download the Raspberry Pi OS image ready for Mender:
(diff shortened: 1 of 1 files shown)


kacf:
@mzedel: Was this the only occurrence, or do you have other examples?

01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md
```diff
@@ -29,7 +29,7 @@ To follow this tutorial, you will need the following:
 
 ## Step 1 - Flash the Raspberry Pi
 
-Download the Raspberry Pi OS image with Mender integrated:
+Download the Raspberry Pi OS image:
```
lluiscampos, line 32: While I understand the motivation of this PR, I think we should keep something, otherwise these might very well be links to the upstream images.

This "something" is what I informally call "Mender ready", meaning "an image with dual rootfs, bootloader with Mender integration, and data partition".

So what about "Download the Raspberry Pi OS image ready for Mender:"

kacf: Good point, fixed.

lluiscampos (approved):

kacf:
Cherry-picked to 3.0.x as well. Older releases are still using different images, where this is not true, so not cherry-picking there.

PR #1533, opened 2021-09-07
FIX: Docs test using old integration/keygen parameters

alfrunes:


alfrunes:
@lluiscampos I missed a spot, this should also fix the broken integration pipeline as well.

lluiscampos (approved):

lluiscampos:
Easy to miss... this test runs only on `master`

PR #1551, opened 2021-09-24
Make sure mender-monitor installs dependencies too.

kacf:


lluiscampos (approved):
Thank you

PR #1553, opened 2021-09-24
Update documentation around mender-convert requiring mender.conf

lluiscampos:


lluiscampos:
Not completely correct. I'll rework.

PR #1569, opened 2021-10-20
MEN-5149 Production installation tests: updating containers count.

merlin-northern:
ChangeLog:none


kacf (approved):
:+1:

PR #1579, opened 2021-11-16
Add missing state transition path

Granjow:
### Description

According to the state descriptions, reboot is only entered/exited when the device actually reboots, and if not, the state would probably transition to ArtifactCommit directly, right?

If that is true, I'd convert this PNG to a drawio PNG which can be edited with a draw.io/diagrams.net editor (sources are included in the PNG text area). (Unless the source is already drawn in draw.io?)


Granjow:
I'd also like to have that clarified:

> Therefore it is important to remember that when deploying a new update, all scripts will be run from the currently running root file system until ArtifactInstall, at which point the scripts from the new Artifact will take over.

Is this true also for rootfs updates? Will the ArtifactInstall_Leave be run from the new rootfs on the other partition? Will it be copied to the current partition? Or is the rootfs update a special case where the old scripts until ArtifactReboot_Enter run and after rebooting the scripts from the new rootfs partition are run (i.e. ArtifactReboot_Leave and onwards)?

kacf:
> According to the state descriptions, reboot is only entered/exited when the device actually reboots, and if not, the state would probably transition to ArtifactCommit directly, right?

Hey @Granjow, you're right, this transition is missing from the diagram. I have added them [here](https://github.com/mendersoftware/mender-docs/pull/1582).

> If that is true, I'd convert this PNG to a drawio PNG which can be edited with a draw.io/diagrams.net editor (sources are included in the PNG text area). (Unless the source is already drawn in draw.io?)

They are hosted as Google Draw documents, which unfortunately aren't public, but you can see the result in the link I posted.

> I'd also like to have that clarified:
> 
> > Therefore it is important to remember that when deploying a new update, all scripts will be run from the currently running root file system until ArtifactInstall, at which point the scripts from the new Artifact will take over.
> 
> Is this true also for rootfs updates? Will the ArtifactInstall_Leave be run from the new rootfs on the other partition? Will it be copied to the current partition? Or is the rootfs update a special case where the old scripts until ArtifactReboot_Enter run and after rebooting the scripts from the new rootfs partition are run (i.e. ArtifactReboot_Leave and onwards)?

I think there is a misunderstanding: It says is that "the scripts from the new Artifact" will take over, not that it's switching root filesystem. It means that the scripts are taken from the new artifact, but it has not rebooted yet, so the operating system is still using the old root filesystem. The scripts are run from the data partition, so they are not copied to either of the active/passive rootfs partitions.

Closing this PR, since it is superseded by #1582.

Granjow:
@kacf Thanks for your answer and the updates to the diagrams!

Regarding the rootfs update, I have to state the question a bit more precisely. The docs simply states that at ArtifactInstall the new scripts will take over. I assume there are two different use cases:

* If a reboot is not required, e.g. when installing some piece of software which also installs a state script, then ArtifactInstall will add/overwrite the existing state script and already `ArtifactInstall_Leave` will use that new script.
* If a reboot is required due to a rootfs update, then ArtifactInstall sets up the other partition. So all state scripts that are included in the new version are on the new partition. In that case, I'd assume  that `ArtifactInstall_Leave` still only considers the state scripts of the *old* artifact, and so does `ArtifactReboot_Enter`, and after ArtifactReboot, the state scripts from the *new* artifact (partition) will be used, i.e. from `ArtifactReboot_Leave`  on onwards.

Is that correct?

kacf:
No, that's not correct. All state scripts that start with "Artifact" do not exist on the filesystem at all, so there are no "existing scripts". The "Artifact" scripts are taken from the Artifact and are temporarily stored on the data partition while the installation is happening. Then they are deleted afterwards.

Granjow:
@kacf Okay. How would that work in case of rootfs updates? Can they also contain `Artifact*` scripts or not?

kacf:
> Okay. How would that work in case of rootfs updates? Can they also contain `Artifact*` scripts or not?

Yes, they can.

PR #1584, opened 2021-12-01
Fix relative link in Add-ons/Monitor section

lluiscampos:
The `./XX.Something` syntax is not rendered properly in grav.


oleorhagen (approved):
Thanks :)

PR #1605, opened 2022-01-05
Add information about hddimg and installer

Granjow:
Signed-off-by: Simon A. Eugster <simon.eu@gmail.com>

### Description

Adds example on installing an image on a target system which does not have (easily) removable storage.


kacf:
@Granjow: Thanks! I don't think we can include the part about `hddimg` in the official documentation, because this is not something we test, and I would not expect this to work in all scenarios. I have modified the text a little bit and resubmitted in https://github.com/mendersoftware/mender-docs/pull/1606.

kacf:
Superseded by #1606.

Granjow:
@kacf I understand – thanks for merging!

PR #1612, opened 2022-01-18
Set versions for Mender 3.2.0-build2

oleorhagen:
Changelog: None


kacf (approved):
The code looks good but the receiver branch must be 3.2.x, not master. Pre-approving though.

PR #1615, opened 2022-01-19
MEN-5391: autoversion: Add logic to query either git or docker version

lluiscampos:
With our latest release_tool we cannot anymore use the "trick" of
querying docker versions to pure git components (like mender-artifact,
mender, etc) and expect it to fallback to git version and work.

By default the tool will now use the git versions, unless an special
string is used for in the AUTOVERSION expression to trigger the switch
to docker version. See updated README.

### DIFF
--- a/04.System-updates-Debian-family/02.Convert-a-Mender-Debian-image/docs.md
+++ b/04.System-updates-Debian-family/02.Convert-a-Mender-Debian-image/docs.md
@@ -94,7 +94,7 @@ documentation for your host OS if you encounter connection issues with docker.
 
 Clone `mender-convert` from the official repository:
 
-<!--AUTOVERSION: "-b % https://github.com/mendersoftware/mender-convert"/mender-convert-->
+<!--AUTOVERSION: "-b % https://github.com/"/mender-convert-->
 ```bash
 git clone -b master https://github.com/mendersoftware/mender-convert.git
 ```
--- a/05.System-updates-Yocto-Project/03.Build-for-demo/docs.md
+++ b/05.System-updates-Yocto-Project/03.Build-for-demo/docs.md
@@ -174,7 +174,7 @@ here are the basic steps needed to do this however your actual setup may require
 Please make sure you are standing in the directory where `poky` resides,
 i.e. the top level of the Yocto Project build tree, and run these commands:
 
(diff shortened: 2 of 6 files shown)


oleorhagen:
> @kacf @oleorhagen This is not ideal, but I couldn't think about a better approach.
> 
> I considered modifying the API of the AUTOVERSION expressions to have an extra parameter to explicitly decide on git/docker version, but discarded it as it seems an overkill for **one** expression that we have today in the docs that uses the docker version.

agreed

lluiscampos:
> I considered modifying the API of the AUTOVERSION expressions to have an extra parameter to explicitly decide on git/docker version, but discarded it as it seems an overkill for **one** expression that we have today in the docs that uses the docker version.

I mean something like `<!--AUTOVERSION: "mendersoftware/deployments:%"/deployments/git-->` and `<!--AUTOVERSION: "mendersoftware/deployments:%"/deployments/docker-->`

oleorhagen (approved):

kacf:
> This is not ideal, but I couldn't think about a better approach.

I imagined doing this instead:

```diff
-<!--AUTOVERSION: "mendersoftware/deployments:%"/deployments-->
+<!--AUTOVERSION: "mendersoftware/deployments:mender-%"/integration-->
```

I am not super keen on the change which has been made here, because it is a bit surprising that the pattern plays a role in deciding.

It also removes important matching criteria. For example, imagine someone inserts this, and doesn't think about the AUTOVERSION:

```diff
 <!--AUTOVERSION: "-b % git://github.com/"/meta-mender-->
 git clone -b master git://github.com/mendersoftware/meta-mender
+git clone -b master git://github.com/mendersoftware/mender
```

Previously it would complain, because it matches the full name, now it will instead insert:

```
git clone -b dunfell git://github.com/mendersoftware/meta-mender
git clone -b dunfell git://github.com/mendersoftware/mender
```

IMHO I think we should revert this and use "integration" as the version target instead. What do you think?

lluiscampos:
@kacf Thank you for your feedback. I was very eager to wait for your feedback before merging it, but with the intention of unblocking the release I went for it yesterday.

All right, I completely agree with you and I feel stupid for not having thought about such a simple solution before! I'll revert and follow your advice on this.

lluiscampos:
@kacf See #1619

PR #1619, opened 2022-01-20
autoversion.py: Switch to git versions

lluiscampos:
Reverts #1615 and fixes the problem in a simpler way.

### DIFF
--- a/08.Server-integration/03.Mutual-TLS-authentication/docs.md
+++ b/08.Server-integration/03.Mutual-TLS-authentication/docs.md
@@ -229,7 +229,7 @@ You also need to specify a username and password pair. The ambassador will use i
 To start the edge proxy, run the following command:
 
 <!-- AUTOMATION: execute={ -->
-<!--AUTOVERSION: "registry.mender.io/mendersoftware/mtls-ambassador:%"/mtls-ambassador-->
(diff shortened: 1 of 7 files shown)


kacf (approved):
:+1:

08.Server-integration/03.Mutual-TLS-authentication/docs.md
```diff
@@ -229,7 +229,7 @@ You also need to specify a username and password pair. The ambassador will use i
 To start the edge proxy, run the following command:
 
 <!-- AUTOMATION: execute={ -->
-<!--AUTOVERSION: "registry.mender.io/mendersoftware/mtls-ambassador:%"/mtls-ambassador-->
+<!--AUTOVERSION: "registry.mender.io/mendersoftware/mtls-ambassador:mender-%"/integration-->
```
lluiscampos, line 232: @kacf This has a problem now on `hosted` branch, where integration version resolves to `staging` and (obviously) not `3.2.0` as we would like to.

What could we do here? I have no capacity nor ideas right now so I'll create a task to capture the details

oleorhagen: Just stumbled across this

PR #1634, opened 2022-01-25
wrong version of mtls-ambassador in staging

oleorhagen:
fails the test, hence it's not deploying.


oleorhagen:
it will pass, but it really should be 3.2 right?

oleorhagen:
Point is, hosted is not deploying cause of this

tranchitella (approved):
Approving, but as mentioned on slack I believe the version `mender-staging` is wrong and we need to fix it.

oleorhagen:
> Approving, but as mentioned on slack I believe the version `mender-staging` is wrong and we need to fix it.

agreed. Just want to deploy the site.

PR #1658, opened 2022-02-08
Downloads: Use HTML links to prevent Grav from mangling the URL.

kacf:
We need the '%2B' character, which is a '+' sign, to remain as is,
otherwise the download doesn't work. But Grav auto-translates them,
so use HTML links to avoid that.

### DIFF
--- a/09.Downloads/docs.md
+++ b/09.Downloads/docs.md
@@ -261,23 +261,18 @@ Mender in [standalone
 mode](../02.Overview/01.Introduction/docs.md#client-modes-of-operation).
 
 <!--AUTOVERSION: "keeps \"%\" version"/ignore-->
-<!-- The second column points to pre-release software and keeps "master" version in the name and link -->
-<!--AUTOVERSION: "mender-client %][mender-client_x.x.x"/mender "mender-client %][mender-client_%_"/ignore-->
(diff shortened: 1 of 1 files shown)


09.Downloads/docs.md
```diff
@@ -261,23 +261,18 @@ Mender in [standalone
...
+|----------------|------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| armhf (ARM-v6) | ARM 32bit distributions, for example Raspberry Pi OS for Raspberry Pi or Debian for BeagleBone | <a class="mender_docs_versioned_link" href="https://downloads.mender.io/master/dist-packages/debian/armhf/mender-client_master-1%2Bdebian%2Bbuster_armhf.deb">mender-client master</a> | <a class="mender_docs_latest_link" href="https://downloads.mender.io/master/dist-packages/debian/armhf/mender-client_master-1%2Bdebian%2Bbuster_armhf.deb">mender-client master</a> (Pre-release) |
```
lluiscampos, line 273: We use "latest" usually to refer to "latest stable" release. Not that it matters much because it is not visible to users, but maybe for our own sanity I suggest renaming to `mender_docs_pre_release_link`

PR #1670, opened 2022-02-14
Link to lastest mender-montior deb.

merlin-northern:
Customers require a reliable link to the latest mender-monitor package.
It is already in use in the Mender deployments, so we should document
and support it.

ChangeLog:none

### DIFF
--- a/09.Downloads/docs.md
+++ b/09.Downloads/docs.md
@@ -436,3 +436,23 @@ Then install the package with:
 ```bash
 dpkg -i mender-monitor_demo_master-1_all.deb
 ```
+
+Please note that the latest Debian package can be found at:
+
+
+[ui-tabs position="top-left" active="0" theme="lite" ]
+[ui-tab title="hosted"]
+<!--AUTOVERSION: "mender-monitor %][x.x.x_mender-monitor-"/mender-monitor "mender-monitor %][%_mender-monitor-"/ignore-->
+[mender-monitor master][master_mender-monitor-debian]
+[/ui-tab]
+[ui-tab title="enterprise"]
+<!--AUTOVERSION: "mender-monitor %][x.x.x_mender-monitor-"/mender-monitor "mender-monitor %][%_mender-monitor-"/ignore-->
+[mender-monitor master][master_mender-monitor-debian-onprem]
+[/ui-tab]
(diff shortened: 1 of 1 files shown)


09.Downloads/docs.md
```diff
@@ -436,3 +436,23 @@ Then install the package with:
 ```bash
 dpkg -i mender-monitor_demo_master-1_all.deb
 ```
+
+Please note that the latest Debian package can be found at:
+
+
+[ui-tabs position="top-left" active="0" theme="lite" ]
+[ui-tab title="hosted"]
+<!--AUTOVERSION: "mender-monitor %][x.x.x_mender-monitor-"/mender-monitor "mender-monitor %][%_mender-monitor-"/ignore-->
```
lluiscampos, line 445: I don't really understand this autoversion expressions - what are you trying to ignore?

09.Downloads/docs.md
```diff
@@ -436,3 +436,23 @@ Then install the package with:
...
+[mender-monitor master][master_mender-monitor-debian]
```
lluiscampos, line 446: We are missing the actual instructions to use the enterprise credentials:
```
HOSTED_MENDER_EMAIL=<your.email@example.com>
curl --fail -u "$HOSTED_MENDER_EMAIL" -O ...
```

See links above

lluiscampos (commented):
I personally don't think we should publish these links in the docs.

These were meant to be consumed by get.mender.io, and by publishing them here we are forced to support them in the future.

I pinged @eysteins to see his opinion.

merlin-northern:
> I personally don't think we should publish these links in the docs.
> 
> These were meant to be consumed by get.mender.io, and by publishing them here we are forced to support them in the future.
> 
> I pinged @eysteins to see his opinion.

I will defer to @eysteins to make the decision.
The reason for this PR is: I believe we should support them.

estenberg:
I am skeptical of this, especially since it is inconsistent with the other tools and I don't see the strong case for it (except onboarding). Maybe I'm missing something, is this more important for Monitor than others, or is it "just" that one user was using these specifically?

https://docs.mender.io/downloads#monitor looks fine to me to be honest.

PR #1681, opened 2022-02-16
3.0.x: Correct the deb packages filenames in the install commands

lluiscampos:
Partial cherry-pick of #1676.


merlin-northern (approved):
as a review:
* checked that debs exist
* checked that around dpkg there are no other URL encoded versions (pulling the branch and grepping)

PR #1685, opened 2022-02-17
Switch from curl to wget for commercial packages download

lluiscampos:
wget renames the files correctly. This way also the download commands
are consist with the OS packages.


merlin-northern (approved):
lgtm. I checked that wget to hosted works for existing packages (there is a known issue with master-1+debian+buster) and also grepped on this branch in search for curl.*deb -- no results.

PR #1691, opened 2022-02-21
Update dockumentation to make docker container independent from mender-convert repository

sensslen:
### Description
This is a companion PR to https://github.com/mendersoftware/mender-convert/pull/433

### DIFF
--- a/04.System-updates-Debian-family/02.Convert-a-Mender-Debian-image/docs.md
+++ b/04.System-updates-Debian-family/02.Convert-a-Mender-Debian-image/docs.md
@@ -121,8 +121,9 @@ The easiest, and most straight-forward way, is to integrate the client with
 
 #### Using [hosted Mender](https://hosted.mender.io?target=_blank)
 ```bash
-./scripts/bootstrap-rootfs-overlay-hosted-server.sh \
-    --output-dir ${PWD}/rootfs_overlay_demo \
+mkdir -p input
+$MENDER_CONVERT_LOCATION/bootstrap-rootfs-overlay-hosted-server.sh \
+    --output-dir ${PWD}/input/rootfs_overlay_demo \
     --tenant-token "Paste token from https://hosted.mender.io/ui/#/settings/my-organization"
 ```
 
@@ -136,6 +137,7 @@ Move your *golden disk image* into an input subdirectory:
 ```bash
 mkdir -p input
 mv <PATH_TO_MY_GOLDEN_IMAGE> input/golden-image-1.img
+mv <PATH_TO_OVERLAY> input/overlay_folder
 ```
 
 ### Use the mender-convert container image
@@ -143,10 +145,15 @@ mv <PATH_TO_MY_GOLDEN_IMAGE> input/golden-image-1.img
 Run mender-convert from inside the container with your desired options, e.g.
 
 ```bash
+# move overlay to the input folder
+mkdir -p input/overlay
+mv <PATH_TO_MY_OVERLAY>/* input/Overlay/*
+
+# convert the image
 MENDER_ARTIFACT_NAME=release-1 ./docker-mender-convert \
     --disk-image input/golden-image-1.img \
     --config configs/raspberrypi3_config \
-    --overlay rootfs_overlay_demo/
+    --overlay input/overlay/
 ```
 
 Conversion will take 10-30 minutes, depending on image size and resources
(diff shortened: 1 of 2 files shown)


04.System-updates-Debian-family/02.Convert-a-Mender-Debian-image/docs.md
```diff
@@ -90,39 +90,18 @@ permissions to connect to the docker daemon. In Ubuntu 18.04, the user must be a
...
-```bash
-./docker-build
-```
-
 ## Configure the Mender client server configuration
 
 The easiest, and most straight-forward way, is to integrate the client with
-[hosted Mender](https://hosted.mender.io?target=_blank):
+[hosted Mender](https://hosted.mender.io?target=_blank). For this you should 
+crate an overlay e.g. using the scripts in the `scripts/` folder of
```
kacf, line 97: ~crate~ create

04.System-updates-Debian-family/02.Convert-a-Mender-Debian-image/docs.md
```diff
@@ -146,7 +126,7 @@ Run mender-convert from inside the container with your desired options, e.g.
 MENDER_ARTIFACT_NAME=release-1 ./docker-mender-convert \
```
kacf, line 126: `docker-mender-convert` isn't going to exist if the repository isn't cloned first. And having looked at its contents, I realize quite a lot of stuff is happening inside it. The idea I had was that using the tool would be something like:

```bash
docker run mendersoftware/mender-convert --disk-image etc...
```

But now I'm wondering if it's even possible, since the required command inside `docker-mender-convert` is quite long. Any suggestions?

sensslen: I see your point - for me the "docker-mender-convert" file acts as an example on how to use mender-convert inside a docker container. It is kind of like part of the documentation (and of course used in the ci builds). Of course there is stuff happening there (which is mostly creating folders to seperate mount points) but it can be seen as "providing the ground for the container to work". For my appliance I would use the container directly without docker-mender -convert....

kacf: It's not really optional arguments though; it won't work without the volume flags, for example. Pretty much [all of these flags](https://github.com/mendersoftware/mender-convert/blob/03e5c1bcc8efa3fb14932630e5501670c73cc135/docker-mender-convert#L35-L43) are mandatory; do you call it with all of them?

If only we could do that from inside Docker, but right now I don't see how this is easier than what we already have.

sensslen: From what I understand about docker: Mount Points are in fact optional. So if I don't need access to the logs, I can safely omit the mount point. The reason for all of this is that this in fact allows to use mender-convert without checking out the repository and provide a clean enviroment for it to use. With the current approach - where the docker container does not contain the application - It may well stop working all of a sudden because the application changes and so change the required software components required. 

When building the docker container in each run the issue is that the docker container may well change because canonical decides to update a package used.

Both of these scenarios are not desirable. 

The solution to this is to build a docker container for each version of mender convert and then make sure that you always use the correct combination. I wouldn't call this especially easy.

According to this reasoning I'd much rather live with a more verbose command line syntax than to require users to ensure they provide matching components.

For Now I changed the documentation to not reference docker-mender-convert and provide the raw command line.

kacf: I agree completely with your points, but we put a high emphasis on the usability when using the tool on the command line, and this is compromising that. We would welcome a change which could somehow deal with the problems you describe, and provide a similar or shorter command line, if you can find a way to do that. But without that I think we have to reject this, I'm sorry.

sensslen: This is very sad - especially as I am not able to follow your reasoning. I mean the current usage of the docker container (using dockeer-mender-convert) is still possible and does not introduce any new required command line parameters. However there is the possibility to also use the docker container directly (which was not possible before).

kacf: Alright, point taken about the docker container itself. This is something we could upload during releases and provide. I was hoping this would be a replacement for the current workflow, but it has to be an addition. This means that only a small subset of this pull request is relevant, namely the ones dealing with the `input` folder. The rest should stay as is. Sorry for making you do this work unnecessarily.

There are some changes needed in https://github.com/mendersoftware/mender-convert/pull/433 as well; I'll post additional comments there.

kacf (approved):

PR #1694, opened 2022-02-22
Small fix python example for mender configure

oleorhagen:
As you can see from logs created inside the same file on the device this is the value of sys.argv.

14:41:02,820 root INFO ['/usr/lib/mender-configure/apply-device-config.d/example_script.py', '/var/lib/mender-configure/device-config.json'

Also run into troubles with the arguments of subprocess.run so I prefered to let all the command inside a list.

Changelog: None


kacf (approved):
Good catch!

oleorhagen:
@mender-test-bot cherry-pick to: `hosted`, `3.2.x`

PR #1700, opened 2022-03-02
RetryPollCount description

merlin-northern:
ChangeLog:none

### DIFF
--- a/03.Client-installation/07.Configuration-file/50.Configuration-options/docs.md
+++ b/03.Client-installation/07.Configuration-file/50.Configuration-options/docs.md
@@ -154,6 +154,15 @@ to run, before it is considered hung and killed. The process will first be sent
 a SIGTERM signal, and one minute later, if it has not exited, a SIGKILL signal
 is sent. The default is 4 hours.
 
+#### RetryPollCount
+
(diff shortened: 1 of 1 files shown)


kacf (changes requested):
* I would swap the order of the two `RetryPoll..` options to keep the alphabetic ordering.

* I would also add to the description:
  > If the setting is zero (the default), the maximum number of retries is `3 * ceil(log2(RetryPollIntervalSeconds) + 1)`.

  Many people [have asked about this](https://hub.mender.io/t/artifact-installs-failing-on-tried-maximum-amount-of-times/4272/2), so I think it makes sense to post the full formula. Agree?

merlin-northern:
> * I would swap the order of the two `RetryPoll..` options to keep the alphabetic ordering.
> * I would also add to the description:
>   > If the setting is zero (the default), the maximum number of retries is `3 * ceil(log2(RetryPollIntervalSeconds) + 1)`.
>   
>   
>   Many people [have asked about this](https://hub.mender.io/t/artifact-installs-failing-on-tried-maximum-amount-of-times/4272/2), so I think it makes sense to post the full formula. Agree?

have you actually ever seen me not agree? :)

kacf (approved):

PR #1701, opened 2022-03-05
chore: improve documentation for the monitoring add-on

tranchitella:
Changelog: none

### DIFF
--- a/01.Get-started/05.Monitor/docs.md
+++ b/01.Get-started/05.Monitor/docs.md
@@ -15,8 +15,8 @@ allows you to monitor various parts of your system.
 
 ## Prerequisites
 
-To follow this tutorial, you will need to install the [Monitor
-Add-on](../../09.Add-ons/20.Monitor/10.Installation/docs.md) on your device. If
+To follow this tutorial, you will need to [install the Monitor
+Add-on on your device](../../09.Add-ons/20.Monitor/10.Installation/docs.md). If
 you have followed the [get started tutorial to prepare your
 device](../01.Preparation/docs.md), the Monitor Add-on should already be
 installed.
@@ -47,14 +47,15 @@ shows how to detect that a USB-connected peripheral gets disconnected.
 First enable the alert through running:
 
 ```bash
-mender-monitorctl enable log usb_disconnect
+sudo mender-monitorctl enable log usb_disconnect
 ```
 
- Now remove a USB device from the device (you can insert it first, e.g. a
- thumbdrive or mouse, if you don't have any USB devices inserted). The Alert now
- triggers in the Mender UI once you remove the USB device:
+Now remove a USB device from the device (you can insert it first, e.g. a
+thumbdrive or mouse, if you don't have any USB devices inserted). Once you remove
+the USB device, the log subsystem triggers an alert which you can inspect in the
+device details in the Mender UI:
 
-![Connectivity alarm OKd](log-usb-alarm.png)
+![Connectivity alarm OK](log-usb-alarm.png)
 
 #### Disk usage
 
@@ -66,7 +67,7 @@ remediate it before it causes any downtime.
 First enable the diskusage alert for the root space partition.
 
 ```bash
-mender-monitorctl enable diskusage root_space
+sudo mender-monitorctl enable diskusage root_space
 ```
 
 Then check the current disk usage.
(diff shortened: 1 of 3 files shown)


lluiscampos (approved):
Thank you

01.Get-started/05.Monitor/docs.md
```diff
@@ -47,14 +47,14 @@ shows how to detect that a USB-connected peripheral gets disconnected.
...
-mender-monitorctl enable log usb_disconnect
+sudo mender-monitorctl enable log usb_disconnect
 ```
 
- Now remove a USB device from the device (you can insert it first, e.g. a
- thumbdrive or mouse, if you don't have any USB devices inserted). The Alert now
- triggers in the Mender UI once you remove the USB device:
+Now remove a USB device from the device (you can insert it first, e.g. a
+thumbdrive or mouse, if you don't have any USB devices inserted). The Alert now
+triggers in the Mender UI, in the device details, once you remove the USB device:
```
merlin-northern, line 55: suggested change: "The Alert now...device:" -> "_Once you remove the USB device, the log subsystem triggers an alert which you can inspect  in the device details in the Mender UI:_"

01.Get-started/05.Monitor/docs.md
```diff
@@ -105,6 +105,12 @@ And an alert shows up in the UI.
 
 ![Disk usage alarm triggered](diskusage-alarm.png)
 
+If you remove the file, the alert hides in the UI:
```
merlin-northern, line 108: I would rephrase "the alert hides" somehow it does not seem a good fit, perhaps "the alert is cancelled"? I mean: we are not hiding the alert, we are having device in normal operations state wrt to this check.

01.Get-started/05.Monitor/docs.md
```diff
@@ -171,23 +175,25 @@ Extract the container name:
 CONTAINER_NAME=$(docker ps | tail -1 | awk '{print $NF}')
 ```
 
-Create the file `dockerevents_testcontainer.sh` with the contents:
+Create the file `/etc/mender-monitor/monitor.d/available/dockerevents_testcontainer.sh` running the following command:
```
merlin-northern, line 178: so `mender-monitorctl` does not  support `create` subcommand on the `dockerevents` subsystem? if not then we need a ticket for it.

merlin-northern (commented):
thanks. dropped some comments, treat as suggestions.

oleorhagen (approved):

snusifer:
@merlin-northern 
> I would rephrase "the alert hides" somehow it does not seem a good fit, perhaps "the alert is cancelled"? I mean: we are not hiding the alert, we are having device in normal operations state wrt to this check.

Yeah, the alert is never hidden. What happens when the reader deletes the file is that the user will get an OK alert, with a green check mark. The old alert for when the user created the file will always stay in the alert history.

snusifer:
With all the added `sudo`s now, i think reword [this](https://github.com/mendersoftware/mender-docs/blob/08629236c05d291da56d9a2bdb0796c55916bf22/01.Get-started/05.Monitor/docs.md?plain=1#L37-L39) a bit differently?

01.Get-started/05.Monitor/docs.md
tranchitella: Sorry, I don't get what you mean. Can you please follow-up on this topic?

01.Get-started/05.Monitor/docs.md
```diff
@@ -162,7 +167,7 @@ events and notifies you in the event that a container was restarted.
 First spin up the container to monitor:
 
 ```bash
-docker run --rm -d alpine sleep infinity
+sudo docker run --rm -d alpine sleep infinity
```
snusifer, line 170:
```suggestion
sudo docker run --name demo --rm -d alpine sleep infinity
```

01.Get-started/05.Monitor/docs.md
```diff
@@ -171,23 +176,25 @@ Extract the container name:
 CONTAINER_NAME=$(docker ps | tail -1 | awk '{print $NF}')
```
snusifer, line 176: We could just give this container a name, and not depend on that `$CONTAINER_NAME` variable. Then we can remove this step.

01.Get-started/05.Monitor/docs.md
merlin-northern: sure: https://tracker.mender.io/browse/MEN-5507

tranchitella:
@snusifer @oleorhagen what's the status of this PR? Can we complete it and merge it?

oleorhagen:
@snusifer has created a follow-up PR here: https://github.com/mendersoftware/mender-docs/pull/1713

This should be closed.

PR #1711, opened 2022-03-11
Draft: mender-artifact reference documentation

TheYoctoJester:
First draft of reference documentation for `mender-artifact`. Not publication ready, subject to discussion+feedback.


TheMeaningfulEngineer:
Hey, skimmed a bit.
The description and the upcoming examples are great introduction to the tool.
The duplication of the CLI help I don't see the value given the risk of having the two out of sync.

TheYoctoJester:
@TheMeaningfulEngineer yes, already discussed it a bit with @oleorhagen. The main problem that the CLI help has is that one need to run the CLI in order to read it. And its not searchable, but more like a poke-and-discover thing.

The best solution would probably be to have `mender-artifact` provide a man page, and autogenerate the documentation reference straight from it. Then no risk of de-syncing would exists, Ole already offered to look into a setup for this.

PR #1727, opened 2022-03-30
feat: add section about rate limits in overview/security

tranchitella:
Ticket: MEN-5270


alfrunes (approved):
Pre-approving, although we should expose the IP-based rate limit as a configuration in the Helm chart and Compose installation.

kjaskiewiczz (approved):

TheMeaningfulEngineer:
@tranchitella 
Hey,
This is only a nice to have if there are technical reasons to not add this to the docs (i.e. changing it faster without having to have the docs follow).
Can you expand that with the rate limits currently applied to HM?

PR #1747, opened 2022-05-06
mender-cli no longer supports just two use cases and will be extended

estenberg:
in the future. Remove the explicit listing.

### DIFF
--- a/08.Server-integration/01.Using-the-apis/docs.md
+++ b/08.Server-integration/01.Using-the-apis/docs.md
@@ -13,10 +13,7 @@ There are many ways to interact with Mender's REST APIs and the most common ones
 
 `mender-cli` is a standalone CLI tool that works as a client against the Mender server management APIs in order to make it much easier to interact with the APIs.
 
-Currently two use cases are supported:
-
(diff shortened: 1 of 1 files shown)


estenberg:
This corrects an issue, but mender-cli may be a bit under-marketed in general, doesn't seem like many people know about it.
Granted, it is a bit feature limited at this time, but can still be useful for the key features it supports. :)

tranchitella (approved):

TheYoctoJester (approved):
Generally agreed, but the casing feels off. Do we have a taxonomy decision on the upper-/lower-case of
- "Artifact"
- "Remote terminal"
- "Port forwarding"

I have spotted various versions throughout the existing documentation.

estenberg:
@TheYoctoJester these are (feature/object) names, that's why they are uppercase. Taxonomy page is here: https://docs.mender.io/overview/taxonomy. We do have a design doc for the docs as well, let me know if you don't have it.

That said, this can likely be adjusted / made more consistent, especially when a name has "Multiple Words" (vs. Multiple words).

PR #1809, opened 2022-07-25
3.3.x cherry-picks

tranchitella:


merlin-northern (approved):
there is a passive voice there ;-)

PR #1821, opened 2022-08-15
Fixed typo in 'Getting started' tutorial

larsewi:
Removed '.' after '!'. `s/!./!`

### DIFF
--- a/01.Get-started/03.Deploy-a-system-update/docs.md
+++ b/01.Get-started/03.Deploy-a-system-update/docs.md
@@ -182,7 +182,7 @@ this modification *is not part of your system snapshot* created above.
 !! Be careful when running `apt upgrade` on a device with Mender system updates enabled. Integration
 !! with `apt upgrade` (through the `grub.d` framework) is only implemented for x86 as of
 !! mender-convert version master. For ARM and other non-x86 architectures, always update single
-!! applications only, because running `apt upgrade` may brick your device!. If you need to run `apt
+!! applications only, because running `apt upgrade` may brick your device! If you need to run `apt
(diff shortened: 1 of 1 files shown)


01.Get-started/03.Deploy-a-system-update/docs.md
```diff
@@ -182,7 +182,7 @@ this modification *is not part of your system snapshot* created above.
 !! Be careful when running `apt upgrade` on a device with Mender system updates enabled. Integration
 !! with `apt upgrade` (through the `grub.d` framework) is only implemented for x86 as of
 !! mender-convert version master. For ARM and other non-x86 architectures, always update single
-!! applications only, because running `apt upgrade` may brick your device!. If you need to run `apt
+!! applications only, because running `apt upgrade` may break your device! If you need to run `apt
```
lluiscampos, line 185: `brick` is intentional, though. In embedded or IOT world, it is an (slang?) way of saying that the device became unusable due to software miss-function (so it is like a brick, you can't interact with it anymore). Also `break` will most likely mean physically broken which is not the case here. See [wikipedia](https://en.wikipedia.org/wiki/Brick_(electronics)).

The dot typo after the `!` is good, so please amend this PR to only submit the latter.

lluiscampos (changes requested):

01.Get-started/03.Deploy-a-system-update/docs.md
larsewi: Awesome, thanks! :rocket: I did not know that

lluiscampos (approved):

PR #1822, opened 2022-08-16
Add definition of DFU to the first use of the abbreviation

larsewi:
I had to look up the definition of the DFU. By adding the definition to
the first use of the abbreviation, other IoT newbies like my self won't
have to do this.


lluiscampos (approved):
:+1:

PR #1840, opened 2022-09-07
test: Add missing AUTOMATION tag in Downloads page

lluiscampos:


09.Downloads/docs.md
```diff
@@ -263,6 +263,7 @@ repository. Afterwards, you can install and update the Mender client using the
    ```
    [/ui-tab]
    [ui-tab title="Ubuntu Jammy"]
+<!--AUTOMATION: ignore -->
```
alfrunes, line 266: Isn't this just working around the original issue?

lluiscampos: The test cannot install one repository (the one corresponding to the distro running in CI, Ubuntu Focal). If you look above, you can see that all the other distros are also ignored.

When we publish this respository, then we could upgrade both CI and test to use Ubuntu Jammy.

oleorhagen (approved):

PR #1919, opened 2022-10-26
chore: Revamp the release and support policy page

TheMeaningfulEngineer:
Ticket: None

Changelog: None

Signed-off-by: Alan <alan.martinovic@northern.tech>

### DIFF
--- a/302.Release-information/01.Release-schedule/docs.md
+++ b/302.Release-information/01.Release-schedule/docs.md
@@ -1,15 +1,40 @@
 ---
-title: Release schedule
+title: Release & support policy
 taxonomy:
     category: docs
 ---
 
-Our requirements for release quality are always the same, with only the length of support differing between non-LTS releases and LTS releases.
-LTS releases are maintained through patch releases for one year, whereas non-LTS releases get patch releases only if an urgent vulnerability is found between minor releases.
 
-The lifecycle of an LTS release begins on the release date and ends one year later at the scheduled support end date. Keeping these dates in mind helps decide which version to adopt and when to upgrade to a new version. To ensure the optimal operation of Mender within a release lifecycle, the release components should use any released patch updates once they become available.
+####  Release Cadence
 
-##### Current LTS releases
+New releases are released approximately every three months, either as an LTS (Long Term Support) or a non-LTS.
+
+####  Support policy
+
+**LTS**
+* LTS (Long Time Supported release) are maintained through patch releases for one year
+* Non critical patch releases for LTS are published within a month after the latest release
+* In case of an urgent vulnerability discovered, the LTS can be released earlier
+* The lifecycle of an LTS begins on the release date and ends one year later - the table below lists those dates
+
+**Non-LTS**
+* If the Non-LTS is the latest release, it might get a patch releases in the exceptional case of an urgent vulnerability
+* Any non-LTS which isn't the latest release version will never get patch releases
+
+
+Example
+
+```
+2022-06   3.3.0     LTS is released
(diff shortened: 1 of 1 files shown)


302.Release-information/01.Release-schedule/docs.md
```diff
@@ -1,15 +1,40 @@
...
-##### Current LTS releases
+New releases are released approximately every three months, either as an LTS (Long Term Support) or a non-LTS.
+
+####  Support policy
+
+**LTS**
+* LTS (Long Time Supported release) are maintained through patch releases for one year
+* Non critical patch releases for LTS are published within a month after the latest release
+* In case of an urgent vulnerability discovered, the LTS can be released earlier
+* The lifecycle of an LTS begins on the release date and ends one year later - the table below lists those dates
```
kacf, line 18: I would change this whole section a bit:

```
* LTS (Long Time Supported) releases are maintained for one year
* On average, patch releases are released every three months, but in case of security
  vulnerabilities or other important bug fixes, patch releases may be made available sooner.
* The lifecycle of an LTS begins on the release date and ends one year later -
  the table below lists those dates
```

In general we don't make any timing guarantees, so it's important to express this.

302.Release-information/01.Release-schedule/docs.md
```diff
@@ -1,15 +1,40 @@
...
+
+**Non-LTS**
+* If the Non-LTS is the latest release, it might get a patch releases in the exceptional case of an urgent vulnerability
+* Any non-LTS which isn't the latest release version will never get patch releases
```
kacf, line 22: I would change this a bit too:

```
* Non-LTS releases are maintained until the next minor release. New minor releases are made
  available every three months on average, and can be either LTS or non-LTS.
* Non-LTS releases will receive patch releases in case of security vulnerabilities or other important
  bug fixes, otherwise it will be succeeded by the next minor release.
```

302.Release-information/01.Release-schedule/docs.md
```diff
@@ -1,15 +1,40 @@
...
+Example
+
+```
+2022-06   3.3.0     LTS is released
+2022-09   3.4.0     Non-LTS is released
+2022-10   3.3.1     LTS is released - a patch version of 3.3
+2022-09   3.4.0     Non-LTS is released
+2022-12   3.5.0     Non-LTS is released
+2023-01   3.3.2     LTS is released - a patch version of 3.3
+```
```
kacf, line 34: I don't think we should have this example. Patch releases aren't fixed, and we are frequently deviating from this (we are about to release 3.4.1, for example). If you want to give an example, it's better to focus on the maintenance time, so something like this:

```
2022-06   3.3.0     LTS is released, maintenance of 3.3.0 begins.
2022-09   3.4.0     Non-LTS is released, maintenance of 3.4.0 begins.
2022-12   3.5.0     LTS is released, maintenance of 3.4.0 ends, and 3.5.0 begins.
2023-03   3.6.0     Non-LTS is released, maintenance of 3.6.0 begins
2023-06   3.7.0     LTS is released, maintenance of 3.3.0 and 3.6.0 ends, and 3.7.0 begins.
```

You could also consider a visual option, which is IMHO a better way of visualizing it, like this example here:

LTS visualized

We should make it clear that this is just a plan though, and we are not guaranteed to release exactly at these times. The only thing which is *guaranteed*, are the currently maintained LTS releases today, that is the only promise we make.

kacf (changes requested):
Good to have this improved a bit! I do have some requests though, see below.

We can take care of the CI failures once we agree on the content.

PR #1926, opened 2022-11-08
QA-453

mzedel:


checklinks.js
```diff
@@ -17,7 +15,6 @@ engine(
     extensions: ['md'],
     pluginPrefix: 'remark',
     packageField: 'remarkConfig',
-    injectedPlugins: [parse, stringify],
```
oleorhagen: why is this not needed anymore?

oleorhagen (approved):
Thanks!

checklinks.js
mzedel: because of the `pluginPrefix` defined above - any package with the _remark_ prefix is considered a plugin 👌

oleorhagen: Nice!

mzedel:
 * 3.4.x
 * 3.3.x

PR #1929, opened 2022-11-10
chore: Document mender-client Debian package contents and limitations

lluiscampos:
Ticket: ME-13

### DIFF
--- a/09.Downloads/docs.md
+++ b/09.Downloads/docs.md
@@ -94,6 +94,15 @@ software updates packaged as Mender Artifacts.
 See [Client installation](../03.Client-installation/chapter.md) for more information
 about how to configure and use the Mender client.
 
+The `mender-client` Debian package installs:
+* the binary,
+* a systemd service,
+* the default [identity script](../03.Client-installation/03.Identity/docs.md),
+* the default [inventory scripts](../03.Client-installation/04.Inventory/docs.md)
+* and the default [update modules](../03.Client-installation/05.Use-an-updatemodule/docs.md)
+(and its generators).
+
+
 ### Installation methods
 
 You can install the Mender client in different ways depending on your preference.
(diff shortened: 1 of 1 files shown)


09.Downloads/docs.md
```diff
@@ -94,6 +94,14 @@ software updates packaged as Mender Artifacts.
 See [Client installation](../03.Client-installation/chapter.md) for more information
 about how to configure and use the Mender client.
 
+The `mender-client` Debian package installs the binary, a systemd service, the default [identity
+script](../03.Client-installation/03.Identity/chapter.md), the default [inventory
+scripts](../03.Client-installation/04.Inventory/chapter.md) and the default [update
+modules](../03.Client-installation/05.Use-an-updatemodule/chapter.md) (and its generators).
+
+!!! If you customize any of the above (for example modifying identity or inventory scripts)
```
TheMeaningfulEngineer, line 102: I'd move this in the upgrade section instead.
It's there where this gets experienced (you do `apt-get upgrade` -> no more customisation

lluiscampos: Okey. I am moving there the warning but keep here the description (the above paragraph) - I think the description makes sense to keep it in `Mender client` section.

09.Downloads/docs.md
```diff
@@ -94,6 +94,14 @@ software updates packaged as Mender Artifacts.
 See [Client installation](../03.Client-installation/chapter.md) for more information
 about how to configure and use the Mender client.
 
+The `mender-client` Debian package installs the binary, a systemd service, the default [identity
```
TheYoctoJester, line 97: I would prefer this to be in list form
```suggestion
The `mender-client` Debian package installs
- the client binary
- a systemd service
- the default [identity script](../03.Client-installation/03.Identity/chapter.md)
...
```

Rationale: for readers it is much easier to skim the list if they are looking for something specific, rather than block text. Content wise I am fine.

09.Downloads/docs.md
```diff
@@ -94,6 +94,14 @@ software updates packaged as Mender Artifacts.
...
+script](../03.Client-installation/03.Identity/chapter.md), the default [inventory
+scripts](../03.Client-installation/04.Inventory/chapter.md) and the default [update
+modules](../03.Client-installation/05.Use-an-updatemodule/chapter.md) (and its generators).
+
+!!! If you customize any of the above (for example modifying identity or inventory scripts)
+!!! they will get overridden when upgrading or re-installing the `mender-client` Debian package
```
TheYoctoJester: This could be a bit expanded. Like

TheYoctoJester: This could be a bit expanded. Example:

```suggestion
!!! If you customize any of the above ...  then plese make sure to also save your work in an additional place. Files at pathes that match the defaults shipped by the package will be overwritten if the client is upgraded or re-installed, so you might lose your work if you only modified the original files.
```

TheYoctoJester (changes requested):

TheYoctoJester (approved):

PR #1968, opened 2022-12-12
chore: Document `mender-artifact` dependency on OpenSSL 1.1

lluiscampos:


TheYoctoJester (approved):

10.Downloads/docs.md
```diff
@@ -62,6 +62,15 @@ Follow the correct link according to your host platform to download
 | Linux    | [mender-artifact master][x.x.x_mender-artifact-linux]  | [mender-artifact master][master_mender-artifact-linux] (Pre-release)  |
 | Mac OS X | [mender-artifact master][x.x.x_mender-artifact-darwin] | [mender-artifact master][master_mender-artifact-darwin] (Pre-release) |
 
+!!! The `mender-artifact` pre-built binaries depend on OpenSSL 1.1 shared library. If you are
+!!! running a system that already migrated to OpenSSL 3, like Alpine Linux 3.17 or Ubuntu 22.04, the
+!!! binary cannot be run directly. Follow one of these workarounds:
```
tranchitella, line 67:
```suggestion
!!! running a system that has already migrated to OpenSSL 3, like Alpine Linux 3.17 or Ubuntu 22.04, you
!!! cannot run the binary directly. Follow one of these workarounds:
```

10.Downloads/docs.md
```diff
@@ -62,6 +62,15 @@ Follow the correct link according to your host platform to download
...
+!!! * For Alpine Linux, install `openssl1.1-compat` package
```
tranchitella, line 68:
```suggestion
!!! * For Alpine Linux, install the `openssl1.1-compat` package
```

10.Downloads/docs.md
```diff
@@ -62,6 +62,15 @@ Follow the correct link according to your host platform to download
...
+!!! * For Ubuntu 22.04 or newer, the recommended process is to install `mender-artifact` through the
+!!! [Mender APT repositories](#install-using-the-apt-repository).
+!!! * For others where the distribution does not provide a compatibility package, build
+!!! `mender-artifact` from source.
```
tranchitella, line 72:
```suggestion
!!! * For the other cases where the distribution does not provide a compatibility package, build
!!! `mender-artifact` from the source.
```

tranchitella (approved):
I approved with a few minor wording suggestions.

10.Downloads/docs.md
TheMeaningfulEngineer, line 72: FWIW
For me this ended up more complicated, using the docker container approach that @mzedel shared was much quicker.

```
docker build -t mender-artifact . and then
docker run -v $PWD:/go/bin/data mender-artifact <anyExtraArgs>… (edited) 
```

The only downside is I still haven't figured yet how to add the docker based approach to path so it can be used with artifact generators for update modules...

PR #1982, opened 2023-01-02
ci: Hard-code mender-ci-workflows-version to 1.0.0

lluiscampos:
As this component is out of the release_tool, the ci test has no way to know which is the latest version (for now).

Amends commit 72079d479137940b6afcecf35c2f8271585dc335 and it will be reworked together with the original commit once we have a design for where to store the latest version of independent components.


kacf (approved):
Is there a task we can refer to?

lluiscampos:
> Is there a task we can refer to?

Not yet. Actually, today Ole and me are having a workshop to start breaking down this work and this specific problem is on my notes :memo:

PR #1985, opened 2023-01-09
feat: add a paragraph about Device Systems in the Mender Gateway docs

tranchitella:
Changelog: Title
Ticket: MEN-6035

### DIFF
--- a/01.Get-started/06.Mender-Gateway/docs.md
+++ b/01.Get-started/06.Mender-Gateway/docs.md
@@ -95,6 +95,39 @@ You can inspect the configuration file `/etc/mender/mender-gateway.conf`:
 Please refer to the [Mender Gateway's configuration file](../../08.Server-integration/04.Mender-Gateway/99.Configuration-file/)
 documentation for detailed information about the various settings.
 
+## Device Systems
+
+A Device System is a group of devices connected to a Mender Gateway instance and belonging to the same product or logical system.
+To define a Device System, each device must report to the Mender Server a special inventory attribute named `mender_gateway_system_id`, containing a unique identifier for the given system.
+
+The Mender Gateway can set the System ID for all the devices connected to it thanks to the `SystemID` configuration setting. When enabled, this feature injects this inventory attributes for all the devices connected to the gateway.
+
+An example configuration follows:
+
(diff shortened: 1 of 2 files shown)


TheYoctoJester (approved):

01.Get-started/06.Mender-Gateway/docs.md
```diff
@@ -95,6 +95,36 @@ You can inspect the configuration file `/etc/mender/mender-gateway.conf`:
...
+		"Listen": ":443",
+		"ServerCertificate": "/usr/share/doc/mender-gateway/examples/cert/cert.crt",
+		"ServerKey": "/usr/share/doc/mender-gateway/examples/cert/private.key"
+	},
+	"Features": {
+		"ArtifactsProxy": {
+			"Enabled": true,
+			"GatewayURL": "https://gateway.docker.mender.io",
+			"DomainWhitelist": ["s3.amazonaws.com"]
+		}
```
lluiscampos, line 120: (sorry if I wasn't clear yesterday),

For consistency with the rest of the _opt-in_ features, we still need an explicit boolean to enable the feature
```suggestion
	"Features": {
		"ArtifactsProxy": {
			"Enabled": true,
			"GatewayURL": "https://gateway.docker.mender.io",
			"DomainWhitelist": ["s3.amazonaws.com"]
		},
		"DeviceSystems" {
		        "Enabled" : true
		}
```

The fact that `SystemID` is in the root of the configuration file I see it more of a convenience. Same as we have for example `UpstreamServer` in the root and then used for `ArtifactsProxy` feature.

Another argument is that having a feature toggle on/off based on an string being empty or not is rather obscure. As a user I might want to disable the feature but keep the string there for reference or re-enabling it in the future (json does not allow comments)

lluiscampos (changes requested):

01.Get-started/06.Mender-Gateway/docs.md
tranchitella: Right, reworked (as `DeviceSystem`, singular, to make it consistent with other features/options).

lluiscampos (approved):

PR #1999, opened 2023-02-02
fix: disk image compression is none by default

casambi-jeikki:
In mender-convert config (configs/mender_convert_config) default value for MENDER_COMPRESS_DISK_IMAGE is gzip but because it is missing quotation marks the default behavior corresponds to the same as "none" meaning the mender-convert by default leaves the disk image uncompressed.

It would be better to change the default value to "none" in mender-convert repository in addition to this PR but I leave it for you to decide what to do.

### DIFF
--- a/04.System-updates-Debian-family/99.Variables/docs.md
+++ b/04.System-updates-Debian-family/99.Variables/docs.md
@@ -160,7 +160,7 @@ The version of the Mender Configure add-on to include in the update.
 
 #### `MENDER_COMPRESS_DISK_IMAGE`
 
-> Values: gzip(default)/lzma/none
+> Values: gzip/lzma/none(default)
 
 This is useful when you have large disk images, compressing them makes it easier
 to transfer them between a build server and a local machine, and saves space.


lluiscampos:
Hi @casambi-jeikki 

> In mender-convert config (configs/mender_convert_config) default value for MENDER_COMPRESS_DISK_IMAGE is gzip but because it is missing quotation marks the default behavior corresponds to the same as "none" meaning the mender-convert by default leaves the disk image uncompressed.

What I think is happening is that your input image is uncompressed and `mender-convert` is overriding the `MENDER_COMPRESS_DISK_IMAGE` value to match the input image (in your case to "none"). This is a feature introduced some time ago, see https://github.com/mendersoftware/mender-convert/pull/241

Is this the case?

You can modify this behavior by setting `MENDER_COMPRESS_DISK_IMAGE=gzip` in your own config file and pass it along with `--config my_config_file`.

lluiscampos:
I launched a separate PR to clarify this matter. Let me know if it makes sense to you!
* https://github.com/mendersoftware/mender-docs/pull/2011

casambi-jeikki:
@lluiscampos Makes sense. Thanks for clarification :)

PR #2006, opened 2023-02-04
fix: mTLS ambassador runs as non-root

tranchitella:
Changelog: None
Ticket: None


lluiscampos (approved):

lluiscampos:
Thank you @tranchitella :+1:

PR #2014, opened 2023-02-10
chore: Update Raspberry Pi pre-built images

lluiscampos:


oleorhagen (approved):
Nice, so no issues with the images this time?

lluiscampos:
> Nice, so no issues with the images this time?

No issues in October when we did the upgrade. It is only that we forgot to follow-up updating `master` docs after we decided not to release Mender 3.4.1 :+1:

PR #2026, opened 2023-02-21
QA-348: ci: adjust test to run on relevant integration branch

0lmi:
Changelog: None
Ticket: QA-348

- [ ] Make sure that all commits follow the conventional commit [specification](https://www.github.com/mendersoftware/mendertesting/commitlint/grammar.md) for the Mender project.

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!

### DIFF
--- a/test_docs.py
+++ b/test_docs.py
@@ -76,6 +76,14 @@ with open(sys.argv[1]) as f:
     contents = ""
     for line in f.readlines():
         line = line.strip()
+        if line.startswith("git clone"):
+            integration_branch = os.environ.get("INTEGRATION_BRANCH")
+            if integration_branch is not None:
+                default_branch = re.search(r"-b ([a-zA-Z0-9\.\-\_]*) ", line).group(1)
+                line = line.replace(default_branch, integration_branch)
+                line = line.replace(
+                    "github.com/mendersoftware", "gitlab.com/Northern.tech/Mender"
+                )
         if re.search(ignore_line, line):
             ignore_next_codeblock = True
             ignore_count += 1


test_docs.py
```diff
@@ -75,6 +75,12 @@
     inside_bash_code_block = False
     contents = ""
     for line in f.readlines():
+        if line.startswith("git clone"):
+            int_branch_env_var_name = "INTEGRATION_BRANCH"
+            integration_branch = os.environ.get(int_branch_env_var_name)
```
lluiscampos, line 80: If it is only used here, there is much value on having it as a variable.
```suggestion
            integration_branch = os.environ.get("INTEGRATION_BRANCH")
```

test_docs.py
```diff
@@ -75,6 +75,12 @@
     inside_bash_code_block = False
     contents = ""
     for line in f.readlines():
+        if line.startswith("git clone"):
```
lluiscampos, line 78: This will not really work. The instructions in the docs are `git clone -b ... github...` while on our pull requests `INTEGRATION_BRANCH` will only exist in GitLab (remember that we run the pipelines from these `pr_...` branches.

test_docs.py
```diff
@@ -75,6 +75,12 @@
...
+            int_branch_env_var_name = "INTEGRATION_BRANCH"
+            integration_branch = os.environ.get(int_branch_env_var_name)
+            if integration_branch is not None:
+                default_int_branch = "master"
```
lluiscampos, line 82: This will only work on mender-docs/master, but not in hosted or any of the release branches

lluiscampos (changes requested):
I see it is an smart hack to modify only the `git clone` commands, but there are some caveats that will make this not working. See comments below.

Also, please find a way to test this in CI. For example you can launch some debug PRs in `integration` that have a hardcoded version (one containing the changes of this PR) for `mender-docs` branch to use. And you should also test it with `mender-docs` branches based on `hosted` and one release branch.

test_docs.py
0lmi: It checks if a line starts from `git clone` substring, I tested it and pretty sure it works.

test_docs.py
0lmi: That is how it was originally done, but it doesn't pass python formatting check.

test_docs.py
0lmi: Good point 👍

test_docs.py
0lmi: Ok, now I got what you mean here. The comment is actually not about the check, but about the logic behind.

0lmi: Good point 👍. Added the logic for GitHub URL replacement to GitLab.

0lmi:
CI part is improved to be able to run pipelines for any combination of integration<->mender-docs branches.

There is job's [example](https://gitlab.com/Northern.tech/Mender/integration/-/jobs/3813863713#L1150) of running the pipeline from [pr_2339](https://gitlab.com/Northern.tech/Mender/integration/-/commits/pr_2339) branch in GitLab (which corresponds to integration/2339 PR) with [pr_2026](https://gitlab.com/Northern.tech/Mender/mender-docs/-/tree/pr_2026) branch (which corresponds to mender-docs/2026 PR).

test_docs.py
```diff
@@ -76,6 +76,15 @@
     contents = ""
     for line in f.readlines():
         line = line.strip()
+        if line.startswith("git clone"):
+            integration_branch = os.environ.get("INTEGRATION_BRANCH")
+            if integration_branch is not None:
+                words = line.split(" ")
+                default_int_branch = words[3]
```
lluiscampos, line 83: This is quite weak. Using a regex you can make it more robust and fail explicitly when the expression does not match.

lluiscampos (approved):

PR #2031, opened 2023-02-22
Operating system updates

tranchitella:


01.Get-started/02.Deploy-an-application-update/docs.md
```diff
@@ -125,5 +125,6 @@ using Mender!
 
 Proceed to one of the following tutorials (listed in recommended order):
 
-1. [Deploy a system update](../03.Deploy-a-system-update/docs.md)
-1. [Deploy a container update](../04.Deploy-a-container-update/docs.md)
+1. [Deploy an Application update](../02.Deploy-an-application-update/docs.md)
```
lluiscampos, line 128: This is a self-reference. At this point the reader has tone the Application update and the "next steps" should be only the other two tutorials.

lluiscampos (approved):
One thing needs amending, see below. Pre-approving :+1:

01.Get-started/02.Deploy-an-application-update/docs.md
tranchitella: oh, right

PR #2063, opened 2023-03-27
chore: rename `Mender server` to `Mender Server`

tranchitella:
Changelog: None


oldgiova (approved):

merlin-northern (approved):
that was you commenting on "we usually use Mender Server" comment, right? thank you!

PR #2100, opened 2023-05-03
fix: other broken links to download the raspberry PI images

tranchitella:


tranchitella:
Merging without review to speed up fixes in the production docs.

PR #2125, opened 2023-06-06
chore: Simplify commands to generate RSA key pair

lluiscampos:
The second command in the private key generation had no effect. We put the two together for simplification and consistency with how ECDSA keys are generated below.

### DIFF
--- a/06.Artifact-creation/07.Sign-and-verify/docs.md
+++ b/06.Artifact-creation/07.Sign-and-verify/docs.md
@@ -45,22 +45,16 @@ After generating the keys you will have a file `private.key`, which you use only
 
 #### RSA
 
-You can generate a private RSA key by executing the command below:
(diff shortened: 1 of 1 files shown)


06.Artifact-creation/07.Sign-and-verify/docs.md
```diff
@@ -45,16 +45,10 @@ After generating the keys you will have a file `private.key`, which you use only
 
 #### RSA
 
-You can generate a private RSA key by executing the command below:
+In order to generate a public and private RSA key pair use the commands below:
```
tranchitella, line 48: Feel free to ignore: I usually replace "In order to" with "To".

```suggestion
To generate a public and private RSA key pair use the commands below:
```

tranchitella (approved):

PR #2137, opened 2023-06-21
Replace index with number to match what the code is actually looking for

robertalexa:
During implementation at my end, i've discovered some discrepancies between some variables referenced in the docs and what mender-convert actually looks for.

PS: there may be other discrepancies but the ones submitted are the only ones i have looked into, for now.

### DIFF
--- a/04.System-updates-Debian-family/99.Variables/docs.md
+++ b/04.System-updates-Debian-family/99.Variables/docs.md
@@ -85,7 +85,7 @@ in the configuration.
 
 > Value:
 
-This option allows fine grained control of the boot partition device path and overrides `MENDER_STORAGE_DEVICE` and `MENDER_BOOT_PART_INDEX` settings. If part-uuid is used then you will need to also enable `MENDER_ENABLE_PARTUUID`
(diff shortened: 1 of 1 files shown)


lluiscampos (changes requested):
@robertalexa Good catch! This was changed [long time ago](https://github.com/mendersoftware/mender-convert/pull/160) but it went unnoticed until now. Thank you very much :rocket:

The change is good, but there is one formality to take care of: the commit needs to be signed. Check the [contribution](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md) guide.

Once you amend your commit with the signature I'll merge it and port it to all relevant branches.

robertalexa:
@lluiscampos All done now. Commit message does not contain a note about signing, but if you check the Verified mark you can see it is signed with my personal key rather than the previously used Github (i have submitted the PR via the website itself)

Hope this is what you were after!

lluiscampos (approved):
@robertalexa Yes, it looks good now!

lluiscampos:
* master
* 3.5.x
* 3.3.x

PR #2200, opened 2023-08-17
Docker tags

tranchitella:


lluiscampos (commented):
Good catch, thanks.

Once you fix the missing AUTOVERSION tag (CI failing) these references will be updated automatically during releases.

lluiscampos (approved):

PR #2213, opened 2023-08-23
fix: sort the LTS releases from the most recent to the oldest

tranchitella:
Changelog: None
Ticket: NOne


merlin-northern (approved):
if it is a cherry-pikc, do we use -x?
minor: "Ticket: NOne" -> "Ticket: None" MAy be :)

tranchitella:
> if it is a cherry-pikc, do we use -x? minor: "Ticket: NOne" -> "Ticket: None" MAy be :)

This is the original commit (which I already cherry picked to `hosted`), so no `-x`.
Capitalization is wrong but I cannot change it without breaking the other cherrypicks. Therefore, We'Ll LiVe WiTh It.

PR #2218, opened 2023-08-30
fix: fixed an issue that prevented access to the download page due to an inbalance of tab closures

mzedel:
- rebalanced tab shortcodes after 0117c6127f4a96457368ae608e3eb51aedb386e8

Ticket: None
Changelog: None


tranchitella (approved):

lluiscampos:
Great catch (or guess) :tada:

PR #2219, opened 2023-08-31
Docs: Adds local mender-configure artifacts generation

MuchoLucho:

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!

### DIFF
--- a/09.Add-ons/10.Configure/docs.md
+++ b/09.Add-ons/10.Configure/docs.md
@@ -21,7 +21,7 @@ It is possible to create and apply configuration for the device using Mender UI 
 
 Mender applies configuration in a similar way as software updates.
 `deviceconfig` is a microservice on the server responsible for "translating" a configuration change made in the UI into a deployment.
-It then proceeds to trigger a deployment to the device using the same mechanism used in OTA, the `deployments` microservice. 
+It then proceeds to trigger a deployment to the device using the same mechanism used in OTA, the `deployments` microservice.
 
 On the device side, a specialized update module is responsible for applying the configuration.
 You can find more information about the configuration update module on [Mender Hub](https://hub.mender.io).
@@ -34,4 +34,57 @@ The picture below shows the configuration update flow.
 There are no special device requirements for using Mender Configure, but editing the configuration
 from the UI is not supported if the device is using **signed Artifacts**. It is still possible to
 see reported configuration the UI. Configure Artifacts can be
-deployed with a manually signed Artifact, by following the regular Artifact deployment flow.
+deployed with a manually signed Artifact, by following the regular Artifact deployment flow as described in [Handle signed Artifacts](#handle-signed-artifacts).
+
+## How to deploy a configuration to multiple devices at once
+
+The `mender-configure` add-on was designed to be executed from the UI for one device at a time however, you can generate the artifact manually and then create deployments to groups (static or dynamic) taking advantage of the `mender-configure` Update Module:
+
+### Prerequisite
+
+You will need a script already located in `/usr/lib/mender-configure/apply-device-config.d` as described in ["How to make an apply-device-config script"](01.Device-integration/docs.md#how-to-write-an-apply-device-config-script) section of these docs.
+
+### Create the `mender-configure` artifact
+
+1. Create a json file that represents the key=value pair (In this example with the name `device-config.json`):
+
+```json
+{
+    "key":"value"
+}
+```
+
+2. Create an artifact with the following command, replace all the data between `<angle-brackets>`.
+
+!!!!! Please note the `-m` flag points to the previously created file, this adds the json as metadata which is expected by the `mender-configure` Update Module on the device side.
+
+```bash
+mender-artifact write module-image  \
+    -T "mender-configure" \
+    -o "<an-artifact-name>.mender" \
+    -n "<an-artifact-name>" \
+    -t "<your-device-type>" \
+    --software-version "<1>" \
+    --software-filesystem "data-partition" \
+    -m "device-config.json"
+```
+
+Then just upload it to the _Releases_ section of the Mender Server and you are ready to create the deployment in the _Deployment_ UI for groups of devices.
+
+#### Handle signed Artifacts
+
+You can take advantage of this approach to handle signed artifacts as you can include signatures from your workstation as opposed by using the Mender Configure UI per device to deploy a configuration. Just generate the key and add the `-k private.key` flag into the previous command as described in the [Artifact Signing section](../../06.Artifact-creation/07.Sign-and-verify/docs.md#signing).
+
+The command should look something like this instead:
+
+```bash
+mender-artifact write module-image  \
+    -T "mender-configure" \
+    -o "<an-artifact-name>.mender" \
+    -n "<an-artifact-name>" \
+    -t "<your-device-type>" \
+    --software-version "<1>" \
+    --software-filesystem "data-partition" \
+    -m "device-config.json" \
+    -k "private.key"
+```


09.Add-ons/10.Configure/docs.md
```diff
@@ -35,3 +35,38 @@ There are no special device requirements for using Mender Configure, but editing
 from the UI is not supported if the device is using **signed Artifacts**. It is still possible to
 see reported configuration the UI. Configure Artifacts can be
 deployed with a manually signed Artifact, by following the regular Artifact deployment flow.
+
+## How to deploy a configuration to multiple devices at once
+
+The `mender-configure` add-on was designed thinking on being executed from the UI for one device at a time, however, you can generate the artifact manually and then create deployments to groups (static or dynamic) taking advantage of the `mender-configure` Update Module:
```
kacf, line 41: Maybe rephrase:

```
The `mender-configure` add-on was designed to be executed from the UI for one device at a time
```

09.Add-ons/10.Configure/docs.md
```diff
@@ -35,3 +35,38 @@ There are no special device requirements for using Mender Configure, but editing
...
+
+### Prerequisite
+
+You will need an script already localted in `/usr/lib/mender-configure/apply-device-config.d` as described in ["How to make an apply-device-config script"](01.Device-integration/docs.md#how-to-write-an-apply-device-config-script) section of these docs.
```
kacf, line 45: "You will need **a** script already **located** in..."

09.Add-ons/10.Configure/docs.md
```diff
@@ -35,3 +35,38 @@ There are no special device requirements for using Mender Configure, but editing
...
+    -T "mender-configure" \
+    -o "<an-artifact-name>.mender" \
+    -n "<an-artifact-name>" \
+    -t "<your-device-type>" \
+    --software-version "<1>" \
+    --software-filesystem "data-partition" \
+    -m "device-config.json"
+```
+
+Then just upload it to the Mender Server and you are ready to create the deployment for groups of devices.
```
kacf, line 72: Maybe also mention that you need to deploy from the normal Deployment section of the UI, not from the Configure section. It should be obvious, since Configure is per-device, but just to make it super clear.

kacf (approved):
Looks good, just a couple of minor corrections from me. I'm approving now so you can merge immediately after fixing.

09.Add-ons/10.Configure/docs.md
kacf: I think this should fix the passive voice checker as well.

09.Add-ons/10.Configure/docs.md
```diff
@@ -35,3 +35,38 @@ There are no special device requirements for using Mender Configure, but editing
...
+}
+```
+
+2. Create an artifact with the following command, replace all the data between `<angle-brackets>`.
+
+!!!!! Please note the `-m` flag points to the previously created file, this adds the json as metadata which is expected by the `mender-configure` Update Module on the device side.
+
+```bash
+mender-artifact write module-image  \
+    -T "mender-configure" \
```
TheMeaningfulEngineer, line 63: Cool!
So device type `mender-configure` combined with the configuration key value pairs as json in the meta data is the 'configuration format'?

TheMeaningfulEngineer: "T" is the update module name, not device type.

09.Add-ons/10.Configure/docs.md
MuchoLucho: Handled in 77fac55c119ebb15acf141faf240312c4e841f2b

09.Add-ons/10.Configure/docs.md
MuchoLucho: Handled in 77fac55c119ebb15acf141faf240312c4e841f2b

09.Add-ons/10.Configure/docs.md
MuchoLucho: Handled in 77fac55c119ebb15acf141faf240312c4e841f2b

MuchoLucho:
Hello folks, I added a new part of "signed mender-configure" artifacts. Can any of you give me your blessing?

09.Add-ons/10.Configure/docs.md
```diff
@@ -34,15 +34,15 @@ The picture below shows the configuration update flow.
...
 
 ## How to deploy a configuration to multiple devices at once
 
-The `mender-configure` add-on was designed thinking on being executed from the UI for one device at a time, however, you can generate the artifact manually and then create deployments to groups (static or dynamic) taking advantage of the `mender-configure` Update Module:
+The `mender-configure` add-on was designed to be executed from the UI for one device at a time however, you can generate the artifact manually and then create deployments to groups (static or dynamic) taking advantage of the `mender-configure` Update Module:
 
 ### Prerequisite
 
-You will need an script already localted in `/usr/lib/mender-configure/apply-device-config.d` as described in ["How to make an apply-device-config script"](01.Device-integration/docs.md#how-to-write-an-apply-device-config-script) section of these docs.
+You will need a script already localted in `/usr/lib/mender-configure/apply-device-config.d` as described in ["How to make an apply-device-config script"](01.Device-integration/docs.md#how-to-write-an-apply-device-config-script) section of these docs.
```
kacf, line 45: `localted` -> `located`

09.Add-ons/10.Configure/docs.md
```diff
@@ -69,4 +69,22 @@ mender-artifact write module-image  \
     -m "device-config.json"
 ```
 
-Then just upload it to the Mender Server and you are ready to create the deployment for groups of devices.
+Then just upload it to the _Releases_ section of the Mender Server and you are ready to create the deployment in the _Deployment_ UI for groups of devices.
+
+#### Handle signed Artifacts
+
+You can take advantage of this approach to handle signed artifacts as you can include signatures from your workstation as opposed by using the Mender Configure UI per device to deploy a configuration. Just generate they key and add the `-k private.key` flag into the previous command as described in the [Artifact Signing section](../../06.Artifact-creation/07.Sign-and-verify/docs.md#signing).
```
kacf, line 76: "...generate **the** key..."

kacf (approved):
Looks good!

TheMeaningfulEngineer (changes requested):
Nice!
One thing that is missing is how will the UI behave once you deploy the configuration this way.

The reason why this isn't fully equivalent to using the UI is that this way only the `reported` (of the [configuration API](https://docs.mender.io/api/#management-api-device-configure-get-device-configuration) get changed while the `configure` remains the same.
If you use the API the `configured` changes first and that triggers the generation of the new update.

image

You should confirm that the UI will function normally after using this apprach and document that.

MuchoLucho:
image
I can confirm after usin this approach the UI reflects the change made by using the Artifact instead of using the UI itself

TheMeaningfulEngineer:
@lluiscampos @MuchoLucho 
With the latest discussion where we shouldn't add things to the docs that aren't supported, I'd vote for moving this to mender hub instead.
The configure feature has been developed to work with one device only for the moment.
This is a workaround how to make it work with two.
Seems like ti works ok, but no guarantee that this will be the case in the future.

So I vote for mender hub instead the official docs.

lluiscampos:
FYI @estenberg as docs owner,

> So I vote for mender hub instead the official docs.

Me too :raising_hand_man: 

Closing the PR. @MuchoLucho If capacity permits, follow-up in Mender Hub :innocent:

estenberg:
This is useful, but agree it's fair to have in MH instead since we don't conveniently support multi-device-updates to configure.

However, @MuchoLucho I think a better approach could be instead of creating the Artifacts you just make a script that calls the deviceconfig API to change a configuration attribute? I can take a list of Device IDs and a JSON (or key values).

PR #2228, opened 2023-10-03
Update configuration reference to reflect support in C++ client.

kacf:


kacf:
@oleorhagen: Check out the SSL settings especially!

oleorhagen (approved):

lluiscampos (approved):

kacf:
Fabio was talking about a potential 3.7 release yesterday, before this will be finished, so I played it safe and created a feature branch instead of master, like the other repos.

PR #2234, opened 2023-10-09
Hosted

tranchitella:


lluiscampos (approved):
The update looks good to me.

However after merge the test to detect the correct versions will fail because `integration/staging` is not updated correctly, see [here](https://github.com/mendersoftware/integration/blob/staging/git-versions.yml#L47) and [here](https://github.com/mendersoftware/integration/blob/staging/other-components.yml#L6) (mender-convert, mender-artifact and others have old versions).

(Arguably, this we can move from as we modernize the release process, @tranchitella . I am just trying to explain how things are designed as per today)

tranchitella:
@lluiscampos what about https://github.com/mendersoftware/integration/pull/2451, and in particular https://github.com/mendersoftware/integration/pull/2451/commits/169880f9cc21f3371f7369ed7483ecf43d9fa663?

TheMeaningfulEngineer (approved):

PR #2243, opened 2023-10-27
fix: Emphasised _Error states are run instead of _Leave states

dimsanius:
During implementation I was quite confused that even though usual state script flow and error state script flow is different, there seemingly is no handling of error states. However, after looking through examples I have realised that `_Leave` state scripts are not run on error and instead `_Error` scripts are run. I just wanted to put more emphasise on this specific idea as it is quite crucial in handling and distinguishing usual flow from the error flow.


kacf (approved):
Thanks, I'm also submitting this to master: #2247.

PR #2253, opened 2023-11-10
Add alert of the Update Control Map feature deprecation

estape11:
Ticket: MEN-6648
Changelog: None

- [ ] Make sure that all commits follow the conventional commit [specification](https://github.com/mendersoftware/mendertesting/blob/master/commitlint/grammar.md) for the Mender project.

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!


TheMeaningfulEngineer (approved):

scripts/run-passive.sh
```diff
@@ -28,5 +28,5 @@ for file in $files; do
   fi
 done
 
-exit $EXIT_CODE
+exit 0
```
tranchitella, line 31: Why?

tranchitella (changes requested):

scripts/run-passive.sh
TheMeaningfulEngineer: This check hinders more than it helps and we couldn't get around it ("We will remove this feature in 4.0") sounded to aggressive.

Also is we wanted to workaround the passive checker the we'd just had to submit this same content just as part of a large commit:

> The script always outputs any passive voice it finds, but only fails on default
> if there are more than 3% in the file total. This cutoff can be modified through


It's a hacky workaround on our end, I agree.
It was late Friday and it's an inappropriate hack, in hindsight would of been more transparent to just delete it all together and ask for disagreements.

So, do you mind if we kill it with the argument that it slows down changes in the docs but provides no value?

kacf: I would not mind. It's meant to enforce better language, and it does, sometimes. But it has way too many false positives.

I agree that if it's killed, it should be killed properly, not just silently disabled in a way that makes it look like it passes.

nickanderson: perhaps instead of a check, it could just comment on the issue with the findings.

TheMeaningfulEngineer: I feel this is overprioritizing a thing which has close to no impact to quality of the docs in there current form.
There are things inconsistent content in the docs, concepts need to be cleared out and the copy paste examples that need to be fixed. 

I can imagine this gathered a lot of attention because of my expedient hack which I agree deserves every scold. :)

I also agree that checking content for language is important and especially one written by non native speakers.
For that we (CE doc reviews) push large content changes through grammarly.
And once we're out of topics to improve in the docs, we might even push the entire content through it.
And at some point might even explore some kind of continuous integration (docs is all public stuff).

So I'm going to remove the script the proper way and am open for a debate in a live call if someone wants to challenge that or a veto from the docs owner.

kacf (approved):

tranchitella (changes requested):
Sorry, I continue to disagree with the idea of disabling the passive voice checker.
I've been hit by it many times, and it always pushed me to improve the text. We (non-native English speakers) often tend to use passive voice more often and this is a general reminder that we should keep an eye on it.

If you want to by-pass it that's fine but please don't remove it.

Add `allow_failure: true` instead:

```
test:passive-voice:
    ...
    allow_failure: true
```

kacf:
Ok, if people get value from it, I suppose it shouldn't be removed. I don't think we should use `allow_failure` though, then we will never look at it, and it will become useless almost immediately (errors will accumulate, and every PR will have tons of preexisting errors).

If people still want to keep it, then I suggest that @TheMeaningfulEngineer use the trick I've used sometimes: If you find that it's complaining about something you can't or shouldn't rewrite, put a line break between the two words that it's complaining about. Then it will ignore that specific phrase.

tranchitella (approved):
We agreed to remove the passive voice checker as it seems most of the team sees value in it.

PR #2259, opened 2023-11-15
chore: remove the unsafe update module and make migration ent only

TheMeaningfulEngineer:
Suggesting an unsafe update module in the docs is unsafe, even with the note.
We promise brick free devices, so suggesting risky steps in the docs doesn't align with that message.

Ticket: None
Changelog: None


MuchoLucho:
I agree, still we should evaluate if we want to translate the Update Module as a Hub post

MuchoLucho (approved):

PR #2165, opened 2023-07-14
CE reviews - mtls

TheMeaningfulEngineer:
Rewrite of the mtls section to provide a copy pastable working example.

PR #2272, opened 2023-12-08
fix: misprints in the MTLS env. variables definition

merlin-northern:
Changelog: Title
Ticket: None

### DIFF
--- a/08.Server-integration/03.Mutual-TLS-authentication/docs.md
+++ b/08.Server-integration/03.Mutual-TLS-authentication/docs.md
@@ -222,8 +222,8 @@ You also need to specify a username and password pair. The ambassador will use i
 
 <!--AUTOMATION: ignore -->
 ```bash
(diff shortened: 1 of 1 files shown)


TheMeaningfulEngineer (changes requested):
We're rewriting this whole section: https://github.com/mendersoftware/mender-docs/pull/2165

merlin-northern:
> The current docs don't give result in a working example. We're rewriting this whole section: #2165 It is almost done to give a copy-paste way to set up everything.
> 
> I'd reject this and hope to have your review on the other PR soon.

I checked the other one, and you fixed it there, so I think when you merge #2165 you can close this one.

tranchitella (approved):

TheMeaningfulEngineer (approved):
Think this is just cosmetics which doesn't improve the mtls docs.
But the mtls rewrite is taking ages so can't push back with that argument.
Approving.

PR #2308, opened 2024-01-17
Added UBoot integration checklist steps

dimsanius:
I have decided to check latest available bullseye RPi OS release for compatibility with Mender. When doing so, I used  `Integration checklist` page of a manual as a guide to confirm a successful integration.

However, the default integration of Debian based RPi OS image is via UBoot, however there were no detailed steps within a manual on the way to check the integrations with UBoot bootloader. There was only a brief that mentions a possible difference of integration either via GRUB or UBoot but there were no practical steps showcased for UBoot integration. So, I fixed that. 

PS  I used [this file](https://github.com/mendersoftware/mender-docs/blob/hosted/01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md) as inspiration

### DIFF
--- a/03.Client-installation/08.Integration-checklist/docs.md
+++ b/03.Client-installation/08.Integration-checklist/docs.md
@@ -186,12 +186,22 @@ We will confirm the bootloader can read the environment and is behaving correctl
 In the previous step, we identified the currently running partition to be `nvme0n1p2` and the inactive one `nvme0n1p3`.
 Set the bootloader variables manually so it boots from the currently inactive partition on the next reboot.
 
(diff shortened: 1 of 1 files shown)


TheMeaningfulEngineer:
Pretty nice, thanks for the contribution!
I'll take a look at this by the end of the month, need to render it and test that the commands are copy pastable and working for both cases.

dimsanius:
Thanks for the acknowledgement! Will keep an eye on this PR

TheMeaningfulEngineer (approved):
Tested this, works.
Thanks

PR #2325, opened 2024-02-22
chore: add a troubleshooting section to the Debian packages upgrade page

tranchitella:
Changelog: None
Ticket: None

### DIFF
--- a/03.Client-installation/02.Install-with-Debian-package/01.Upgrading/docs.md
+++ b/03.Client-installation/02.Install-with-Debian-package/01.Upgrading/docs.md
@@ -87,6 +87,24 @@ In addition, this package will also automatically remove:
 After the upgrade, please update your scripts invoking the `mender` CLI interface to use the
 correct binaries (`mender-auth` or `mender-update`).
 
+For old Debian and Ubuntu distributions, if the Mender add-ons Debian packages (e.g.,
+`mender-connect` and `mender-configure`) are already installed in the system, the upgrade
+can fail with an error message similar to this:
+
+<!--AUTOVERSION: "Conflicts: mender-client but 1:%-1+debian+buster is to be installed"/ignore-->
+```
+The following packages have unmet dependencies:
+ mender-client4 : Conflicts: mender-client but 1:3.5.2-1+debian+buster is to be installed
+E: Unable to correct problems, you have held broken packages.
+```
+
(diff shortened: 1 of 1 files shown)


03.Client-installation/02.Install-with-Debian-package/01.Upgrading/docs.md
```diff
@@ -87,6 +87,23 @@ In addition, this package will also automatically remove:
 After the upgrade, please update your scripts invoking the `mender` CLI interface to use the
 correct binaries (`mender-auth` or `mender-update`).
 
+If the Mender add-ons Debian packages (e.g., `mender-connect` and `mender-configure`) are already
+installed in the system, the upgrade can fail with an error message similar to this:
```
lluiscampos, line 91: It only affects old `dpkg`/`apt`. I would rephrase like:
```suggestion
For old Debian and Ubuntu distributions, if the Mender add-ons packages (e.g., `mender-connect` and `mender-configure`) are already
installed in the system, the upgrade can fail with an error message similar to this:
```

03.Client-installation/02.Install-with-Debian-package/01.Upgrading/docs.md
```diff
@@ -87,6 +87,23 @@ In addition, this package will also automatically remove:
...
+ mender-client4 : Conflicts: mender-client but 1:3.5.2-1+debian+buster is to be installed
+E: Unable to correct problems, you have held broken packages.
+```
+
+In this case, ensure to upgrade the Mender add-ons Debian packages at the same time, for
+example running:
+
+<!--AUTOMATION: ignore -->
+```bash
+apt-get install mender-client4 mender-connect mender-configure
```
lluiscampos, line 105: Have you tested this explicitly? I have only tested the workaround documented in the [test](https://github.com/mendersoftware/mender-dist-packages/blob/ab857f522e48888998f9ae38cdb6bb1ca07cc51a/tests/test_install_mender_sh.py#L273):
```suggestion
apt-get install mender-update mender-client4
```

tranchitella: Yes, though your command is better so I'm applying it.

lluiscampos (approved):
Thank you

PR #2336, opened 2024-03-15
fix: correct typo in Artifact-creation session

Mygao:
### Description

Fixed one typo on Artifact-creation document:)


kacf:
Submitted to master in #2337. It will make its way to hosted eventually.

PR #2373, opened 2024-05-16
Update versions for mender-convert 4.2.1 release.

kacf:


kacf:
Good catch, it should be. Fixed (in the other one)!

PR #2374, opened 2024-05-16
Update Raspios image for mender-convert.

kacf:


merlin-northern (approved):
I checked that the images exist.

kacf:
Thanks! And I've tested that they work! 🙂

PR #2375, opened 2024-05-16
Update versions for mender-convert 4.2.1 release.

kacf:


merlin-northern (approved):
I checked the links and commands, all work, read the comment which made me dizzy, looked at the versions and rpi corresponding images.

PR #2380, opened 2024-05-28
CE-390: Amend `boot` partition detection for latest Raspberry Pi OS image

lluiscampos:
The snippet now will find either `boot` or `bootfs` mount point.

### DIFF
--- a/01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md
+++ b/01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md
@@ -67,13 +67,13 @@ default location:
 [ui-tabs position="top-left" active="0" theme="lite" ]
 [ui-tab title="Linux"]
 ```bash
-RPI_BOOT="/media/$(whoami)/boot"
+RPI_BOOT="$(find /media/$(whoami)/ -name boot -or -name bootfs)"
 [ ! -d "$RPI_BOOT" ] && echo "ERROR: RPI boot directory not found"
 ```
 [/ui-tab]
 [ui-tab title="Mac OS"]
 ```bash
-RPI_BOOT="/Volumes/boot"
+RPI_BOOT="$(find /Volumes/ -name boot -or -name bootfs)"
 [ ! -d "$RPI_BOOT" ] && echo "ERROR: RPI boot directory not found"
 ```
 [/ui-tab]


01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md
```diff
@@ -67,13 +67,13 @@ default location:
...
 ```bash
-RPI_BOOT="/media/$(whoami)/boot"
+RPI_BOOT="$(find /media/$(whoami)/ -mindepth 1 -maxdepth 1  | sed -n 's|\(.*boot\(fs\)\{0,1\}\)$|\1|p');"
 [ ! -d "$RPI_BOOT" ] && echo "ERROR: RPI boot directory not found"
 ```
 [/ui-tab]
 [ui-tab title="Mac OS"]
 ```bash
-RPI_BOOT="/Volumes/boot"
+RPI_BOOT="$(find /Volumes/ -mindepth 1 -maxdepth 1  | sed -n 's|\(.*boot\(fs\)\{0,1\}\)$|\1|p');"
```
lluiscampos, line 76: @merlin-northern Can you test this locally on your Mac OS? :crossed_fingers:

TheMeaningfulEngineer: Tested on Linux:

```
$ RPI_BOOT="$(find /media/$(whoami)/ -mindepth 1 -maxdepth 1  | sed -n 's|\(.*boot\(fs\)\{0,1\}\)$|\1|p');"

$ [ ! -d "$RPI_BOOT" ] && echo "ERRO
R: RPI boot directory not found"
ERROR: RPI boot directory not found

$ ls /media/$(whoami)
90a5ff5d-c3ef-4fbc-8efa-611ae498fca4   bootfs
ad862a08-31f1-42ad-b6c3-2c5c77b157c9   f382b36d-642a-492e-907d-535f09ac2ebe
ad862a08-31f1-42ad-b6c3-2c5c77b157c91
```

Didn't go deeper

merlin-northern: I ran the command just like that, it does not print an error. I have a question: is there a reason why you cant use `-regextype` and `-regex` to find?

merlin-northern: and what if there are two boot and bootfs? wont your script return both?

lluiscampos: > I ran the command just like that, it does not print an error. I have a question: is there a reason why you cant use `-regextype` and `-regex` to find?

I didn't know about these flags

> and if it is always boot or bootfs why we cant do `\( -or -name -boot -or -name bootfs \)` ?

I am always concern of adding flags to commands that should run both on Linux and on Mac OS, so I will go for this solution which is explicit and simple

> and what if there are two boot and bootfs? wont your script return both?

Yes, it will return both and the check below will fail. I would argue that that is an improvement, as having two mount points should be a red flag for the rest of the tutorial not to continue. (Previously if you were to have the two mount points the tutorial would continue but that could lead to unexpected behavior)

merlin-northern: for portability, assuming you have bash:
```bash
boots=(./Volumes/boot*)
RPI_BOOT=${boots[0]}
```

merlin-northern (approved):
looks good.

01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md
lluiscampos: > ```shell
> RPI_BOOT="${boots[0]}" # heads up on no glob matches it will be `"/Volumes/boot*"`
> ```

I experimented with this before but discarded the idea for this very reason :+1:

lluiscampos:
* 3.7.x
* 3.6.x

01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md
merlin-northern: psst: `shopt -s nullglob`

PR #2389, opened 2024-06-03
chore: Update end of support date for Mender 3.7

lluiscampos:
An oversight in the previous commit

### DIFF
--- a/302.Release-information/01.Release-schedule/docs.md
+++ b/302.Release-information/01.Release-schedule/docs.md
@@ -42,7 +42,7 @@ At this time, we support the following LTS releases: 3.6, 3.7.
 
 | LTS         | Supported until |
 | ----------- | --------------- |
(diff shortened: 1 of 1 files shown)


estenberg (approved):

estenberg:
Nice, thank you.

One last branch missing I think: hosted - which is the default.

Small change with lots of pieces to it :)

lluiscampos:
> Nice, thank you.
> 
> One last branch missing I think: hosted - which is the default.
> 
> Small change with lots of pieces to it :)

The regular process for `hosted` is to merge (not cherry-pick) as they need to be aligned once on a while. But because I run into other unexpected problems (ref: Slack private thread) then is best that we do cherry-pick for `hosted` too. Done in https://github.com/mendersoftware/mender-docs/pull/2392

PR #2411, opened 2024-07-02
MEN-7264: Documentation for mender-convert installation using mender-convert

danielskinstad:
Documentation for installing mender-monitor with mender-convert


lluiscampos (approved):
The PR title reads like a riddle.

But other than that the text is good! If we merge it now you can also show it in the upcoming Sprint review :popcorn:

PR #2425, opened 2024-08-02
test: Update expected number of packages after adding new distro

lluiscampos:
Missing from 84600922dd9fe13abd92a8795f4042219a28e3ce.

This test only runs on protected branches (master) so the mistake went unnoticed in the PR.


lluiscampos:
The failure can be seen in [the master pipeline](https://gitlab.com/Northern.tech/Mender/mender-docs/-/jobs/7495099564),

tranchitella (approved):

PR #2435, opened 2024-08-13
docs: Replace OS codenames with numbers

lluiscampos:


mzedel (approved):
I approve this ~message~ change!

nickanderson:
I always struggle to recall what number a codename is related to, so this would be a +1 in my book. I might however suggest having both, e.g. `Debian 12 (Bookworm)`

lluiscampos:
> I always struggle to recall what number a codename is related to, so this would be a +1 in my book. I might however suggest having both, e.g. `Debian 12 (Bookworm)`

Thanks for your feedback @nickanderson.

There is however an extra hidden motivation in doing this change: to have our tabs render in one line! See how they look right now:

image

So having both the codename and the number would defeat this.

nickanderson:
> So having both the codename and the number would defeat this.

Screen space is the toughest final boss

lluiscampos:
* hosted
* 3.7.x

PR #2439, opened 2024-08-20
Docker compose revamp

TheMeaningfulEngineer:
CE-429

Currently blocked by: https://northerntech.atlassian.net/browse/MEN-7477


01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+
+
+In this tutorial, we will conduct a container update to a device using a
+multi-container Docker composition with a Traefik gateway and a backend service
+`whoami`. First we will create the update Artifact with a new Docker composition
+and deploy it to the device. In the second part will show how update the Docker
+images in the composition using binary delta to save size and data transfer
+bandwidth.
+
+In the example, we'll observe Traefik as the primary container in the composition.
```
estenberg, line 20: we will instead of we'll (I think we spell those out)

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+### Prerequisites
+
+To get started using the Docker Compose Update Module, we need to prepare the
+target devices for accepting the deployment and the workstation for creating the
+deployment.
+
+The tutorial assumes the usage of the Raspberry Pi with the reference OS image.
+Please read [this section](../../../../01.Get-started/01.Preparation/01.Prepare-a-Raspberry-Pi-device/docs.md) of the Getting started to prepare the device.
+
+Throughout the tutorial you'll be expected to execute commands on the device.
```
estenberg, line 35: you will

estenberg: you will be expected is a bit long-weinded

maybe something like "you will execute"

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+You can use the [Remote terminal](../../../../09.Add-ons/50.Remote-Terminal/docs.md) to gain access to the shell of the device and execute the command.
+
+
+##### Installing dependencies on the devices
+
+Before installing the Update Module on the device, you need to ensure that the following
```
estenberg, line 41: remove "you need to"

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+
+<!--AUTOVERSION: "app-update-module/blob/%/"/ignore-->
+
+!!! The Docker Compose Update Module is a sub-module of the
+!!! [Application Update Module](https://github.com/mendersoftware/app-update-module/blob/master/docs/README-submodule-api.md#applications-updates).
+!!! That is why we install two Update Modules in the snippet above.
+
+##### Prepare the workstation
+
+To generate the update artifacts we'll need to install the `Application Update Artifact Generator`.
```
estenberg, line 90: To generate Mender Artifacts supporting docker-compose, we install the ...

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+```
+
+<!--AUTOVERSION: "app-update-module/blob/%/"/ignore-->
+
+!!! The Docker Compose Update Module is a sub-module of the
+!!! [Application Update Module](https://github.com/mendersoftware/app-update-module/blob/master/docs/README-submodule-api.md#applications-updates).
+!!! That is why we install two Update Modules in the snippet above.
+
+##### Prepare the workstation
+
```
estenberg, line 89: Make it even more explicit that we are changing place here, e.g.:

Now open a terminal on your **workstation**, which will be used to package the software for your device.

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+# Simple tool to generate Mender Artifact suitable for App Update Module
+# 
+# Usage: ./app-gen [options] [-- [options-for-mender-artifact] ]
+# ...
+```
+
+!!! When integrating into the CI/CD pipelines, you might consider adding the `app-gen` to `PATH`. 
+
+
+##### Workstation variables
```
estenberg, line 133: Change this to a note, not a separate header, as it doesn't have any actionable steps. E.g. with !!!

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+### Create a deployment
+
+The `app-gen` script we installed previously extends the
+[mender-artifact](../../../10.Downloads/docs.md#mender-artifact) tool to create
+Artifacts for container updates. We will use this tool to create a Mender
+Artifact containing the Docker Compose manifest and the Docker images used by
+the composition. Including the Docker images is optional, excluding the images
+will make the devices try to pull the images from the Docker registry.
+
+#### Create the Mender artifact
```
estenberg, line 152: Artifact

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+EOF
+```
+
+To generate the artifact, we need to know the target platform of the devices we
+want to deploy to. 
+In the following example, where we assume the device to be the Raspberry Pi,  we are deploying to is `linux/arm/v7` (`os/arch/variant`).
+
+You can check more details regarding this notation in [Multi-platform images](https://docs.docker.com/build/building/multi-platform/) and [Architectures other than amd64](https://github.com/docker-library/official-images#architectures-other-than-amd64).
+
+Run the command below to generate the update artifact:
```
estenberg, line 190: update artifact -> Artifact

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+        -- \
+        --software-name="$ARTIFACT_NAME" \
+        --software-version="$VERSION"
+```
+
+!!! For Raspberry pi 3 use  `DEVICE_TYPE="raspberrypi3"`
+!!! All arguments after `--` are passed directly to `mender-artifact write module-image`. 
+
+
+The generated artifact (`.mender` extension) is now ready to be deployed on the device.
```
estenberg, line 215: Artifact

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+Open your browser and navigate to the "Releases" column in the [Mender UI](https://hosted.mender.io/ui/releases) and upload your newly created artifact. 
+Once uploaded, navigate to the "Devices" column and select your device and then click `Create a deployment for this device` in the `Device actions` in the bottom right corner. 
+Select your newly created artifact and click `CREATE DEPLOYMENT`.
+
+##### Verify your deployment
+
+Once deployed, the device will start serving a simple server on port 8080. 
+You can test the application by sending a request to path `/whoami` and the server will echo the request.
+
+Open the Remote terminal from the Hosted Mender UI and run the following:
```
estenberg, line 229: Where is Remote terminal? I would have one sentence where to find it, e.g. In hosted Mender, under UI, open the Devices tab and click on your device. Then click "Remote terminal" and run the following commands:

estenberg: Just as we're assuming that the reader is completely new to Mender if they're following Get started.

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+
+
+# Output
+#  CONTAINER ID   IMAGE            COMMAND                  CREATED         STATUS         PORTS                                   NAMES
+#  3a27e70761b8   traefik:v2.9     "/entrypoint.sh --pr…"   9 minutes ago   Up 9 minutes   0.0.0.0:8080->80/tcp, :::8080->80/tcp   traefik-composition-gateway-1
+#  01ca7e0e2f93   traefik/whoami   "/whoami"                9 minutes ago   Up 9 minutes   80/tcp                                  traefik-composition-whoami-1
+```
+
+If you get the similar output as above it shows that the containers on the device are running the correct versions.
+Command below will test for functionality:
```
estenberg, line 242: Run the following command to verify it works:

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+
+#### Update your composition using delta
+
+! This section requires [xdelta3](https://github.com/jmacd/xdelta) to be
+! installed on both your workstation and your device. Please make sure that
+! this dependency is installed before proceeding.
+
+Now that your Docker composition is running on the device, it is time to upgrade
+the Traefik container to the next version. Create a new manifest directory and
+bump gateway service to `traefik:v2.10`:
```
estenberg, line 274: bump -> update the

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+      - "traefik.enable=true"
+      - "traefik.http.routers.whoami.rule=Path(\`/whoami\`)"
+      - "traefik.http.routers.whoami.entrypoints=web"
+EOF
+```
+
+Since we are only upgrading the Traefik service, we do not need to include the
+image for the `whoami` service in the command below.
+
+Proceed with creating the artifact:
```
estenberg, line 307: Artifact

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+And compare the difference:
+
+``` bash
+du -h NO-DELTA-traefik-composition-v2.10.mender traefik-composition-v2.10.mender 
+# Output: 
+# 38M     NO-DELTA-traefik-composition-v2.10.mender
+# 28M     traefik-composition-v2.10.mender
+```
+
+So in this case we're seeing a ~28% decrease in size.
```
estenberg, line 400: size, though the size savings of deltas vary depending on the changes between Artifact as well as parameters used and can go up to 90% in some cases.

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+# 6cacafb1e9d3   traefik/whoami   "/whoami"                58 seconds ago   Up 55 seconds   80/tcp                                  traefik-composition-whoami-1
+```
+
+If you get the similar output as above it shows that the containers have been updated to the new versions. 
+
+
+### Extra: what are the delta benefits?
+
+
+The motivation for using the delta updates is to have a smaller artifact for the same effect.
```
estenberg, line 364: Artifact

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+# Output 
+# CONTAINER ID   IMAGE            COMMAND                  CREATED          STATUS          PORTS                                   NAMES
+# 0a81f8f74b30   traefik:v2.10    "/entrypoint.sh --pr…"   58 seconds ago   Up 55 seconds   0.0.0.0:8080->80/tcp, :::8080->80/tcp   traefik-composition-gateway-1
+# 6cacafb1e9d3   traefik/whoami   "/whoami"                58 seconds ago   Up 55 seconds   80/tcp                                  traefik-composition-whoami-1
+```
+
+If you get the similar output as above it shows that the containers have been updated to the new versions. 
+
+
+### Extra: what are the delta benefits?
```
estenberg, line 361: Benefits of Delta Updates

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
```diff
@@ -0,0 +1,400 @@
...
+        --image docker.io/traefik/whoami:latest \
+        --orchestrator docker-compose \
+        --manifests-dir $MANIFEST_DIR \
+        --output-path "$ARTIFACT_NAME-$VERSION".mender \
+        -- \
+        --software-name="$ARTIFACT_NAME" \
+        --software-version="$VERSION"
+```
+
+!!! For Raspberry pi 3 use  `DEVICE_TYPE="raspberrypi3"`
```
estenberg, line 211: It's possible to supply --device-type twice to avoid this issue, it seems: https://github.com/mendersoftware/app-update-module/blob/master/gen/app-gen#L33. Mender Artifacts support a list of device types compatible.

So if --device-type raspberrypi3 --device-type raspberrypi4 is supplied this should work in both cases.

In this tutorial we can remove the DEVICE_TYPE and PLATFORM variables and have them inline in the command because we are explicitly targetting Raspberry Pi only here (otherwise supplying device-type multiple times may become difficult?).

estenberg (changes requested):
Thanks for updating this, it's overdue. :)

Have you seen that we already have a tutorial here: https://docs.mender.io/artifact-creation/create-an-artifact/docker-compose?

I think that one is mis-placed, and we should remove it after you add yours to Get started. 

There is also another one on Kubernetes I think we should remove: https://docs.mender.io/artifact-creation/create-an-artifact/kubernetes. Perhaps you can just add a note that it's possible to use Kubernetes too at the end of your tutorial instead?

estenberg:
Sorry to expand the scope here so much, but we should also consider the Mender Hub Update Module section:
* https://hub.mender.io/c/update-modules/13?ascending=false&order=views

In particular there is a "competing" Kubernetes implementation that we made: https://hub.mender.io/t/kubernetes/1939

The problem is that it will confuse customers if we provide multiple Update Modules for the same update type.

estenberg:
@kjaskiewiczz FYI, ref the QA discussion.

My opinion is that we need to move the app-update module to https://github.com/mendersoftware/mender-update-modules, add a Mender Hub Update Module tutorial (maybe one for Docker-Compose and another for Kubernetes).

I don't understand why the app-update module was done separate? Was this a more "strategic" decision to split out the update modules we support or was it just done oneoff?

kjaskiewiczz:
Thanks for the info.

> I don't understand why the app-update module was done separate? Was this a more "strategic" decision to split out the update modules we support or was it just done oneoff?

I don't know, but maybe @merlin-northern knows something about it?

merlin-northern:
> Thanks for the info.
> 
> > I don't understand why the app-update module was done separate? Was this a more "strategic" decision to split out the update modules we support or was it just done oneoff?
> 
> I don't know, but maybe @merlin-northern knows something about it?

who me? let's face it my knowledge is a bit on a low level.
so it was a decision that was made, to keep it separate:
1. app-update-module is really an answer to the Mender for Applications, implemented as an update module, true, but we had in mind the release and it had been decided to release from a separate repo
2. it goes a bit beyond the update modules as it introduces so-called submodules
3. the mender-update-modules has or had slightly different idea

I can dig further. if you want to move it to the other repo we can discuss.

TheMeaningfulEngineer:
@estenberg 

Thanks for the review, will go through it.

> Have you seen that we already have a tutorial here: https://docs.mender.io/artifact-creation/create-an-artifact/docker-compose?
> I think that one is mis-placed, and we should remove it after you add yours to Get started.


I've used that one as the base and describe the high level changes in the commit messages.
The current PR changes are doing the move.

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
TheMeaningfulEngineer: Adjusted the case for all occurrences of lower case.

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
TheMeaningfulEngineer: This was an approach I dropped and didn't delete it.
I removed it completly.

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
TheMeaningfulEngineer: Good call.
If QEMU ever becomes Debian based we can revisit.
Will change and retest.

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
TheMeaningfulEngineer: Will expand the text but want do go with:

```
So in this case we are seeing a ~28% decrease in size, though the size savings of deltas vary depending on the changes between Artifacts and can go up to 90% in some cases.
```

Not clear what parameters are you referring to.
If it's the [delta parameters](https://docs.mender.io/artifact-creation/create-a-delta-update-artifact#generating-the-delta-custom-compression-values-special-use-case) I don't know a way to customize these for this update type.

01.Get-started/04.Deploy-a-docker-compose-update/docs.md
TheMeaningfulEngineer: I've added that to the beginning of the document with the link to the Remote terminal section where all the parameters are.
I've expanded the link to every mention of the remote terminal.
Would prefer this approach than to start superficially explaining the Remote terminal here. Fair? :)

estenberg (approved):

oldgiova:
The app-update-module `1.1.0` has been released. The links has been updated already

PR #2441, opened 2024-08-22
[feat]: document the key rotation

TheMeaningfulEngineer:
I've documented the key rotation as part of testing it. Couldn't get the evaluation to be copy-paste so fixed that as well.


MuchoLucho (approved):
It worked on my computer after copy+pasting all the commands

PR #2472, opened 2024-10-16
QA-712: Restructure docs for a flat content structure

lluiscampos:

PR #2473, opened 2024-10-17
QA-712: Uppercase for Mender Mender Client [Server, Gateway]

lluiscampos:
Follow-up from #2472


estenberg (approved):
Quite a few :sweat_smile: 

I skimmed through all and looks good. Thanks for taking care of this!

PR #2479, opened 2024-10-23
Feature: Orchestrator docs

kacf:


09.Add-ons/05.Orchestrator/01.Installation/docs.md
```diff
@@ -0,0 +1,24 @@
...
+
+1. Compile mender-update-orchestrator.
+
+2. When preparing the `orch-install` folder in the Example section, drop the
+   `mender-update-orchestrator` binary into the `orch-install/bin/` folder.
+
+3. When asked to execute `manual_install.sh`, execute it like this:
+   ```
+   ORCH_CUSTOM_INSTALL=1 orch-install/manual_install.sh
+   ```
```
kacf, line 23: These should be the only unofficial steps while we wait for the binary package to be ready.

09.Add-ons/05.Orchestrator/02.Examples/docs.md
```diff
@@ -0,0 +1,424 @@
...
+
+The Orchestrator will merge the two manifests and update the system as a whole.
+
+Remove the configuration and reinstall the original manifest to return the setup to using one
+manifest only:
+
+```bash
+rm -f /etc/mender-update-orchestrator/mender-update-orchestrator.conf
+mender-update-orchestrator  install $MOCK_DIR/system-core-v1/manifest.yaml
+```
```
TheMeaningfulEngineer, line 248: I'd kill these.
They were a useful something until there wasn't an e2e case.
They only server as nose otherwise.

09.Add-ons/05.Orchestrator/docs.md
```diff
@@ -0,0 +1,204 @@
...
+
+For the Orchestrator to be able to update the system, the Manifest and the artifacts for the system components need to be accessible.
+Those two things (Manifest + artifact) are called the Orchestrator bundle.
+
+The Orchestrator is agnostic about the way the bundle is delivered to it.
+For the default case, the Mender client (in daemon mode, communicating with the Mender server) is used to deploy the bundle to the device so that the Orchestrator can act on it.
+
+It is also possible to just pass a Manifest file without the artifacts.
+In that case, nothing is 'bundled', and the Manifest is the sole input.
+The Orchestrator pulls the needed artifacts from the server on its own.
```
TheMeaningfulEngineer, line 204:
```suggestion
### Manifests and artifacts

For the Orchestrator to be able to update the system, it needs both Manifest and the artifacts for the system components.

The Orchestrator is agnostic about the way it gets the required content.

For the default case, the Mender client (in daemon mode, communicating with the Mender server) will download the manifest as part of an update and the Orchstrator will pull the artifacts it needs based on the manifest.

However it is also possible to just deliver both the manifest and artifacts to the Orchestrator in other ways (i.e. an USB stick) and have the system update happen unrelated to the server.
```

The bundle was a thing when we only had the offline flow.
But I don't see at as a concept to highlight that deserves its own term.

09.Add-ons/05.Orchestrator/02.Examples/docs.md
```diff
@@ -0,0 +1,315 @@
...
+MOCK_DIR=/data/mender-update-orchestrator/mock_env
+rm -f $MOCK_DIR/mock_instances/R456/ArtifactCommit.FAIL
+```
+
+Then repeat the steps to start the deployment, and wait for it to finish. It should finish with
+success this time.
+
+Take a look at the updated inventory for the device by clicking the "Devices" tab in the UI, then
+click on the device. Then click on "Software" and "Inventory" tabs to view the installed software
+and the rest of the inventory, respectively.
```
kacf, line 250: @TheMeaningfulEngineer: Just an FYI: It asks you to view the Software and Inventory tabs, but this does not render properly due to [MEN-7640](https://northerntech.atlassian.net/browse/MEN-7640). The docs here are correct though, it's the bug that needs to be fixed.

kacf:
I've tested the instructions, and the current ones work correctly.

As discussed yesterday, I propose we merge now, and you can continue independently on improving these examples.

In addition to any items you might have, these are the TODOs I can think of:

* Rename manifests and artifacts from v3 and v5, to v1 and v2, and make sure they are used in ascending order in the examples.
    * This requires changes in https://github.com/mendersoftware/mender-orchestrator-update-interfaces too.
    * Should update the instructions for making new artifacts there, or even make a script that can autogenerate them.
* Should rearrange example flow so that the success case comes first.
    * This requires some care in which artifacts are used though, since you can't use an already-installed artifact to test a failure case. It will just say it's already installed.
* The things that were commented on [above](https://github.com/mendersoftware/mender-docs/pull/2479#pullrequestreview-2435660659).

kacf:
Got verbal approval from @TheMeaningfulEngineer (who is not here today), so merging.

PR #2487, opened 2024-11-06
fix: Note that unix style line endings are required for state scripts

nickanderson:


06.Artifact-creation/04.State-scripts/docs.md
```diff
@@ -74,6 +74,10 @@ For example, `Download_Enter_05_wifi-driver` and `Download_Enter_10_ask-user` ar
 **There are no arguments passed to the scripts.**
 
 
+## Unix line endings
+
+State scripts must use Unix style line endings. If Windows line endings (CRLF/`\r\n`) are used the state script will not function properly.
```
lluiscampos, line 79: @nickanderson The client does not parse nor do anything special with the contents of the scripts, it just copies them to a location and then execute. So it must be script interpreter (bash or whatever the script has in the shebang) that triggers this issue.

I just run a quick test on my machine and I could execute scripts with CRLF

nickanderson: Alright then, you can reject this. I was just ferrying information from support. There was a case where `ArtifactInstall_Enter`  was not behaving properly and the customer figured out that their script had windows line endings and everything was fixed after converting to unix line endings.

PR #2488, opened 2024-11-13
MEN-7672: Update Deploy a container update tutorial

lluiscampos:
The update module won't be part of the default installation for the next Mender Client release.


estenberg (approved):
Thanks for taking care of this @lluiscampos

PR #2494, opened 2024-11-25
"docs: added comment to run script as root"

victormlg:
- **docs: updated command from mender to mender-update**
- **docs: added comment to run script as root**

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!

### DIFF
--- a/01.Get-started/05.Monitor/docs.md
+++ b/01.Get-started/05.Monitor/docs.md
@@ -188,7 +188,7 @@ Assume you want to monitor the state of a systemd service,
 and you wish to receive _CRITICAL_ alerts if the service is not running,
 and _OK_ alerts when it is back up. 
 
-Create a service that will count some seconds and then fail.
+Create a service that will count some seconds and then fail. You will have to run this script as root:
 
 ```bash
 cat > countdown.sh << "EOF"
@@ -221,6 +221,12 @@ EOF
 
 systemctl daemon-reload
 systemctl start countdown.service
+```
+
+You can for example put the above code in a file called `setup-countdown.sh` and run that with sudo:
+
+```bash
+sudo bash setup-countdown.sh
 ```
 
 You can confirm the countdown with `journalctl -fu countdown.service`.
--- a/01.Get-started/05.Monitor/docs.md
+++ b/01.Get-started/05.Monitor/docs.md
@@ -223,7 +223,7 @@ systemctl daemon-reload
 systemctl start countdown.service
 ```
 
-You can for example put the above code in a file called `setup-countdown.sh` and run that with sudo:
+You can for example put the above code in a file with the name `setup-countdown.sh` and run it with sudo:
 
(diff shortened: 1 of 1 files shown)


lluiscampos (changes requested):
Thanks Victor for taking initiatives in improving the docs :rocket: 

As a general remark, it is a good idea to request for reviews after you create a PR and are happy to present it. This can be done in the top right corner where it says "Reviewers" - in this case I would have pinged your favorite buddy.

When it comes to this PR, please read up our style guide. One thing that we want to avoid as much as possible in Mender Docs is the use of passive voice. Read more [here](https://github.com/mendersoftware/mender-docs/blob/master/README-styleguide.markdown#active-voice-is-preferred-over-passive-voice).

01.Get-started/05.Monitor/docs.md
```diff
@@ -223,6 +223,12 @@ systemctl daemon-reload
 systemctl start countdown.service
 ```
 
+You can for example put the above code in a file called `setup-countdown.sh` and run that with sudo:
```
lluiscampos, line 226: Please rewrite to not use passive voice :+1:

olehermanse: @lluiscampos What do you mean? This is not passive voice.

01.Get-started/05.Monitor/docs.md
```diff
@@ -223,6 +223,12 @@ systemctl daemon-reload
 systemctl start countdown.service
 ```
 
+You can for example put the above code in a file with the name `setup-countdown.sh` and run it with sudo:
```
lluiscampos, line 226: @olehermanse What I meant is something like
```suggestion
In order to run the script privileged, put the above code in a file with the name `setup-countdown.sh` and run it with sudo:
```

So that we avoid the "you can do this", "you may do that".

Our docs owner @estenberg has been vocal about not using these constructions of passive voice.

lluiscampos (approved):
I won't block it further, if you both think that this is not passive voice I accept I am wrong :smiley:

01.Get-started/05.Monitor/docs.md
olehermanse: > So that we avoid the "you can do this", "you may do that".

@lluiscampos Ah, I see. As far as I know, this is not passive voice. Passive voice would be like this:

```
The script below may be run as root (by you).
```

Passive voice means the sentence is turned around so the object comes first and subject comes later ("you can do this" vs "this can be done by you").

Avoiding "you can do", "you may do", is about using shorter, more direct language, in this case with imperative tense (commands - do this, put that, run that).

Nothing wrong with your suggested wording. I think I suggested the wording Victor wrote, because as far as I know, most of our users reading this will know how to run a command / script as root. If not, here is an example of one way it could be done.

PR #2523, opened 2025-01-03
docs: Add a troubleshoot entry for ` mender: command not found` error

danielskinstad:
Ticket: MEN-7831

### DIFF
--- a/301.Troubleshoot/03.Mender-Client/docs.md
+++ b/301.Troubleshoot/03.Mender-Client/docs.md
@@ -4,6 +4,14 @@ taxonomy:
     category: docs
 ---
 
+## `mender: command not found` error when using the `mender-artifact` snapshot feature
+
+This is caused by using an outdated version of the `mender-artifact` tool (e.g from the upstream Debian or
+Ubuntu APT repositories) which cannot detect which version of the Mender Client the device is running.
+
+To solve this issue, install `mender-artifact` from the Mender APT repository following the instructons in
+the [downloads section](../../10.Downloads/docs.md#mender-artifact).
+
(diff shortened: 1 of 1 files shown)


estenberg:
This is a bit more generic entry than the ticket.

In the ticket it is using the "Snapshot" feature of mender-artifact and thus connecting over the network. There is also another very specific error message:

> imgFilesystemType error: imgFilesystemType: blkid command failed: exit status 2

If this error "mender: command not found" is *always* resolved by installing from repositories then this entry is fine.

If not then I would make it more specific, that this happens during snapshot (link to that doc page under Artifact creation) and perhaps include the above error message if relevant.

301.Troubleshoot/03.Mender-Client/docs.md
```diff
@@ -4,6 +4,14 @@ taxonomy:
     category: docs
 ---
 
+## `mender: command not found` error when using `mender-artifact`
```
lluiscampos, line 7: I agree with Eystein, here we must explicitly mention the feature:
```suggestion
## `mender: command not found` error when using `mender-artifact` snapshot feature
```

301.Troubleshoot/03.Mender-Client/docs.md
```diff
@@ -4,6 +4,14 @@ taxonomy:
...
+
+This is caused by using an outdated version of the `mender-artifact` tool (e.g from the upstream Debian or
+Ubuntu APT repositories).
```
lluiscampos, line 10:
```suggestion
Ubuntu APT repositories) which cannot detect which version of the Mender Client the device is running.
```

lluiscampos (approved):

estenberg (approved):

PR #2533, opened 2025-01-17
CE-564: docs: Added note about how to achieve time sync before mender services start

nickanderson:
It's not uncommon for this to be an issue for users. Providing a bit more
specific guidance on how to ensure correct time prior to mender services
starting could help some users avoid running into issues where because of large
polling intervals for production rejections related to time sync can cause long
delays in establishing communication with the server.

Changelog: None
Ticket: CE-564
Signed-off-by: Nick Anderson <nick@cmdln.org>

### DIFF
--- a/301.Troubleshoot/03.Mender-Client/docs.md
+++ b/301.Troubleshoot/03.Mender-Client/docs.md
@@ -204,9 +204,16 @@ To determine the status of your time synchronization, execute the following:
 ```
 
 Note that it can take some time after boot before the time
-synchronization is completed. If after 5-10 minutes, the time still
-has not synchronized, consult with your local network administrator
-for further troubleshooting.
+synchronization is completed.
+This can be avoided by ensuring that time synchronization is complete prior to
+mender services starting. This could be achieved by adjusting mender client
+services like `mender-updated`, `mender-authd`, and `mender-connect` setting
+`Wants` and or `After` to include
+[time-sync.target](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#time-sync.target)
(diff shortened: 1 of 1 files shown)


TheMeaningfulEngineer:
I think this should go into the troubleshooting section instead. 
Also the systemd unit files we install have that ([mender-updated ref](https://github.com/mendersoftware/mender/blob/master/support/mender-updated.service#L3), [mender-authd ref](https://github.com/mendersoftware/mender/blob/master/support/mender-authd.service#L4)).

nickanderson:
> I think this should go into the troubleshooting section instead. 

I can move it there.

> Also the systemd unit files we install have that ([mender-updated ref](https://github.com/mendersoftware/mender/blob/master/support/mender-updated.service#L3), [mender-authd ref](https://github.com/mendersoftware/mender/blob/master/support/mender-authd.service#L4)).

I don't follow. I see that our units do want to order themselves after `systemd-resolved.service` and `network-online.target` but that isn't enough. I came across the special `time-sync.target` https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#time-sync.target
and a user did confirm that in their case to get time sync prior to mender clients starting they:

Enabled `systemd-timesyncd.service` and `systemd-time-wait-sync.service`
and added `Wants=time-sync.target` and `After=time-sync.target` in `mender-updated`, `mender-authd`, `mender-connect`:

nickanderson:
@TheMeaningfulEngineer I moved the content to Troubleshooting, was there more you wanted ?

TheMeaningfulEngineer (approved):

PR #2538, opened 2025-01-22
chore: adjustments to capitalization and taxonomy for multi-tenancy

michaelntech:
Aligned taxonomy with product and marketing – changing "Multi Tenancy" to "multi-tenancy" and keeping "tenant" lower case. "Service Provider" remains capitalized.
Also adjusted some sentence casing of subheadings.

### DIFF
--- a/README-styleguide.markdown
+++ b/README-styleguide.markdown
@@ -61,7 +61,7 @@ Examples:
 > * "helps" rather than "facilitates"
 > * "uses" rather than "utilizes."
 
-## Active voice is preferred over passive voice
+## Prefer active over passive voice
 
 Unless one of the following applies:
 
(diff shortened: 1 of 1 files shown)


02.Overview/16.Taxonomy/docs.md
```diff
@@ -133,10 +133,10 @@ Server. They include management end-points, consumed by users and the UI,
 device end-points, consumed by the Mender components running on the devices,
 and internal end-points.
 
-* _Service Provider Tenant_ - A special _Tenant_ which can create and manage
-children tenants, e.g., different divisions or teams in the same organization
+* _Service Provider tenant_ - A special _tenant_ which can create and manage
```
estenberg, line 136: This one feels a bit more like a name (all caps)?

estenberg:
Looks mostly good, I did some spot checks and we indeed say "tenant" as far as I can see other places too.

Service Provider Tenant - I think we either need to keep it all caps (what I prefer since you could argue it's more of a Mender name - though it's also an industry term not as well known as Multi-tenant) or we say Service provider tenant.

estenberg (approved):

PR #2570, opened 2025-03-09
chore: Add missing `sudo` in sytemctl command

lluiscampos:
The rest of the commands in this tutorial either have `sudo` or an explicit mention to be run as root.


michaelntech (approved):
I have no opinion on whether sudo is needed or not, but the formatting of the change looks good!

lluiscampos:
My opinion is that it should be consistent: this tutorial was using `sudo` in every other command (explicit or implicit).

PR #2571, opened 2025-03-11
MEN-6774: Remove limitation of delta not working on UBIFS

lluiscampos:
Replacing it with an info tip about the version implementing it.

It was implemented one year ago.

### DIFF
--- a/06.Artifact-creation/05.Create-a-Delta-update-Artifact/docs.md
+++ b/06.Artifact-creation/05.Create-a-Delta-update-Artifact/docs.md
@@ -10,7 +10,8 @@ Generation of the delta is something you would execute on a command line of a ma
 
 For a step by step working example please check the [mender hub post](https://hub.mender.io/t/robust-delta-update-rootfs/1144).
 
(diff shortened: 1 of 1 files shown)


estenberg (approved):
Could leave a note that UBIFS is supported since delta-update module XYZ, but if it's a year I think it's not necessary (a year would be about my cutoff for notes like that).

nickanderson (approved):

nickanderson:
My instinct was to approve and merge but i held back in case that messed up your flow

lluiscampos:
> Could leave a note that UBIFS is supported since delta-update module XYZ, but if it's a year I think it's not necessary (a year would be about my cutoff for notes like that).

@estenberg  This is a good remark, thank you. Although all currently supported versions of mender-binary-delta support UBIFS, it is a good practice so I'll add the note

> My instinct was to approve and merge but i held back in case that messed up your flow

On mender-docs, generally would be okey to merge yourself. This time specifically because of Eystein's comment :point_up:  is good that you didn't :)

estenberg (approved):
:+1:

PR #2575, opened 2025-03-18
fix: clarify RetryPollIntervalSeconds/RetryPollCount

TheMeaningfulEngineer:
Ticket:MEN-8151
Changelog: title

### DIFF
--- a/03.Client-installation/07.Configuration/50.Configuration-options/docs.md
+++ b/03.Client-installation/07.Configuration/50.Configuration-options/docs.md
@@ -162,22 +162,53 @@ is sent. The default is 4 hours.
 
 #### RetryPollCount
 
-The maximum number of tries that the Mender Client performs when contacting
-the Mender Server.
+The maximum number of tries that the Mender Client performs when contacting the Mender Server before giving up.
 
-If the setting is zero (the default), the maximum number of retries is `3 * ceil(log2(RetryPollIntervalSeconds) + 1)`.
+It applies to the following Device APIs:
+* [Check Update](https://docs.mender.io/api/#device-api-deployments-v2-check-update)
+* [Inventory reporting APIs](https://docs.mender.io/api/#device-api-device-inventory)
+* [Update Deployment Status](https://docs.mender.io/api/#device-api-deployments-update-deployment-status)
+* [Report Deployment Log](https://docs.mender.io/api/#device-api-deployments-report-deployment-log)
+* [Download Artifact](https://docs.mender.io/api/#device-api-deployments-download-artifact)
+
 
 Introduced in Mender Client 3.3.
 
 #### RetryPollIntervalSeconds
 
-An integer that sets the number of seconds to wait between each attempt to
-communicate with the server. Note that the client may attempt more often
-initially to enable rapid upgrades, but will gradually fall back to this value
-if the server is busy. See also the section about [polling
-intervals](../01.Polling-intervals/docs.md).
+Interval to retry contacting to the server on failed communications. 
 
-As of Mender Client 3.3 this one also applies to inventory updates.
+**Minimum value.** 
+Must be more than 60 seconds. If you specify less than 60, the Mender Client will ignore the setting and use 60 seconds instead.
+
+**Exponential backoff.**
+The specified interval is used for three consecutive attempts. After three failures, the interval doubles for the next three attempts, and so on, until the total number of attempts (as set by [RetryPollCount](#retrypollcount)) is reached.
+
+Example for `RetryPollIntervalSeconds=120` and `RetryPollCount=5`:
+
+```markdown
+try to connect -> fail
+wait 120sec    -> try to connect -> fail
+wait 120sec    -> try to connect -> fail
+wait 120sec    -> try to connect -> fail
+
+wait 240sec    -> try to connect -> fail
+wait 240sec    -> try to connect -> fail
+give up
+```
+
+It applies to the following Device APIs:
+* [Check Update](https://docs.mender.io/api/#device-api-deployments-v2-check-update)
+* [Inventory reporting APIs](https://docs.mender.io/api/#device-api-device-inventory)
+* [Update Deployment Status](https://docs.mender.io/api/#device-api-deployments-update-deployment-status)
+* [Report Deployment Log](https://docs.mender.io/api/#device-api-deployments-report-deployment-log)
+
+
+! [Download Artifact](https://docs.mender.io/api/#device-api-deployments-download-artifact) has a fixed retry set to 60 seconds and can't be configured. 
+! It is **not** affected with the Exponential backoff mechanism, the value remains 60s always.
+
+As of Mender Client 3.3 this configuration option applies to inventory updates as well.
+
 
 #### RootfsPartA
 


03.Client-installation/07.Configuration/50.Configuration-options/docs.md
```diff
@@ -162,22 +162,53 @@ is sent. The default is 4 hours.
...
+
+It applies to the following Device APIs:
+* [Check Update](https://docs.mender.io/api/#device-api-deployments-v2-check-update)
+* [Inventory reporting APIs](https://docs.mender.io/api/#device-api-device-inventory)
+* [Update Deployment Status](https://docs.mender.io/api/#device-api-deployments-update-deployment-status)
+* [Report Deployment Log](https://docs.mender.io/api/#device-api-deployments-report-deployment-log)
+
+
+! [Download Artifact](https://docs.mender.io/api/#device-api-deployments-download-artifact) has a fixed retry set to 60 seconds and can't be configured. 
+! It is **not** affected with the Exponential backoff mechanism, the value remains 60s always.
```
nickanderson, line 208:
```suggestion
! It is **not** affected by the Exponential backoff mechanism, the value remains 60s always.
```

nickanderson (approved):
I don't know if it's true, but it reads pretty well. 

One small wording change suggestion

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
```diff
@@ -162,22 +162,53 @@ is sent. The default is 4 hours.
...
 #### RetryPollCount
 
-The maximum number of tries that the Mender Client performs when contacting
-the Mender Server.
+The maximum number of tries that the Mender Client performs when contacting the Mender Server before giving up.
+
+It applies to the following Device APIs:
+* [Check Update](https://docs.mender.io/api/#device-api-deployments-v2-check-update)
+* [Inventory reporting APIs](https://docs.mender.io/api/#device-api-device-inventory)
+* [Update Deployment Status](https://docs.mender.io/api/#device-api-deployments-update-deployment-status)
```
lluiscampos, line 170: Only for two statues, commit and final success/failure

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
```diff
@@ -162,22 +162,53 @@ is sent. The default is 4 hours.
...
+* [Report Deployment Log](https://docs.mender.io/api/#device-api-deployments-report-deployment-log)
+* [Download Artifact](https://docs.mender.io/api/#device-api-deployments-download-artifact)
 
-If the setting is zero (the default), the maximum number of retries is `3 * ceil(log2(RetryPollIntervalSeconds) + 1)`.
```
lluiscampos, line 168: This may be still true - I haven't verified it. I don't see a reason to remove it in any case, what the default is otherwise?

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
```diff
@@ -162,22 +162,53 @@ is sent. The default is 4 hours.
...
+
+wait 240sec    -> try to connect -> fail
+wait 240sec    -> try to connect -> fail
+give up
+```
+
+It applies to the following Device APIs:
+* [Check Update](https://docs.mender.io/api/#device-api-deployments-v2-check-update)
+* [Inventory reporting APIs](https://docs.mender.io/api/#device-api-device-inventory)
+* [Update Deployment Status](https://docs.mender.io/api/#device-api-deployments-update-deployment-status)
```
lluiscampos, line 203: Same as above, only two statues

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
lluiscampos, line 208: It is yes affected, only that the parameters are harcd-coded to 60 seconds RetryPollIntervalSeconds and 10 times RetryPollCount. But it doubles the same way every three tries.

lluiscampos (approved):
I don't have capacity to look into the magic formula `3 * ceil(log2(RetryPollIntervalSeconds) + 1)`. But I suggest we restore it - by removing it we have no information on what the default is.

Pre-approving, apply suggestions as you see fit :+1:

lluiscampos (changes requested):
> Pre-approving, apply suggestions as you see fit 👍

Please hold on on this PR. I'll reach out privately on JIRA to you guys.

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
```diff
@@ -162,22 +162,53 @@ is sent. The default is 4 hours.
...
+```markdown
+try to connect -> fail
+wait 120sec    -> try to connect -> fail
+wait 120sec    -> try to connect -> fail
+wait 120sec    -> try to connect -> fail
+
+wait 240sec    -> try to connect -> fail
+wait 240sec    -> try to connect -> fail
+give up
+```
```
lluiscampos, line 198:
```suggestion
Maximum interval to retry contacting to the server on failed communications. 

**Minimum value.** 
Must be more than 60 seconds. If you specify less than 60, the Mender Client will ignore the setting and use 60 seconds instead.

**Exponential backoff.**
The algorithm starts with 60 seconds between retries. After three failures, the interval doubles for the next three attempts, and so on, until reaching  `RetryPollIntervalSeconds`. From there, the interval won't be increased further and it will keep retrying until the total number of attempts (as set by [RetryPollCount](#retrypollcount)) is reached.

Example for `RetryPollIntervalSeconds=300` and `RetryPollCount=15`:

```markdown
try to connect -> fail
wait 60sec    -> try to connect -> fail
wait 60sec    -> try to connect -> fail
wait 60sec    -> try to connect -> fail

wait 120sec    -> try to connect -> fail
wait 120sec    -> try to connect -> fail
wait 120sec    -> try to connect -> fail

wait 240sec    -> try to connect -> fail
wait 240sec    -> try to connect -> fail
wait 240sec    -> try to connect -> fail

wait 300sec (max)    -> try to connect -> fail
wait 300sec (max)    -> try to connect -> fail
wait 300sec (max)    -> try to connect -> fail
wait 300sec (max)    -> try to connect -> fail
wait 300sec (max)    -> try to connect -> fail
wait 300sec (max)    -> try to connect -> fail

give up
```
```

**Review by lluiscampos — CHANGES_REQUESTED** [review]
Algorithm reviewed. See new comments :eyes:

File: 03.Client-installation/07.Configuration/50.Configuration-options/docs.md [hunk 6]
```diff
@@ -162,22 +162,65 @@ is sent. The default is 4 hours.
...
+wait 240sec    -> try to connect -> fail
+wait 240sec    -> try to connect -> fail
+wait 240sec    -> try to connect -> fail
+
+wait 300sec (max)    -> try to connect -> fail
+wait 300sec (max)    -> try to connect -> fail
+wait 300sec (max)    -> try to connect -> fail
+wait 300sec (max)    -> try to connect -> fail
+wait 300sec (max)    -> try to connect -> fail
+wait 300sec (max)    -> try to connect -> fail
```
lluiscampos, line 208: Missing \`\`\`

lluiscampos (approved):
Please squash the commits. Also, my comments from [the fmy review](https://github.com/mendersoftware/mender-docs/pull/2575#pullrequestreview-2697021424) still apply.

I am pre-aproving in light that the gist of the algorithm is now correct. Use your judgment to rework based on the original comments.

TheMeaningfulEngineer:
@lluiscampos 
Applied the changes.
Please check the current version of the content.

I'll squash the commits once the content is ok.

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
```diff
@@ -162,22 +162,67 @@ is sent. The default is 4 hours.
...
 
-The maximum number of tries that the Mender Client performs when contacting
-the Mender Server.
+The maximum number of tries that the Mender Client performs when contacting the Mender Server before giving up.
+
+It applies to the following Device APIs:
+* [Check Update](https://docs.mender.io/api/#device-api-deployments-v2-check-update)
+* [Inventory reporting APIs](https://docs.mender.io/api/#device-api-device-inventory)
+* [Update Deployment Status](https://docs.mender.io/api/#device-api-deployments-update-deployment-status)
+  * Only when reporting status "success" and "failure"
```
lluiscampos, line 171:
```suggestion
  * Only when reporting status "commit" and "success" or "failure"
```

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
```diff
@@ -162,22 +162,67 @@ is sent. The default is 4 hours.
...
+wait 300sec (max)    -> try to connect -> fail
+
+give up
+```
+
+It applies to the following Device APIs:
+* [Check Update](https://docs.mender.io/api/#device-api-deployments-v2-check-update)
+* [Inventory reporting APIs](https://docs.mender.io/api/#device-api-device-inventory)
+* [Update Deployment Status](https://docs.mender.io/api/#device-api-deployments-update-deployment-status)
+  * Only when reporting status "success" and "failure"
```
lluiscampos, line 219:
```suggestion
  * Only when reporting status "commit" and "success" or "failure"
```

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
```diff
@@ -162,22 +162,67 @@ is sent. The default is 4 hours.
...
+
+
+! [Download Artifact](https://docs.mender.io/api/#device-api-deployments-download-artifact) has a fixed retry duration of 60 seconds and can't be configured.
```
lluiscampos, line 222: This is dangling here with no context. From the list above there is no mention of Download. Also, this has nothing to do with the Artifact download **API**, that is never retried. And to be clear, this is not configurable so not really related to this docs page at all.

Based on the previous reasoning **I suggest to remove it**. If anything, I would write: "The same algorithm is also applied when resuming an Artifact download: when the connection with the server is lost in the middle of a Download, the client will resume from the last received byte using the same algorithm explained above, but with a fixed retry duration of 60 seconds that cannot be configured"

Also, style:

```suggestion
!!! [Download Artifact](https://docs.mender.io/api/#device-api-deployments-download-artifact) has a fixed retry duration of 60 seconds and can't be configured. 
```

:point_up: It does not need to be a warning, IMO an info note (blue) is enough.

lluiscampos (approved):
I approved last week. But you asked for more review so there you got it :eyes: 

Pre-approving once again and trusting your rework once again.

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
TheMeaningfulEngineer: > Also, this has nothing to do with the Artifact download API, that is never retried. 

Let me see if I got this.

Calling the "API do get the download link from s3" ([GET /artifacts/{id}/download](https://docs.mender.io/api/#device-api-deployments-download-artifact)) has the fixed 60sec duration.


Downloading the artifact from the artifact storage (i.e https://s3.amazonaws.com/hosted-mender-artifacts), which isn't an API call,  has no retries. 

Is this correct?

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
TheMeaningfulEngineer: Does this behavior make sense to you?

My first opinion is this is a bug and not something to document.

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
lluiscampos: > Is this correct?

No. Calling the API has no retries. When downloading, we have a separate feature that we call "download resume" so that I you already downloaded a part of the Artifact and you loose the connection, you resume a bit later (in 60 seconds) using HTTP Range requests to skip the already downloaded data.

So, in other words, it has nothing to do in my opinion with the configuration settings that we are documenting in this page.

03.Client-installation/07.Configuration/50.Configuration-options/docs.md
lluiscampos: In my opinion it is a limitation on the API (because there is no "committing" state) that the client is doing a workaround for. It is a very important step for the Client to try contacting the server just before committing to detect for example aborted deployments that must not be committed. That is why we retry here and treat the failure as a fatal one (note that the status updates that are not retried are not treated as failures, the client goes on).

We shall document that during commit we use this algorithm to submit the status, we don't need to document which exact string are we sending to do so.

PR #2576, opened 2025-03-21
ME-498: docs: Removed documentation for deprecated synchronized updates

nickanderson:
Synchronized updates were removed in Mender Client 4.0 which was released over 1
year ago. With Client 5 out and
5.0.1 on the horizon it makes sense to remove this documentation so as to not
confuse people.

Ticket: ME-498


TheMeaningfulEngineer:
It's the API as well: 
https://docs.mender.io/device-side-api/io.mender.update1

PR #2581, opened 2025-04-08
Fix getting container ID in getting IP of a virtual device

vpodzime:
The image is called `docker.io/mendersoftware/mender-client-docker-addons:latest` so `mender-client-qemu` doesn't match. Let's do `mender-client-`, that should be a good enough filter.


lluiscampos (approved):
Thank you

vpodzime:
> Commit 103ac7bd0ad145c027dd5a0349846cef4d4fdcfb is not signed off! Use --signoff with your commit. Make sure that the Author of the commit matches the one in Signed-off-by

:cry: Looks like I cannot simply make changes in the GH UI/editor.

PR #2591, opened 2025-04-29
APT repo based download instructions improvements

vpodzime:


11.Downloads/docs.md
```diff
@@ -206,95 +206,95 @@ repository. Afterwards, you can install and update the Mender Client using the
...
+    !!! the above command. Do not use the `experimental` repository in production
+    !!! as these releases are not fully tested.
+
+4. Update the package index and install the Mender Client (or some Mender tool, e.g. `mender-artifact`):
+
+    <!--AUTOMATION: ignore -->
+    ```bash
+    sudo apt-get update
+    sudo apt-get install mender-client4
+    ```
```
lluiscampos, line 297: @vpodzime I think I wouldn't mix "Mender Client or Mender tool like mender-artifact" because these two examples are for two different use cases (on a device or on a laptop). If any, I would put other example like mender-setup, mender-connect.

Note that all this is in `## Mender Client` block. If you want to improve the visibility of mender-artifact I then think is best to add a note about the APT repos in the existing mender-artifact section above - or take the APT repos out of Mender Client

vpodzime: I agree. But could we leave that to a follow-up PR? I got to this by following the link to use APT repos that is in the `mender-artifact` installation instructions.

vpodzime: I think we should actually split the whole page in two sections -- _workstation tools_ and _device components_.

lluiscampos: Okey.

lluiscampos (approved):

PR #2598, opened 2025-05-16
fix: indent call outs in device API version section

dhaustein:
Attempts to fix the indentation for call-outs in https://docs.mender.io/release-information/supported-releases#mender-client-subcomponents

image

I am assuming MkDocs is used to generate the documentation and that requires multi-line call-outs to be indented https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage so they are rendered as one piece.


oldgiova:
Hi @dhaustein thank you and great job :+1: ! Could you please squash your commits into a single one, and adjust your signoff commit with the email from your new Github account?

PR #2602, opened 2025-05-28
ME-544: fix: Clarified notes about Personal Access Tokens

nickanderson:
PATs are invalidated on any Mender plan change. Upgrade, downgrade, Add-on addition or removal.

Additionally I reworded to avoid the use of "impersonate" as it has a negative connotation.

Ticket: ME-544

Signed-off-by: Nick Anderson <nick@cmdln.org>


nickanderson:
Cool, I will leave it here for whoever to merge.

PR #2608, opened 2025-06-11
chore: update mender-convert config path

danielskinstad:
The configs were moved, so the config example no longer exists at that path.


vpodzime (approved):
Looks good to me.

PR #2616, opened 2025-06-25
chore: Use the workstation-tools repository for mender-artifact installation

vpodzime:
We want users to prefer the new workstation-tools repository over installation of standalone packages. Currently, it provides `mender-artifact` so we adjust its download/installation instructions accordingly.

The client APT repo instructions were used as a template, but we only have pacakges for a single architecture (amd64).

Ticket: MEN-8431
Changelog: none

### DIFF
--- a/11.Downloads/docs.md
+++ b/11.Downloads/docs.md
@@ -53,28 +53,92 @@ use this utility.
 
 ### Debian/Ubuntu
 
-Follow the correct link according to your host platform to download the
-`mender-artifact` deb package:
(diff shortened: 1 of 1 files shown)


estenberg (approved):
Nice.
Once we get mender-cli here we need to restructure it a bit (rather make a section about setting up the repos and refer to that before you install mender-cli or mender-artifact).

vpodzime:
> Once we get mender-cli here we need to restructure it a bit (rather make a section about setting up the repos and refer to that before you install mender-cli or mender-artifact).

Exactly. :+1:

PR #2621, opened 2025-07-14
chore: Add build from source instruction to mender docs

elkoniu:
This PR adds some generic building instructions for mender related projects together with the repository links.

Ticket: MEN-8557
Changelog: Add build from source instruction to mender docs

### DIFF
--- a/12.General/01.Build-from-source/docs.md
+++ b/12.General/01.Build-from-source/docs.md
@@ -1,0 +1,62 @@
+---
+title: Build from source code
+taxonomy:
+    category: docs
+---
+
+Mender project consist of multiple repositiories with various types of the source code.
+Most of the repositories support `make` and `Makefile` based building / testing /deploying.
+The 2nd most popular build chain used in the project is `CMake`.
+
+Every repository contains a README page with how to run and build instructions.
+It is strongly advised to follow build instructions which can be found
+in the project repositories as they may vary from the general building instructions
+and can contain project specific settings.
+
+## Project repositiories list
+
+| Name                    | Type     | Link                                                      |
+|-------------------------|----------|-----------------------------------------------------------|
+| mender                  | CMake    | https://github.com/mendersoftware/mender                  |
+| mender-artifact         | make     | https://github.com/mendersoftware/mender-artifact         |
+| mender-cli              | make     | https://github.com/mendersoftware/mender-cli              |
+| mender-configure-module | make     | https://github.com/mendersoftware/mender-configure-module |
+| mender-connect          | make     | https://github.com/mendersoftware/mender-connect          |
+| mender-convert          | scripts  | https://github.com/mendersoftware/mender-convert          |
+| mender-dist-packages    | scripts  | https://github.com/mendersoftware/mender-dist-packages    |
+| mender-docs             | scripts  | https://github.com/mendersoftware/mender-docs             |
+| mender-flash            | CMake    | https://github.com/mendersoftware/mender-flash            |
+| mender-mcu              | custom   | https://github.com/mendersoftware/mender-mcu              |
+| mender-mcu-integration  | custom   | https://github.com/mendersoftware/mender-mcu-integration  |
+| mender-qa               | scripts  | https://github.com/mendersoftware/mender-qa               |
+| mender-setup            | make     | https://github.com/mendersoftware/mender-setup            |
+| mender-snapshot         | make     | https://github.com/mendersoftware/mender-snapshot         |
+| mender-test-containers  | scripts  | https://github.com/mendersoftware/mender-test-containers  |
+| mendertesting           | scripts  | https://github.com/mendersoftware/mendertesting           |
+| meta-mender             | scripts  | https://github.com/mendersoftware/meta-mender             |
+
+## make based project general rules
+
+**Warning: Always follow repository README instructions first**
+
+For the `make` based project build and deploiment uses some common target:
+* `make` / `make build` - build the project
+* `make test` - run tests (may contain both: unit tests and integration tests)
+* `make install` - deploy the software on the host machine
+* `make uninstall` - remove the software from the host machine
+
+The easiest way to understand supported targets better is to check
+`Makefile` located in the project directory.
+
+## CMake based projects general rules
+
+**Warning: Always follow repository README instructions first**
+
+For building `CMake` based project:
+* `cmake -D CMAKE_INSTALL_PREFIX:PATH=/usr -B build .` - prepare build directory
+* `cmake --build build` - perform a build
+
+In the case of a `CMake` build some project specific flags may be used (like the one above)
+to change the resulting makefile behavior (eg. force specific install directory).
+
+


12.General/01.Build-from-source/docs.md
```diff
@@ -0,0 +1,58 @@
...
+| mender-cli              | make    | https://github.com/mendersoftware/mender-cli              |
+| mender-configure-module | make    | https://github.com/mendersoftware/mender-configure-module |
+| mender-connect          | make    | https://github.com/mendersoftware/mender-connect          |
+| mender-convert          | scripts | https://github.com/mendersoftware/mender-convert          |
+| mender-dist-packages    | scripts | https://github.com/mendersoftware/mender-dist-packages    |
+| mender-docs             | scripts | https://github.com/mendersoftware/mender-docs             |
+| mender-flash            | CMake   | https://github.com/mendersoftware/mender-flash            |
+| mender-gateway          | make    | https://github.com/mendersoftware/mender-gateway          |
+| mender-mcu              | CMake   | https://github.com/mendersoftware/mender-mcu              |
+| mender-mcu-integration  | CMake   | https://github.com/mendersoftware/mender-mcu-integration  |
```
danielskinstad, line 51: I wouldn't call this `CMake` type. And there is nothing to compile for just mender-mcu-integration unless you have a west workspace with mender-mcu as a module.

For compiling this we use `west`

elkoniu: How about renaming it to `template` then? For just one repo I would avoid to mention `west` and explaining it on the top of the page :)

elkoniu: Renamed to `template`

12.General/01.Build-from-source/docs.md
```diff
@@ -0,0 +1,58 @@
...
+| mender-binary-delta     | make     | https://github.com/mendersoftware/mender-binary-delta     |
+| mender-cli              | make     | https://github.com/mendersoftware/mender-cli              |
+| mender-configure-module | make     | https://github.com/mendersoftware/mender-configure-module |
+| mender-connect          | make     | https://github.com/mendersoftware/mender-connect          |
+| mender-convert          | scripts  | https://github.com/mendersoftware/mender-convert          |
+| mender-dist-packages    | scripts  | https://github.com/mendersoftware/mender-dist-packages    |
+| mender-docs             | scripts  | https://github.com/mendersoftware/mender-docs             |
+| mender-flash            | CMake    | https://github.com/mendersoftware/mender-flash            |
+| mender-gateway          | make     | https://github.com/mendersoftware/mender-gateway          |
+| mender-mcu              | CMake    | https://github.com/mendersoftware/mender-mcu              |
```
danielskinstad, line 50: I still think it's a bit misleading -- especially calling this `CMake`. We build this by having mender-mcu-integration as a zephyr project and this as a `module` and then built together with `west`.

12.General/01.Build-from-source/docs.md
```diff
@@ -0,0 +1,58 @@
...
+The easiest way for the better understanding of supported targets is to check
+`Makefile` located in the project directory.
+
+## CMake based projects
+
+For building `CMake` based project:
+* `mkdir build` - create build directory
+* `cd build` - move to build directory
+* `cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr ..` - generate makefile (with flag in this case)
+* `make` - perform a build based on generated makefile
```
danielskinstad, line 30: For cmake I think commands like
```
cmake -D CMAKE_INSTALL_PREFIX:PATH=/usr -B build .
cmake --build build
```
is better / more straight forward; this way you avoid the mkdir and cd'ing

danielskinstad (commented):
I would think that instead of having a list with potentially over-simplified and vague information, we can just add a note that says something like  `For instructions on how to build from source, see the README in the repositories` instead of having a list. Then we'll also avoid having to maintain that list

elkoniu:
> I would think that instead of having a list with potentially over-simplified and vague information, we can just add a note that says something like `For instructions on how to build from source, see the README in the repositories` instead of having a list. Then we'll also avoid having to maintain that list

I wanted to satisfy those requirements from task:
```
    links to the repositories or git clone commands (+1 to the latter),

    basic build instructions or where to find them (incl. build requirements),

    basic installation instructions or where to find them (making sure the tools end up in $PATH).
```

danielskinstad:
> I wanted to satisfy those requirements from task:

Ah, I didn't look at the task. I don't think it's inherently a bad idea, but it's one more thing to maintain. My point was mainly that not all of them are as easy as just `go to repo and build from source`, so it might not be easy to categorize all of them. E.g. `template` doesn't tell me much, but perhaps  `west` wouldn't have done either since you need a zephyr project bla bla bla (maybe this should link to the mender-docs section?). I don't want to over-complicate things though

Another thing; is there any point in linking to closed source repositories?

12.General/01.Build-from-source/docs.md
```diff
@@ -0,0 +1,65 @@
...
+
+**Warning: Always follow repository README instructions first**
+
+For the `make` based project build and deploiment uses some common target:
+* `make` / `make build` - build the project
+* `make test` - run tests (may contain both: unit tests and integration tests)
+* `make install` - deploy the software on the host machine
+* `make uninstall` - remove the software from the host machine
+
+The easiest way for the better understanding of supported targets is to check
```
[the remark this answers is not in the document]
elkoniu: Fixed

12.General/01.Build-from-source/docs.md
```diff
@@ -0,0 +1,65 @@
...
+    category: docs
+---
+
+Mender project consist of multiple repositiories with various types of the source code.
+Most of the repositories support `make` and `Makefile` based building / testing /deploying.
+The 2nd most popular build chain used in project is `CMake`.
+
+Every repository contains a README page with howto and build instructions.
+It is strongly advised to follow build instructions which can be found
+in the project repositories as they may varies from a general building instructions
```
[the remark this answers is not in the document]
elkoniu: Fixed

12.General/01.Build-from-source/docs.md
```diff
@@ -0,0 +1,65 @@
...
+| mender-binary-delta     | make     | https://github.com/mendersoftware/mender-binary-delta     |
+| mender-cli              | make     | https://github.com/mendersoftware/mender-cli              |
+| mender-configure-module | make     | https://github.com/mendersoftware/mender-configure-module |
+| mender-connect          | make     | https://github.com/mendersoftware/mender-connect          |
+| mender-convert          | scripts  | https://github.com/mendersoftware/mender-convert          |
+| mender-dist-packages    | scripts  | https://github.com/mendersoftware/mender-dist-packages    |
+| mender-docs             | scripts  | https://github.com/mendersoftware/mender-docs             |
+| mender-flash            | CMake    | https://github.com/mendersoftware/mender-flash            |
+| mender-gateway          | make     | https://github.com/mendersoftware/mender-gateway          |
+| mender-mcu              | CMake    | https://github.com/mendersoftware/mender-mcu              |
```
danielskinstad, line 31: You don't have to change this, and I don't think `west` would be more correct for `mender-mcu` either as you can't go to the repo and build it with west, but the reason why I am reluctant to call this `CMake` is that you can't get any working build by following the readme in the repo or by doing the cmake commands as illustrated below; you need a bunch of stuff -- but it's probably fine, I'm just being pedantic :sweat_smile:

danielskinstad (approved):
LGTM :shipit:

12.General/01.Build-from-source/docs.md
elkoniu: I called it `custom` for both MCU variants, and also removed private repos :)

danielskinstad (approved):

PR #2626, opened 2025-07-22
chore: Update qemu device type in get started guide

michalkopczan:
Ticket: None
Changelog: Update qemu device type in Get started guide to be in line with actual device type reported by virtual device

### DIFF
--- a/01.Get-started/02.Deploy-an-application-update/docs.md
+++ b/01.Get-started/02.Deploy-an-application-update/docs.md
@@ -97,7 +97,7 @@ A [Mender Artifact](../../02.Overview/03.Artifact/docs.md) contains some mandato
 
   Depending on which device you are using you need to enter:
   * `raspberrypi3` or `raspberrypi4` if you are using a Raspberry Pi
-  * `qemux64-64` if you are using a virtual device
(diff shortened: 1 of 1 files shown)


michalkopczan:
> Is this correct? Which virtual device is this? My mender-client-qemu device reports `qemux86-64`

Yes, I'm sure it's correct. Also verified right now when testing 5.0.2 release. I assume it is because the virutal device created from "Try a virutal device" when setting up hM is a docker image, and does not run qemu at all.

In other words, when using qemu images it reports `qemux86-64`, but the demo image is not qemu and it reports `genericx86-64`

danielskinstad:
Ah, are we talking about mender-client-docker-addons? Then the change is probably correct, just the title and changelog of the commit are a bit misleading, but I see :+1:

michalkopczan:
> Ah, are we talking about mender-client-docker-addons? Then the change is probably correct, just the title and changelog of the commit are a bit misleading, but I see 👍

You're right, now that you mentioned it, it is misleading. I thought that the docker image from "try a virtual device" uses qemu, hence the title.

I guess we should not force-push to master, so there's no way to fix this now...

danielskinstad:
It's fine :)

PR #2638, opened 2025-08-25
doc: Update path to correct log file

elkoniu:
For log processing we use to use `/var/log/auth.log` file, which is not presented on Debian Bookworm. Instead journalctl shall be used for processing the logs.

Ticket: MEN-8618

### DIFF
--- a/01.Get-started/05.Monitor/docs.md
+++ b/01.Get-started/05.Monitor/docs.md
@@ -266,7 +266,7 @@ First, create a check for the `log` monitoring subsystem named `auth_root_sessio
 the following command:
 
 ```bash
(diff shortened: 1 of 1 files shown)


lluiscampos:
@elkoniu Thanks. Once you have fixed the docs, can you please also update the source code? [Here](https://github.com/mendersoftware/monitor-client/blob/1602efa2e099ebb536503f5fe4b0038dd2d6a694/src/lib/ctl-lib.sh#L64) and [here](https://github.com/mendersoftware/monitor-client/blob/1602efa2e099ebb536503f5fe4b0038dd2d6a694/README.md?plain=1#L93).

elkoniu:
@lluiscampos done: https://github.com/mendersoftware/monitor-client/pull/169

PR #2649, opened 2025-09-11
MEN-8729: Add a note for `mender-convert` next for converting old distros

lluiscampos:


estenberg (approved):

estenberg:
> > Looks good to me. I'm just not sure we want to merge this now when the snippet says `4.3.0` anyway. As a reader I'd find it confusing making me think on of those versions must be wrong.
> 
> I have the same concern, actually. I like to think that `mender-convert` release is around the corner (next sprint if we want!) so we could use this as another factor to get it out ASAP.
> 
> In any way I'll wait for @estenberg opinion as this is very much user facing.

I agree best to wait rather than referring to a future unreleased version. If you have a task for releasing mender-convert, simply add merging this as a AC.

PR #2650, opened 2025-09-16
chore: Add file path needed by CCC release process

elkoniu:


elkoniu:
It turned out file path already exists: 302.Release-information/03.Open-source-licenses/30.mender-artifact/docs.md

PR #2669, opened 2025-10-01
chore: Add Open source licenses for `mender-convert` release

danielskinstad:
As we are releasing it independent of the bundle, we need to add a new chapter for it.

Ticket: QA-1210

### DIFF
--- a/302.Release-information/03.Open-source-licenses/32.mender-convert/docs.md
+++ b/302.Release-information/03.Open-source-licenses/32.mender-convert/docs.md
@@ -1,0 +1,204 @@
+---
+title: mender-convert
+taxonomy:
+    category: docs
+shortcode-core:
+    active: false
+---
+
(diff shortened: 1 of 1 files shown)


302.Release-information/03.Open-source-licenses/31.mender-convert/docs.md
```diff
@@ -0,0 +1,204 @@
+---
+title: mender-convert
+taxonomy:
+    category: docs
+shortcode-core:
+    active: false
+---
+
```
lluiscampos, line 8: I think we should keep the same ordering as for the changelogs. What I mean is that this file should be `32.mender-convert`

lluiscampos (commented):
Looks good. But what do you think about 👇 ?

lluiscampos (approved):

PR #2672, opened 2025-10-02
test: update tests after bumping chapters

danielskinstad:
We added a new chapter, 07.Orchestrate-updates, in faa9eee9c37d9c5019d27accd6a3e7de4f7a69e5 which resulted in the chapters after needing to be increases by one


danielskinstad:
I haven't looked at how the test actually works, but it wasn't caught in the PR pipeline, https://gitlab.com/Northern.tech/Mender/mender-docs/-/jobs/11573655516

elkoniu (approved):
LGTM

PR #2673, opened 2025-10-02
chore: Update monitor-client versions to 1.4.2

danielskinstad:
This should have been done in 71164006a7a4027e1c11af31e252d5044336c19c, but autoversion was run with 1.4.1 rather than 1.4.2

Updated with ./autoversion.py --update --component monitor-client --version 1.4.2


danielskinstad:
Ignore, it is being updated in 2664

PR #2719, opened 2025-12-19
docs: clarify Mender compatibility with non-UEFI U-Boot Secure Boot

nickanderson:
A customer misinterpreted the previous phrasing to mean that Secure Boot
is entirely unsupported when using U-Boot without UEFI.

This update explicitly outlines that while Northern.tech does not
provide pre-validated vendor-specific configurations, Mender is
architecturally compatible with signed U-Boot binaries. This removes
the "not actively supported" disclaimer which was causing confusion
regarding technical feasibility.

Ticket: CE-915


nickanderson:
@estenberg here is a re-phrasing with that last section removed completely. Let me know what you think.

estenberg (approved):

PR #2744, opened 2026-02-04
docs(mender-gateway): clarification regarding intermediate ca

oldgiova:
The intermediate CA could stay at the client side, and it's a suggested approach for easy certificate rotation. This new section should clarify that.

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

The intermediate CA could stay at the client side, and it's a suggested approach for easy certificate rotation. This new section should clarify that.

Thank you!

Co-authored-with: Claude

### DIFF
--- a/10.Server-integration/04.Mender-Gateway/10.Mutual-TLS-authentication/04.Certificate-rotation/docs.md
+++ b/10.Server-integration/04.Mender-Gateway/10.Mutual-TLS-authentication/04.Certificate-rotation/docs.md
@@ -1,0 +1,177 @@
+---
+title: Certificate rotation
+taxonomy:
+    category: docs
+---
+
+When using a flat PKI (`Root CA → Device`), rotating client certificates requires updating the gateway and re-provisioning all devices at once. By using a two-level PKI where each device presents the full certificate chain (device cert + intermediate CA cert), you can rotate the intermediate CA without touching the gateway at all — old and new devices coexist until the rollout is complete.
+
+## How it works
+
+The gateway is configured to trust only the Root CA. Devices send both their device certificate and the intermediate CA that signed it during the TLS handshake. Since the Root CA stays constant, the gateway configuration never changes during a rotation.
+
+## Setting up the PKI
+
+### Generate the Root CA
+
+The Root CA is long-lived and its certificate is the only thing the gateway needs.
+
+```bash
+openssl ecparam -genkey -name P-256 -noout -out root-ca.key
+
+cat > root-ca.conf <<EOF
+[req]
+distinguished_name = req_distinguished_name
+prompt = no
+x509_extensions = v3_ca
+
+[req_distinguished_name]
+commonName=Root CA
+organizationName=My Organization
+
+[v3_ca]
+subjectKeyIdentifier = hash
+authorityKeyIdentifier = keyid:always,issuer
+basicConstraints = critical, CA:true, pathlen:1
+keyUsage = critical, keyCertSign, cRLSign
+EOF
+
+openssl req -new -x509 -key root-ca.key -out root-ca.crt -config root-ca.conf -days $((365*10))
+```
+
+### Generate the Intermediate CA
+
+The intermediate CA signs device certificates and is the unit of rotation.
+
+```bash
+openssl ecparam -genkey -name P-256 -noout -out intermediate-ca.key
+
+cat > intermediate-ca.conf <<EOF
+[req]
+distinguished_name = req_distinguished_name
+prompt = no
+
+[req_distinguished_name]
+commonName=Intermediate CA v1
+organizationName=My Organization
+EOF
+
+cat > intermediate-ca-ext.conf <<EOF
+subjectKeyIdentifier = hash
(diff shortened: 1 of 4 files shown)


nt-alan (changes requested):
AFAIC a table only makes it too complicated to practically digest and is crammed in the wrong section (section tries to solve too many things at once).

I would rather have it here:
https://docs.mender.io/server-integration/mender-gateway/mutual-tls-authentication
as a separate chapter on gateway cert rotations.

And instead of a table a working example.

oldgiova:
> AFAIC a table only makes it too complicated to practically digest and is crammed in the wrong section (section tries to solve too many things at once).
> 
> I would rather have it here: https://docs.mender.io/server-integration/mender-gateway/mutual-tls-authentication as a separate chapter on gateway cert rotations.
> 
> And instead of a table a working example.

Docs refactored

10.Server-integration/04.Mender-Gateway/10.Mutual-TLS-authentication/04.Certificate-rotation/docs.md
```diff
@@ -0,0 +1,176 @@
...
+  -v $(pwd)/server.crt:/etc/mender/certs/server.crt \
+  -v $(pwd)/server.key:/etc/mender/certs/server.key \
+  -v $(pwd)/root-ca.crt:/etc/ssl/certs/root-ca.crt \
+  $MENDER_GATEWAY_IMAGE --log-level debug
+```
+
+## Rotating the intermediate CA
+
+To rotate, generate a new intermediate CA signed by the same Root CA. New devices get certificates signed by the new intermediate CA and present the new chain. Devices already in the field continue working — no gateway change required.
+
```
nt-alan, line 134:
```suggestion
You would rotate the intermediate CA if the old one is about to expire or if you have suspicions it was compromised.  For the expiration case the old certificate will just stop being valid as some point it time, but for the compromised case you need to take explicit action.

!! For case of a compromised certificate you will need to___ **do what?**
```

10.Server-integration/04.Mender-Gateway/10.Mutual-TLS-authentication/04.Certificate-rotation/docs.md
```diff
@@ -0,0 +1,176 @@
...
+```bash
+openssl ecparam -genkey -name P-256 -noout -out device-v2.key
+openssl req -new -key device-v2.key -out device-v2.req -config device.conf
+openssl x509 -req -CA intermediate-ca-v2.crt -CAkey intermediate-ca-v2.key -CAcreateserial \
+  -in device-v2.req -out device-v2.crt -days $((365*2))
+
+cat device-v2.crt intermediate-ca-v2.crt > device-v2-chain.pem
+```
+
+The gateway accepts both `device-chain.pem` (v1) and `device-v2-chain.pem` (v2) because both intermediate CAs chain up to the trusted Root CA. Decommission v1 devices at your own pace.
```
nt-alan:
```suggestion
The gateway accepts both `device-chain.pem` (v1) and `device-v2-chain.pem` (v2) because both intermediate CAs chain up to the trusted Root CA. Update the v1 certificates on the devices at your own pace.
```

Reads weird that you'd be decommission devices just for a cert change.

nt-alan (changes requested):
Thanks, much better readability.

Seems to be a difference between a "expiring intermediate certificate" and "revoking a intermediate certificate".
With the naive setup we had a rotation would solve the revoking part as well, but with this setup the revoked certificate needs some kind of dedicated blackilisting.
That step seems to be missing.

Once we solve that, I think we should just default to this kind of key generation for the default gateway evaluation instead having multiple different ways.

10.Server-integration/04.Mender-Gateway/10.Mutual-TLS-authentication/04.Certificate-rotation/docs.md
oldgiova: I think this is a good call for a dedicated section like: "how to deal with a compromised certiifcate"

oldgiova: Here's the follow up for this: 5045e35d663d3d57832fbd6251494bf66dbac516

oldgiova:
> a rotation would solve the revoking part as well, but with this setup the revoked certificate needs some kind of dedicated blackilisting. That step seems to be missing.

Added to a dedicated section. However please note that to address a compromised cert it's not enough to just change the CA; most of the time you also have to replace the full root CA

nt-alan:
@oldgiova 
Will have to set aside some time to reread this, but afaic this all makes sense.

I want to set the intermediate CE setup as the default case for the mtls eval.
Don't see much value to have the root CE approach for eval and then a dedicated section that has intermediate CE.

Lets see how it goes, feel like we can share a cup of coffee to agree on the high level structure, so whoever ends up working on this first can send the coffee invite :)

PR #2801, opened 2026-04-20
docs: add orchestrator configuration documentation

danielskinstad:
Based on https://docs.mender.io/client-installation/configuration and https://docs.mender.io/client-installation/configuration/configuration-options

### DIFF
--- a/07.Orchestrate-updates/07.Artifacts handling/01.Caching/docs.md
+++ b/07.Orchestrate-updates/07.Artifacts handling/01.Caching/docs.md
@@ -22,7 +22,7 @@ To prevent the cache from growing indefinitely, the Orchestrator performs a clea
 
 ### Storage and Fallback
 
-The filesystem hosting the cache must have sufficient free space to store the full Artifact. However, the caching mechanism is designed to be non-blocking. If the Orchestrator cannot cache an Artifact - for example, if the destination filesystem has reached its capacity - it will skip the caching process. The deployment will continue to function normally, ensuring that storage issues do not cause update failures.
+The filesystem hosting the cache must have sufficient free space to store the full Artifact. If the Orchestrator cannot cache an Artifact - for example, if the destination filesystem has reached its capacity - it will fail the deployment.
 
 ### Configuration
 
@@ -32,12 +32,19 @@ Artifact caching is configured in the Mender Orchestrator configuration file, lo
 {
   "ArtifactsCache": {
     "Enabled": true,
-    "Path": "/var/cache/mender-orchestrator"
+    "Path": "/var/cache/mender-orchestrator",
+    "MaxSizeBytes": 2147483648
   }
 }
 ```
 
 * `Enabled`: Set to true to activate caching. Defaults to false.
 * `Path`: The directory where the cached Artifacts are stored.
+* `MaxSizeBytes`: The maximum size in bytes for the cache. `0` means no limit.
+  Defaults to `0`.
 
-!!! Note: The path configured for the cache must exist on the filesystem before the Orchestrator starts. The service will not attempt to create the directory structure if it is missing. In this case the deployments will work normally but without the cache.
+!!! Note: The path configured for the cache must exist on the filesystem before the Orchestrator starts. The service will not attempt to create the directory structure if it is missing.
+
+For a full description of all cache-related options, see the [ArtifactsCache
+configuration
+reference](../../08.Configuration/50.Configuration-options/docs.md#artifactscache).
--- a/07.Orchestrate-updates/08.Configuration/50.Configuration-options/docs.md
+++ b/07.Orchestrate-updates/08.Configuration/50.Configuration-options/docs.md
@@ -1,0 +1,127 @@
+---
+title: Configuration options
+taxonomy:
+    category: docs
+---
+
+This section lists all the available configuration options in the
+`mender-orchestrator.conf` file.
+
+#### ArtifactVerifyKey
+
+There are two options for specifying verification keys:
+
+* `ArtifactVerifyKey` is a single path to a key.
+* `ArtifactVerifyKeys` is a list of paths to keys. When multiple keys are
+  specified, the keys are tried in order, and the first key that verifies an
+  Artifact signature is used. This is useful for key rotation or signing
+  different types of Artifacts.
+
+Only one of `ArtifactVerifyKey` or `ArtifactVerifyKeys` may be specified.
+
+When set, Mender Orchestrator verifies the following:
+
+* All Artifact installs contain a signature. If a signature is not provided,
+  then the Orchestrator rejects the update.
+* The provided public key verifies the signature of the update.
+
+See also the section about [signing and
+verification](../../../08.Artifact-creation/09.Sign-and-verify/docs.md).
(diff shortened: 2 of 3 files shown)


07.Orchestrate-updates/08.Configuration/50.Configuration-options/docs.md
```diff
@@ -0,0 +1,127 @@
...
+    "Path": "/var/cache/mender-orchestrator",
+    "MaxSizeBytes": 2147483648
+  }
+}
+```
+
+See also the section about [Artifacts
+caching](../../07.Artifacts%20handling/01.Caching/docs.md).
+
+#### HttpsClient
```
danielskinstad, line 69: Note: in the client we also document `SSLEngine` (https://docs.mender.io/client-installation/configuration/configuration-options#httpsclient) and we also have this in the orchestrator code (https://github.com/mendersoftware/mender-orchestrator/blob/aaf86dab0250ea9aa9f2151919824c56325a17bd/src/mender-orchestrator/config_parser/config_parser.cpp#L166). However, engines were deprecated in openssl 3 (it still had legacy support in openssl 3, but now it's been deprecated fully in openssl 4), so do you think it makes sense to have support for engines in mender-orchestrator? IMO we should remove this from the orchestrator unless you see any good reasons to support it (it would be a separate task).

https://openssl-library.org/post/2025-12-18-remove-engines/
https://docs.openssl.org/3.6/man1/openssl-engine/#description

07.Orchestrate-updates/08.Configuration/50.Configuration-options/docs.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### ArtifactsCache
+
+Configures Artifacts caching behavior. When enabled, Mender Orchestrator caches
+downloaded Artifacts locally so that Components sharing the same Artifact do not
+cause redundant downloads.
+
+##### Enabled
+
+Set to `true` to activate caching. Defaults to `false`.
```
danielskinstad, line 39: Also note: this currently default to `true`, but we have plans to update it. Before merging this I will hopefully have that merged in the mender-orchestrator code.

07.Orchestrate-updates/08.Configuration/50.Configuration-options/docs.md
estenberg: I agree, let's focus on the future, until we hit any issue with this.

07.Orchestrate-updates/07.Artifacts handling/01.Caching/docs.md
```diff
@@ -22,7 +22,7 @@ To prevent the cache from growing indefinitely, the Orchestrator performs a clea
 
 ### Storage and Fallback
 
-The filesystem hosting the cache must have sufficient free space to store the full Artifact. However, the caching mechanism is designed to be non-blocking. If the Orchestrator cannot cache an Artifact - for example, if the destination filesystem has reached its capacity - it will skip the caching process. The deployment will continue to function normally, ensuring that storage issues do not cause update failures.
+The filesystem hosting the cache must have sufficient free space to store the full Artifact. If the Orchestrator cannot cache an Artifact - for example, if the destination filesystem has reached its capacity - it will fail the deployment.
```
estenberg, line 25: Which "Artifact" is being referred to here? Is it a given Artifact for a component_type?

So, Orchestrator downloads Artifact for component type 1, then installs it, deletes the Artifact, downloads Artifact for component type 2, etc.?

I thought that the Orchestrator would keep all the Artifacts needed for the whole Deployment, then remove them after the Deployment (unless caching is enabled).

Would clarify this a bit, so users know what storage consumption to expect.

estenberg (approved):
Very well written.

A few clarifications that should be addressed, but I mark it as Approved to avoid so much back-and-forth across timezones.

07.Orchestrate-updates/08.Configuration/50.Configuration-options/docs.md
```diff
@@ -0,0 +1,127 @@
...
+(typically `/var/lib/mender-orchestrator/manifests`).
+
+#### ServerCertificate
+
+The location of the public certificate of the server, if any. If this
+certificate is missing, or the one presented by the server does not match the one
+specified in this setting, the server certificate is validated using standard
+certificate trust chains.
+
+#### TopologyComponents
```
michalkopczan, line 124: Do we have the topology split into topology and 'topology components' explained anywhere? I think we missed that, right? If so, please create a task for documenting in in MEN-7014, thanks

07.Orchestrate-updates/08.Configuration/50.Configuration-options/docs.md
michalkopczan: Agreed, you can add info to remove it to https://northerntech.atlassian.net/browse/MEN-8959

michalkopczan (approved):

07.Orchestrate-updates/08.Configuration/50.Configuration-options/docs.md
danielskinstad: Not that I am aware of. I will :+1:

07.Orchestrate-updates/07.Artifacts handling/01.Caching/docs.md
danielskinstad: > Which "Artifact" is being referred to here? Is it a given Artifact for a component_type?

Yes, the artifact being cached for a given component.

> So, Orchestrator downloads Artifact for component type 1, then installs it, deletes the Artifact, downloads Artifact for component type 2, etc.?

No. We cache and install an aritfact for a component at the same time, and if any subsequent components need the same artifacts they will use it from the cache if available. An Artifact is only deleted from the cache once the artifact is no longer part of the currently installed manifest. It stays in the cache until you deploy a manifest that doesn't require it anymore.

> I thought that the Orchestrator would keep all the Artifacts needed for the whole Deployment, then remove them after the Deployment (unless caching is enabled).

Almost, but as I mentioned, a cached artifact is only deleted when not part of the currently installed manifest, meaning cached artifacts stay even after a deployment has ended. 

See https://docs.mender.io/orchestrate-updates/artifacts%20handling/caching#cache-persistence-and-cleanup

danielskinstad: AFAICT it's documented, but if you think it should be clarified further I can follow up in a separate PR

michalkopczan: Just to add to this: we decided to keep the artifacts in cache after the deployment has ended in order to have the cache usable for self-healing. If someone swapped a physical component, then orchestrator could see that something changed, and install the artifact straight from the cache.

As we decided to not do self-healing some time ago, this is no longer needed, and may be a bit confusing. We  could change it so that the cache is cleared right after the deployment ends.

estenberg: I think this is good to have as-is. Let's not clear it after the deployment.

PR #2802, opened 2026-04-20
chore: Clarify mender-artifact flags and troubleshooting for older versions

nickanderson:
The documentation uses the -c flag (introduced in mender-artifact 4.3.0),
which causes issues for users on older versions (e.g. 3.9.0 from Ubuntu
Noble repos) where -c is not defined.

This commit adds:
- A 'Verify your installation' section to the Workstation tools page
  noting the version requirement for the -c flag.
- Troubleshooting tips to the OS update tutorial to explain both
  the '-c' flag error and the 'mender: not found' error (common
  with version mismatch on Mender 4.0+ devices).


nickanderson:
@nt-alan I pulled this from a recent support case. Thoughts?

nt-alan (dismissed):
I would't give that message a prime location in the docs.
Would rather there's a new troubleshooting section for workstation tools, and let it live there.


Image

nickanderson:
> I would't give that message a prime location in the docs. Would rather there's a new troubleshooting section for workstation tools, and let it live there.

Like this?

nt-alan:
Rendered the PR locally but can't see the menu.
Did it work for you?

image

nickanderson:
> Rendered the PR locally but can't see the menu. Did it work for you?

After I fixed it :D 

image

nickanderson:
Guess it still needs work.

nt-alan (approved):

PR #2805, opened 2026-04-21
feat(mender-gateway): add information about standard proxy environment variables support

kjaskiewiczz:
Ticket: ME-637

### DIFF
--- a/10.Server-integration/04.Mender-Gateway/99.Configuration-file/docs.md
+++ b/10.Server-integration/04.Mender-Gateway/99.Configuration-file/docs.md
@@ -205,3 +205,25 @@ This configuration is available from Mender Gateway version <code>1.3.0</code>.
 <dt>InsecureSkipVerify (<code>UPSTREAM_SERVER_INSECURE_SKIP_VERIFY</code>)</dt>
 <dd>Skip verification of certificate claims.</dd>
 </dl>
+
+</dl>
+
+### Point Mender Gateway to a proxy
+
+<!--AUTOVERSION: "version <code>%</code>"/ignore-->
+!!! Available from Mender Gateway version <code>2.1.0</code>
+
+**Please note** This is a not required for the normal operation of Mender Gateway and is only used for specific use cases.
+
+These configurations are available as environmental variables only.
+
+<dl>
+<dt><code>HTTP_PROXY</code></dt>
+<dd>URL of the proxy server for HTTP connections. Supports authentication using the <code>user:password@host</code> URL format.</dd>
+
+<dt><code>HTTPS_PROXY</code></dt>
+<dd>URL of the proxy server for HTTPS connections. Supports authentication using the <code>user:password@host</code> URL format.</dd>
+
+<dt><code>NO_PROXY</code></dt>
+<dd>Space-separated list of hosts or domain suffixes that should bypass the proxy. Items can be domain suffixes/wildcards, for example <code>.example.com</code> to bypass the proxy for all <code>*.example.com</code> URLs.</dd>
+</dl>


nt-alan (changes requested):
Will make a counter suggestion  in a few min

nt-alan:
image

10.Server-integration/04.Mender-Gateway/99.Configuration-file/docs.md
```diff
@@ -66,6 +66,11 @@ This section provides a reference for the configuration variables.
 Starting from Mender Gateway version <code>1.3.0</code>, configurations can be overwritten using environment variables.
 In the description below, the environment variable names are provided in the parenthesis.
 
+<!--AUTOVERSION: "version <code>%</code>"/ignore-->
+Starting from Mender Gateway version <code>2.1.0</code>, it uses HTTP proxies
+as directed by the environment variables HTTP_PROXY, HTTPS_PROXY and NO_PROXY
+(or the lowercase versions thereof).
+
```
nt-alan:
```suggestion
```

10.Server-integration/04.Mender-Gateway/99.Configuration-file/docs.md
nt-alan, line 207:
```suggestion
</dl>

### Point Mender Gateway to a proxy

!!! <!--AUTOVERSION: "version <code>%</code>"/ignore-->
!!! Available from Mender Gateway version <code>2.1.0</code>

**Please note** This is a not required for the normal operation of Mender Gateway and is only used for specific use cases.

These configurations are available as environmental variables only.

<dl>
<dt><code>HTTP_PROXY</code></dt>
<dd>URL of the proxy server for HTTP connections. Supports authentication using the <code>user:password@host</code> URL format.</dd>

<dt><code>HTTPS_PROXY</code></dt>
<dd>URL of the proxy server for HTTPS connections. Supports authentication using the <code>user:password@host</code> URL format.</dd>

<dt><code>NO_PROXY</code></dt>
<dd>Space-separated list of hosts or domain suffixes that should bypass the proxy. Items can be domain suffixes/wildcards, for example <code>.example.com</code> to bypass the proxy for all <code>*.example.com</code> URLs.</dd>
</dl>
```

nt-alan (changes requested):
I couldn't push so had to squeeze it onto a suggestion.
Couldn't test the render of it this way, but rendered normally on my machine.

1. I've moved it to follow the format of the config consistently

2. The lower case feature think makes no sense, none of the env vars for the gateway are lower case and in the Client description of HTTP_PROXY also no lower case. So would skip that all together.

3. I've copied the basic auth that the client supports to attract attention. If it's not, please explicitly document it's not working, as for the client it does.

kjaskiewiczz:
@nt-alan My sentence was taken directly from golang docs - 
https://pkg.go.dev/net/http#ProxyFromEnvironment
> ProxyFromEnvironment returns the URL of the proxy to use for a given request, as indicated by the environment variables HTTP_PROXY, HTTPS_PROXY and NO_PROXY (or the lowercase versions thereof). Requests use the proxy from the environment variable matching their scheme, unless excluded by NO_PROXY.

We don't need to mention env with lowercase, but this is how it works here and how it works in the client. It's nothing specific to Mender and nothing we implemented (originally we just disable it by mistake both here and in the client). It's just how it works in Linux.

I was trying to NOT emphasize support for these evns, because Mender Gateway can act as proxy for devices and artifact proxy, and we use word `proxy` multiple times in the docs, so I wanted to avoid confusion, especially that I don't think using gateway behind proxy is very common use case (but here I don't have any data, maybe except the fact that it was not supported and there were no complains until now). It's like using proxy behind proxy. I don't want people to think that they need to set these env variables to be able to use Mender Gateway as proxy for their devices.

10.Server-integration/04.Mender-Gateway/99.Configuration-file/docs.md
```diff
@@ -66,6 +66,11 @@ This section provides a reference for the configuration variables.
 Starting from Mender Gateway version <code>1.3.0</code>, configurations can be overwritten using environment variables.
 In the description below, the environment variable names are provided in the parenthesis.
 
+<!--AUTOVERSION: "version <code>%</code>"/ignore-->
+Starting from Mender Gateway version <code>2.1.0</code>, it uses HTTP proxies
```
TheYoctoJester:
```suggestion
Starting from Mender Gateway version <code>2.1.0</code>, it respects an optional server side HTTP proxy
```

10.Server-integration/04.Mender-Gateway/99.Configuration-file/docs.md
```diff
@@ -66,6 +66,11 @@ This section provides a reference for the configuration variables.
...
+as directed by the environment variables HTTP_PROXY, HTTPS_PROXY and NO_PROXY
```
TheYoctoJester:
```suggestion
as provided through the HTTP_PROXY, HTTPS_PROXY and NO_PROXY environment variables.
```

TheYoctoJester (approved):
Minor rewording suggestion.

nt-alan:
> I was trying to NOT emphasize support for these evns, because Mender Gateway can act as proxy for devices and artifact proxy, and we use word proxy multiple times in the docs, so I wanted to avoid confusion, especially that I don't think using gateway behind proxy is very common use case (but here I don't have any data, maybe except the fact that it was not supported and there were no complains until now). It's like using proxy behind proxy. I don't want people to think that they need to set these env variables to be able to use Mender Gateway as proxy for their devices.

I get that, but inventing another documentation style to make it implicit doesn't make it less confusing.
I've added an explicit line clarifying the 'not a standard use case' to the suggestion.


> We don't need to mention env with lowercase, but this is how it works here and how it works in the client. It's nothing specific to Mender and nothing we implemented (originally we just disable it by mistake both here and in the client). It's just how it works in Linux.

Doesn't matter how it's implemented, it's a feature we offer and we support it in a way it's is documented.
From that perspective Mender Client doesn't support the lower case, no reason for Gateway to do the same.
Alternative is fix the client docs as well :)

image

kjaskiewiczz:
@eysteins do you have any opinion on this? If not I'll simply apply changes suggested by @nt-alan

10.Server-integration/04.Mender-Gateway/99.Configuration-file/docs.md
estenberg: "server side" - isn't it a proxy client?

estenberg:
Lots of discussions for a small change here :)

I would try to make it consistent with the Client corresponding doc, as mentioned in the ticket as well.

There doesn't seem to be a clear cut standard if these should be UPPER or lower case. Clearly system env variables are uppercase, but you can argue this one can be application or system. If it can be both then let's document it as both, I don't think this is too big an issue, it's worse not documenting that "feature".

kjaskiewiczz:
Updated. Now it looks like this:
image

10.Server-integration/04.Mender-Gateway/99.Configuration-file/docs.md
TheYoctoJester: @estenberg the key point is, the gateway is connection to a Mender Server, hence its "server side", and that's what the change is for.

TheYoctoJester:
No strong opinions here, so I'll defer to @nt-alan. The one thing which feels weird to me now is that on the client side it's a table, whereas on the gateway it's a continuous text, bullet list thing. Maybe that should be addressed some time, but not worth exploding this PR for that.

nt-alan (approved):

PR #2810, opened 2026-04-28
docs: Fix link anchor

lluiscampos:


lluiscampos:
Hello dear Mender people 😊 

They are probably more links like this one broken - could it be some update in the Grav rendering engine or something that has shortened the anchors?

lluiscampos:
@pasinskim Maybe you can triage this one? The fact that CI is failing indicates that the test is testing the wrong thing - in production site docs.mender.io you can check what are the valid links 😃

nickanderson:
> Hello dear Mender people 😊
> 
> They are probably more links like this one broken - could it be some update in the Grav rendering engine or something that has shortened the anchors?

I am not marcin but let me see what I can see

Are you sure that you don't want to come back?

nickanderson:
The CI job `test:internal-links` uses `oleorhagen/remark-validate-links` (https://github.com/oleorhagen/remark-validate-links/tree/grav-slug), which generates heading slugs via `github-slugger` with no length limit (https://github.com/oleorhagen/remark-validate-links/blob/grav-slug/lib/find/find.js#L68).

Grav CMS,  truncates heading anchors to 64 characters (the default `truncate` setting in the Grav Anchors Plugin (https://github.com/getgrav/grav-plugin-anchors/blob/master/anchors.yaml#L8).

In mender-docs-site we only override `selectors` (https://github.com/mendersoftware/mender-docs-site/blob/master/user/config/plugins/anchors.yaml), so that default value should be in effect.

-   Original slug: `the-bootloader-and-the-linux-kernel-do-not-agree-about-the-indexes-of-storage-devices` is 85 chars
-   Grav-truncated slug (live site): `the-bootloader-and-the-linux-kernel-do-not-agree-about-the-index` is 64 chars

So the PR's new anchor matches the live site, but CI rejects it because CI doesn't apply Grav's truncation.

If we want to keep the truncation then I guess we should look to update @oleorhagen `remark-validate-links` so that it truncates in alignment with Grav.

Or we could whitelist that url In the mender-docs repo https://github.com/mendersoftware/mender-docs/blob/master/.checklinks-whitelist

Or maybe remove grav truncation adding `truncate: false` to https://github.com/mendersoftware/mender-docs-site/blob/master/user/config/plugins/anchors.yaml

I don't have much opinion on which of these is most correct. I would sort of lean to removing truncation, but I don't know if there is good reason to truncate.

lluiscampos:
Thanks for your analysis @nickanderson 

My (external) opinion is that we (you guys) should modify the check to match the production side behavior.

> Are you sure that you don't want to come back?

Not just yet. As you can see I am still using Mender so I haven't got far enough 😆

lluiscampos:
I am closing this PR for tidiness 🧹

PR #2816, opened 2026-05-15
CE-1074: Clarified Mender Client 4 EOL date

nickanderson:
Client 4.0 (released 2024-01) reached end-of-life in 2025-07,
six months after Client 5.0 GA. This was missing from the table,
leaving a gap for customers checking on older clients.

Ticket: CE-1074

### DIFF
--- a/302.Release-information/01.Supported-releases/docs.md
+++ b/302.Release-information/01.Supported-releases/docs.md
@@ -18,6 +18,7 @@ in the table below:
 |-----------------------|---------------|-------------------------|
 | Mender Client 6.0     | 2026-03       | Supported               |
 | Mender Client 5.0     | 2025-01       | Supported               |
+| Mender Client 4.0     | 2024-01       | EOL (since 2025-07)     |
 | Mender MCU Client 1.0 | 2026-04       | Supported               |
 | Mender Gateway 2.0    | 2025-01       | Supported               |
 | Mender Server 4.1     | 2026-01       | Supported               |


302.Release-information/01.Supported-releases/docs.md
MuchoLucho, line 1: Mender client 4.X was part of Bundle 3.7 that already appears as EOL.

I agree this is not obvious, but technically speaking by approving this change we are redundant

MuchoLucho, line 26: For the reference. Here this one means Client 4.X is EOL in line 25

nickanderson: Yes, the non-obviousness of it is what we are trying to reduce.

MuchoLucho: Then I would advocate for changing line 25 as follows:

```diff
- | Mender 3.7 (bundle)   | 2024-05       | EOL (since 2025-05)     |
+ | Mender 3.7 (Mender client 4.X)   | 2024-05       | EOL (since 2025-05)     |
```

or similar

nickanderson: I don't object to that, maybe 
`| Mender 3.7 bundle (Client 4.X, Server 3.7) | 2024-05       | EOL (since 2025-05)     |`

It's Eystein to call it i guess

nickanderson: From internal conversation: still , not great client 4 should eol in month 7 not 5. But also, that means the date for mender 3.7 bundle is just never correct because there are separate dates for its components. If we use the latest date, then the earliest claim support they don't have and if we claim the earlies, then the reverse can be true.

estenberg:
I think I would not do this, it may become more confusing. As mentioned 3.7 covers Mender client 4.0, and this is getting older now.

The problem I see is that Mender Client 4.0 is *not* referring to the same thing as Mender Client 5.0 here. Mender Client is also a "bundle" from 5.0 with several components: https://docs.mender.io/release-information/supported-releases#mender-client

Mender Client 4.0 I assume refers to only mender-update and mender-auth --version.

Unless this is a common question that gets asked I would leave it as-is.

nickanderson:
Well, it's certainly not the first time a customer has been confused about what is or isn't supported.

I don't know how CEs could really track and quantify it.

olehermanse:
https://github.com/mendersoftware/mender-docs/pull/2818

PR #2821, opened 2026-05-19
docs: Clarified in a few more places that the tenant token should be treated as a secret

olehermanse:
Ticket: [MEN-9740](https://northerntech.atlassian.net/browse/MEN-9740)

[MEN-9740]: https://northerntech.atlassian.net/browse/MEN-9740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eystein-ts (approved):

estenberg (approved):
This is OK I think, but I don't think it is "very secret".
Or, at least, it's hard to protect a shared value across all devices, and the consequence for losing it is low (can add pending devices).

Also, expect more customer pressure with documentation in hand for features to rotate, hide it and such after you merge this.

nickanderson:
My mental model is that it's a secret more similar to your home address, email address, or telephone number. Not exactly things you want everyone to know, knowledge of it could result in a lot of spam, but still also not something that is changed frequently or without great effort.

estenberg:
> My mental model is that it's a secret more similar to your home address, email address, or telephone number. Not exactly things you want everyone to know, knowledge of it could result in a lot of spam, but still also not something that is changed frequently or without great effort.

That is a good analogy (spam prevention). :+1: 

Not sure if we should reword/soften this "treat it as a secret", it's just I'm worried that users would take this quite literally, so they would treat it like a private key or password, worry about putting it in secure hardware store etc. But at least it qualifies what it enables (add pending devices).

olehermanse:
> This is OK I think, but I don't think it is "very secret".
Or, at least, it's hard to protect a shared value across all devices, and the consequence for losing it is low (can add pending devices).

> Also, expect more customer pressure with documentation in hand for features to rotate, hide it and such after you merge this.

@estenberg there is nothing new here, this wording is already used in some parts of the docs and in the UI. I'm just clarifying / putting it in the other relevant places so you don't miss it.

PR #2824, opened 2026-05-19
Moved "Regions" docs and deleted "General" chapter and "Build from source" docs

olehermanse:


12.Downloads/03.Build-from-source/docs.md
olehermanse, line 1: Q: The "Build from source" doc is mostly about client stuff, but it also mentions some other repos, like mender-docs. So I could see it fitting into either "Downloads" or "Client installation". Either way, the content itself could probably use a refresh.

nickanderson (approved):
I'm not sure how links and redirects for things that maybe pointed to old urls.

12.Downloads/03.Build-from-source/docs.md
estenberg: I would delete this page:
* It is documented in repos, e.g. https://github.com/mendersoftware/mender
* Now we have two clients
* It's out of date like you say
* Don't think it's suitable for docs (too raw)

estenberg (approved):
Good improvement, I think you can also just delete the build from source.

12.Downloads/03.Build-from-source/docs.md
olehermanse: As you wish :)

olehermanse: @danielskinstad @vpodzime FYI

danielskinstad (approved):

vpodzime (approved):
ACK

PR #2826, opened 2026-05-21
docs: Add Deprecations page and align grace period.

pasinskim:
Introduce `302.Release-information/02.Deprecations/docs.md` as the counterpart to Supported releases, listing deprecated APIs and features with EOL dates (hosted Mender and on-premise) and replacements. Renumber `Release-notes-changelog` to `03` and `Open-source-licenses` to `04` to fit it in.

Keep the Server API deprecation grace period at **12 months** in `02.Overview/15.Compatibility-policy/docs.md` (as agreed), and reframe the on-premise rule as a release-support guarantee: the version of Mender Server containing the deprecated API remains supported for at least the same period from the deprecation announcement, so the effective migration window is the same regardless of deployment model.

Rebased on latest `master`; the new `04.Mender-Gateway` open-source-licenses page added on master is carried into the renumbered `04.Open-source-licenses/` directory.


302.Release-information/02.Deprecations/docs.md
```diff
@@ -0,0 +1,29 @@
...
+[Supported releases](../01.Supported-releases/docs.md).
+
+
+## Deprecated APIs
+
+| API | Version | Deprecated since | EOL (SaaS) | EOL (on-prem) | Replacement |
+|-----|---------|------------------|------------|---------------|-------------|
+|     |         |                  |            |               |             |
+
+On-prem APIs are retired with the next major release. SaaS APIs are retired 3 months
```
estenberg: SaaS -> Hosted Mender (we don't call it SaaS anywhere else)

estenberg (approved):
This is nice and clean starting point - I like it!

You say SaaS several places but this is not in our taxonomy AFIK, would use hosted Mender.

The elephant is the time period. 3 months feels too aggressive to me, did you check other similar services? I am also not sure we are saving any time by having a short time, it's just that we would have more things in the air at the same time, so maybe a bit risky.

estenberg:
@pasinskim ref the timeline, sorry if I have conflicted myself earlier. I did more research on comparable B2B SaaS products, it was hard to find explicitly but here are some data points:

* Twilio - 12 months ("At Twilio, we give users a full year's notice before deprecating any API version", says Evan Cummack, Chief Product Officer at Twilio. - https://optiblack.com/insights/api-versioning-best-practices-2024)

* GitHub - 24 months (https://newsletter.systemdesign.one/i/174692880/2-github)

* Google Campaign manager - 7 months (https://developers.google.com/doubleclick-advertisers/deprecation?hl=en)

It is too short with 3 months, especially given that customers have deep infra integrations with Mender.

I think **6 months** is minimum we should do. 12 months would be slightly better, if it does not cause too much extra work.

Also, I think we should make this only apply to Management APIs. Device APIs we can either leave undefined or state 5 years or something like that.

Finally - on-premise is a bit tricky, because it's not exactly tied to timelines. What we can do here is ensure that the version with the deprecated API is supported for at least the same amount of time (e.g. 6 months) from when we announce deprecated API. So the net effect would be the same as for hosted Mender.

Does that make sense? Want to settle for 6 months for now?

302.Release-information/02.Deprecations/docs.md
pasinskim: fixed!

pasinskim:
@estenberg settled on 6 months minimum for Management APIs as you suggested; wording is "at least 6 months" so we can extend without another doc change. Kept the on-premise framing you proposed (the release containing the deprecated API stays supported for the same window, so customers get the same effective migration time regardless of deployment).

Left Device APIs out of the timeline section. We haven't publicly committed to anything there yet, so silent feels safer than picking a number now and being stuck with it; we can address it separately later.

estenberg (approved):
@pasinskim thanks for taking care of this. The only slight ambiguity is on-prem releases here, and when they'd have time to migrate: 1) do we give them 6 months support *after* we release the next on-prem when that API is gone?, or 2) does the clock starts ticking when we send them the announcement even if we don't have a way for them to migrate?

The latter does not make much sense, because they can do nothing with the deprecation announcement until they have the new version without the api / with the replacement API.

So the "perfect" approach is that we only send them deprecation notice once we release a new on-premise version where the API is deprecated.

But still, since it's on-premise it is not so severe because they can keep it running even after it's unsupported.

So all in all, I think let's merge this and move forward.

kjaskiewiczz (approved):

estenberg (approved):

PR #2836, opened 2026-06-16
fix: Do not refer to Releases tab/section

vpodzime:
The tab is now called _Software_.

Ticket: none
Changelog: none

### DIFF
--- a/01.Get-started/02.Deploy-an-application-update/docs.md
+++ b/01.Get-started/02.Deploy-an-application-update/docs.md
@@ -109,7 +109,7 @@ Leave the next prefilled fields as they are and click **UPLOAD ARTIFACT** as bel
 
 ![artifact details 2](artifact_details_2.png)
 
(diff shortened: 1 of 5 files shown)


vpodzime:
I guess this should also go with updated screenshots? What's the procedure for that?

michaelntech:
> I guess this should also go with updated screenshots? What's the procedure for that?

There are some very old screenshots there! I think the design team can take this part; I'm thinking we put them in Figma so they are easily reproducible and update-able for when things change.

estenberg (approved):
Thanks for doing this. These changes are good.

I agree we should take the screenshots at the same time, otherwise I think this might be extra confusing because the text and screenshots don't match.

vpodzime:
> I agree we should take the screenshots at the same time, otherwise I think this might be extra confusing because the text and screenshots don't match.

@michaelntech can you merge this together with the new screenshots?

PR #2837, opened 2026-06-16
chore: Add a warning about temp space when creating a snapshot

vpodzime:
Doing a device snapshot with `mender-artifact` requires a lot of temporary space. On systems where `/tmp` is a tmpfs mount point this can easily result in crashes or OOM situations. Since it's a common setup we should warn users and tell them they can use the new `--tmp` option to tell `mender-artifact` to use a different directory for temporary files.

Ticket: MEN-8479
Changelog: none


estenberg (approved):
Nice improvement, thank you.

rewanrashid-boop:
Looks good to me :rocket: :D

rewanrashid-boop (approved):

PR #2840, opened 2026-06-18
docs: document splitting the Topology into a separate components file

danielskinstad:
Ticket: MEN-9610

### DIFF
--- a/07.Orchestrate-updates/03.Topology/docs.md
+++ b/07.Orchestrate-updates/03.Topology/docs.md
@@ -39,7 +39,9 @@ components:
 | `api_version` | string | ✓ | API version specification |
 | `kind` | string | ✓ | Kind of yaml (should be set to "topology") |
 | `system_type` | string | ✓ | The System type of the System |
-| `components` | list | ✓ | List of all Components in the System |
+| `components` | list | ✓<sup>1</sup> | List of all Components in the System |
+
+<sup>1</sup> Required in this file unless the Components are supplied through a separate Topology components file. See [Splitting the Topology](#splitting-the-topology).
 
 ### Component list
 Each component entry in `components` should have these fields:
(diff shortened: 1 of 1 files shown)


danielskinstad:
@michalkopczan, this should be ready to be reviewed.

07.Orchestrate-updates/03.Topology/docs.md
```diff
@@ -48,4 +50,103 @@ Each component entry in `components` should have these fields:
...
+
+* The main `topology.yaml` contains only the top-level fields and **no**
+  `components` section.
+* A separate file contains **only** the `components` list, using the same
+  Component fields described above.
+
+Point Mender Orchestrator at the separate Components file in one of two ways:
+
+* the `--topology-components` (`-t`) command-line option, or
+* the `MENDER_ORCHESTRATOR_TOPOLOGY_COMPONENTS_FILE` environment variable.
```
michalkopczan: At first, I read it as if you forgot the TopologyComponents config option. And only then noticed that you actually mention it below. I'd say there are THREE ways.

07.Orchestrate-updates/03.Topology/docs.md
```diff
@@ -48,4 +50,103 @@ Each component entry in `components` should have these fields:
...
+To add a Component, list it in the Topology (or the Topology components file) and
+deploy a Manifest that includes it. Mender Orchestrator installs the new
+Component on the next deployment.
+
+Every Component in the Topology must have a matching `component_type` in the deployed
+Manifest, otherwise the deployment fails with a validation error.
+
+### Removing a Component
+
+To remove a Component, delete it from the Topology (or the Topology components
```
michalkopczan, line 147: I think it's worth mentioning that while deploying the new topology that does not include the component, we still need to provide a manifest which actually does include this component - as for this deployment, this component is still part of the System.

This is the part that may be unintuitive, and IMO worth explaining.

It could be good also to mention that in order to not update the component (because it was e.g. already removed physically), the user can assign it an artifact in a manifest, the same that's currently installed - this will result in nothing being installed.
However, if the Interface cannot work with the Component being physically removed, then we have a problem. So we need to document it too - so that the user either makes sure that the Interface returns something if the Component is removed, or that the user makes sure to first deploy a new topology (the one without the component), and only then removes it physically.

michalkopczan (approved):

PR #2841, opened 2026-06-18
docs: Building with demo and commercial together

rewanrashid-boop:
Updated the docs with solution to problem regarding custom mender-gateway.conf in builds with both commercial and demo layers.

Ticket: MEN-6625

### DIFF
--- a/05.Operating-System-updates-Yocto-Project/05.Customize-Mender/docs.md
+++ b/05.Operating-System-updates-Yocto-Project/05.Customize-Mender/docs.md
@@ -421,12 +421,20 @@ augment the `mender-gateway` recipe with the new configuration. For example, cre
 FILESEXTRAPATHS:prepend := "${THISDIR}/<DIRECTORY-WITH-MENDER-GATEWAY-CONF>:"
 SRC_URI:append = " file://mender-gateway.conf"
 
+FILES:${PN} += "${sysconfdir}/mender/mender-gateway.conf"
+
 do_install:append() {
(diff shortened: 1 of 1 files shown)


rewanrashid-boop:
In this  [task (MEN-6625)](https://northerntech.atlassian.net/browse/MEN-6625) the issue really only appears when commercial and demo layer are used together which as i understand only for testing. It happens because the .bbappend in demo's mender-gateway grabs custom files under the name mender-gateway.conf into doc. Therefore i think doing any code change is not really necessary and rather it is better to just update  the docs for this case.

vpodzime (commented):
Looks reasonable to me, but I'm no yocto expert to dare approving this 😊

danielskinstad:
> In this [task (MEN-6625)](https://northerntech.atlassian.net/browse/MEN-6625) the issue really only appears when commercial and demo layer are used together which as i understand only for testing. It happens because the .bbappend in demo's mender-gateway grabs custom files under the name mender-gateway.conf into doc. Therefore i think doing any code change is not really necessary and rather it is better to just update the docs for this case.


IMO this sounds like weird behavior. Any reason why you don't think we should try to address this in the code instead?

rewanrashid-boop:
Yeah that's really what I'm kinda wondering is do still need this thing for examples confs does it have any other purpose so that it would be bad to just remove it. If it is not needed then i can just close this PR open a new one on meta-mender with the line removed if no one has any objections. :shrug:

PR #2842, opened 2026-06-19
chore: Change to new commit linter

rewanrashid-boop:
Replace old commit linter with the new one in mendertesting/templates

Ticket: QA-1561

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -11,8 +11,7 @@ include:
     file:
       - qa-common/runner.yml
       - qa-common/retry.yml
-  - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-check-commits-signoffs.yml'
+  - component: gitlab.com/Northern.tech/Mender/mendertesting/commit-lint@master
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-check-python3-format.yml'
   - project: 'Northern.tech/Mender/mendertesting'


.gitlab-ci.yml
```diff
@@ -75,3 +74,15 @@ trigger:mender-docs-site:production:
   trigger:
     project: Northern.tech/Mender/mender-docs-site
     branch: production
+
+lint:commit:
+  before_script:
+    - npm install --global commitlint-plugin-selective-scope
+    - git branch --move temp-branch || true
+    - git fetch --force origin 'refs/heads/*:refs/heads/*'
+    - |
```
oldgiova: I don't think it's matter that much, but from the upstream template the `- git fetch --tags origin` is missing

.gitlab-ci.yml
```diff
@@ -75,3 +74,15 @@ trigger:mender-docs-site:production:
...
+      if [ ! -f commitlint.config.js ]; then
+        apk add --no-cache curl
+        curl --fail --output commitlint.config.js https://raw.githubusercontent.com/mendersoftware/mendertesting/master/commitlint.config.js
+        sed -i 's/module.exports =/export default/g' commitlint.config.js
```
oldgiova: Why this change with this repository only?

To me this seems
from:
```
module.exports = {
  extends: ['@commitlint/config-conventional'],

```

to: 
```
export default {
  extends: ['@commitlint/config-conventional'],

```
Is this desired?

oldgiova (commented):
I need a little bit more context here ;)

oldgiova (approved):

package.json
```diff
@@ -35,6 +35,5 @@
       "repeated-words",
       "sentence-spacing"
     ]
-  },
-  "type": "module"
+  }
```
danielskinstad: Did you mean to remove this?

rewanrashid-boop: Just testing not going to merge it i wanted to see it that fixed the reference error because when moving from the old commitlinter to the new one in mender-docs it triggered a reference error to do with commitlint.config.js, I think because of the package.json being ES (ECMA Script) while the commitlint.config.js in mendertesting is commonJS.

rewanrashid-boop: Am wondering why do we have this package.json as ecmascript modules, do we have this anywhere else? Another possible fix is changing the commitlint.config.js int mendertesting to commitlint.config.cjs

rewanrashid-boop: @danielskinstad

danielskinstad: Can you show me the error? I have no idea TBH

rewanrashid-boop: https://gitlab.com/Northern.tech/Mender/mender-docs/-/jobs/14941167825 @danielskinstad

danielskinstad: Changing it to .cjs might be fine. Maybe someone more familiar with this has an opinion? @mzedel perhaps

rewanrashid-boop: That is what he suggested as well I'll open a PR in mendertesting to check. If it works merge that and then update this again and see.

mzedel: my suggestion was more for this PR - as this is afaik the only repo this applies to... with the removal of the `"type": "module"` from the package it seems to work here + the `test:internal-links` job still works - so this might be the less impactful/ risky change 👌

rewanrashid-boop: Alright then if there are no objections we can go forward with this change it is the simplest, but i don't really know why only this repo has package.json using ES modules rather than commonjs @mzedel @danielskinstad. If we have other repos using ES modules i think it is better to update commit-lint.yml in mendertesting though.

mzedel: it’s likely the presence of the `package.json` in the repo root (as we're a majority non-js shop, most repos don't have any) + the `module` type + the `.js` extension that triggers this…

PR #2849, opened 2026-07-07
MEN-9907: Test devices definition

mineralsfree:
Add related limits section & mention in device tiers

Ticket: MEN-9907

### DIFF
--- a/02.Overview/08.Inventory/docs.md
+++ b/02.Overview/08.Inventory/docs.md
@@ -87,6 +87,31 @@ autocomplete options. For this reason, it is a good idea to strive to use the
 same naming schemes across the organization, avoid making multiple similar
 names and never putting anything sensitive into the names of tags / inventory.
 
+### Test devices
+
+`test_device` is a reserved _system attribute_ (an attribute in the `system`
+scope, managed by the Mender Server rather than reported by the device) used to
+designate a device as a _test device_. Test devices are exempt from the normal
+update check and inventory poll rate limiting, and are allowed to poll for updates
+and publish inventory frequently. This is useful when trying out Mender features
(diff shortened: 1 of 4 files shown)


estenberg:
Well written.

One thing I'd adjust a bit, is this which is mentioned a few times:

"Test devices are exempt from update check and inventory poll rate limiting, so they receive responsive, near interactive deployments and inventory updates."

I think the "exempt" and "near interactive" is a bit too strong. It also depends on the client configuration. Would rather make it a bit less explicit, e.g. 

"Test devices are exempt from the normal update check and inventory poll rate limiting, and are allowed to poll for updates and publish inventory frequently."

Not sure if we should set an exact limit here, I guess it's like every second. We could set a limit like every 5 seconds, if this becomes too vague.

estenberg (approved):

mineralsfree:
> Well written.
> 
> One thing I'd adjust a bit, is this which is mentioned a few times:
> 
> "Test devices are exempt from update check and inventory poll rate limiting, so they receive responsive, near interactive deployments and inventory updates."
> 
> I think the "exempt" and "near interactive" is a bit too strong. It also depends on the client configuration. Would rather make it a bit less explicit, e.g.
> 
> "Test devices are exempt from the normal update check and inventory poll rate limiting, and are allowed to poll for updates and publish inventory frequently."
> 
> Not sure if we should set an exact limit here, I guess it's like every second. We could set a limit like every 5 seconds, if this becomes too vague.

Softened the phrasing.

As for the limit for test_devices, I think we should not specify the explicit limit; that might push users a bit towards using this limit, which might result in a bigger load IMO. We have a section on excessive use in our HM agreement, so we still can throttle the abusive behavior. 

If you think otherwise, then we could specify it in limits section, lmk

PR #2854, opened 2026-07-10
chore: pin mender-mcu-integration to zephyr-4.2.0

danielskinstad:
We haven't released any support for Zephyr 4.4 and we should start documenting a tagged mender-mcu version rather than using main.


danielskinstad:
AFACIT this should wait until we have a PR in mender-mcu-integration that can pin a supported mender-mcu version rather than main

vpodzime:
> AFACIT this should wait until we have a PR in mender-mcu-integration that can pin a supported mender-mcu version rather than main

Oh, right! That's another "minor" problem! :sweat_smile: But I think this improves things alone/already, I'd merge it.

rewanrashid-boop (approved):
Looks good, why did we have 4.4.0 in docs when we have no 4.4.x branch?

danielskinstad:
> Looks good, why did we have 4.4.0 in docs when we have no 4.4.x branch?

Because it was done by CEs and not us. I didn't revert it until now because documenting 4.2  was technically incorrect. We branched out mender-mcu-integration for this specific purpose (to not use main) and we don't currently have any tagged mender-mcu verions that use zephyr 4.4 -- the support is only in main.

PR #2859, opened 2026-07-14
docs(orchestrator): Add orchestrated ESP32-S3 update example

michalkopczan:
Add a real-hardware Orchestrate-updates example where a Raspberry Pi 5 updates an ESP32-S3 component that does not run Mender. Split the Examples section into a Mock components page and the new ESP32 walkthrough, and fix the inbound links accordingly.


danielskinstad (approved):
Reads well to me. My initial thought was that this could perhaps fit better in mender hub, as mender-docs is a bit more "official", but I don't really have other arguments than "it initially felt that way" 😅 Please add the ticket to the commit as well.

michalkopczan:
@danielskinstad , thanks. I added just now a commit with orchestrator packages bump, can you take a look?

danielskinstad:
LGTM

PR #2862, opened 2026-07-15
docs(orchestrator): document upgrading a device to the system tier

michalkopczan:
Add a guide for converting an existing standard tier device into a system tier device over-the-air, split into three deployments to keep the device recoverable if the tier change rolls back.

Ticket: MEN-9797


danielskinstad (approved):
LGTM

PR #2876, opened 2026-08-17
Delta docker-compose

vpodzime:
**This should be on-hold until mender-delta-docker-compose-1.1.0 is released.**

### DIFF
--- a/08.Artifact-creation/05.Create-a-docker-compose-update-Artifact/docs.md
+++ b/08.Artifact-creation/05.Create-a-docker-compose-update-Artifact/docs.md
@@ -202,3 +202,48 @@ skopeo login registry.example.com
 ```
 
 After authenticating, `gen_docker-compose` will be able to download images from the private registry.
+
+
+## Create a Delta docker-compose update Artifact
+
+When updating devices with a docker-compose Artifact installed, a delta between
+the old (installed) and the new (update) docker-compose Artifacts can be
+generated and packed as an Artifact of type **delta**-docker-compose. Such an
+Artifact can then be deployed as an update instead of the new docker-compose
+artifact in order to save bandwidth.
+
+!!! The delta-docker-compose Update Module and generator are not publicly
+!!! available and are not part of Mender Client.
+
+See the [mender-delta-docker-compose download
+instructions](../../12.Downloads/02.Device-components/docs.md#mender-delta-docker-compose)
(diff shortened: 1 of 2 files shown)


vpodzime:
@estenberg is it OK to just extend the https://docs.mender.io/artifact-creation/create-a-docker-compose-update-artifact page instead of creating a new one? We have a separate page for creation of rootfs-image deltas (mender-binary-delta), but I think the docker-compose page is a better fix for the short section about docker-compose deltas.

08.Artifact-creation/05.Create-a-docker-compose-update-Artifact/docs.md
```diff
@@ -202,3 +202,47 @@ skopeo login registry.example.com
...
 
 After authenticating, `gen_docker-compose` will be able to download images from the private registry.
+
+
+## Create a Delta docker-compose update Artifact
+
+When updating devices with a docker-compose Artifact installed, a delta between
+the old and the new docker-compose Artifacts can be generated and packed as a
+new, **delta**-docker-compose, Artifact that can be deployed instead of the new
+docker-compose artifact saving bandwidth.
```
estenberg: "saving bandwidth" does not append well here.

Would consider splitting this sentence - it is long, or at least you need a comma before "saving bandwidth".

12.Downloads/02.Device-components/docs.md
```diff
@@ -1416,3 +1416,60 @@ sudo cp mender-binary-delta-1.5.3/x86_64/mender-binary-delta-generator /usr/bin
 ```
 
 
+## mender-delta-docker-compose
+
+### Download
+
+If you are using *hosted Mender*, set the following variables with your credentials:
```
estenberg: Would be a bit better if we can handle the hosted/on-premise split with tabs instead, like this: https://docs.mender.io/downloads/device-components#mender-orchestrator

estenberg (approved):
Works, though added some suggestions for making it a bit cleaner.

estenberg:
> @estenberg is it OK to just extend the https://docs.mender.io/artifact-creation/create-a-docker-compose-update-artifact page instead of creating a new one? We have a separate page for creation of rootfs-image deltas (mender-binary-delta), but I think the docker-compose page is a better fix for the short section about docker-compose deltas.

Yes, this works I think.

estenberg (approved):
Eggselent!

PR #2877, opened 2026-08-18
Update images in chapter 2 (Overview) to a consistent style

michaelntech:
### Description

First PR in chapter-by-chapter update of diagram images to have a consistent style.

Where images should fit the whole content width, they are simply provided at 2x the max width size. Where images should be less than the whole content width, @2x versions are provided along with a "size" parameter in the markdown to ensure responsive images are displayed (see [Grav docs](https://learn.getgrav.org/17/content/media#responsive-images)). This is to explain why in some pages the markdown has been slightly adjusted.


michaelntech:
Annoyingly, using responsive images [this way](https://learn.getgrav.org/17/content/media#higher-density-displays) breaks our link checking test in the pipeline.

02.Overview/01.Introduction/application-updates.svg
michalkopczan, line 1: Should we have the arrow touching the File(s) Update module? The intention of this drawing is just showing examples of update modules.

PR #2897, opened 2026-09-15
docs: clarify what the "Force update" option overrides

tactmaster:
### Description

The Deployment overview stated that a device not taking part in the same Deployment twice "can be overridden with the 'Force update' option". That isn't what the flag does.

In the deployments service, `force_installation` has a single use site (`backend/services/deployments/app/app.go`, and the equivalent in the enterprise repo):

```go
if !deployment.ForceInstallation &&
    d.isAlreadyInstalled(request, deviceDeployment) &&
    deviceDeployment.Status == model.DeviceDeploymentStatusPending {
    return nil, d.handleAlreadyInstalled(ctx, deviceDeployment)
}
```

`isAlreadyInstalled` is a comparison of the device's reported `artifact_name` against the assigned Artifact's name, so the flag only disables the already-installed check — which is exactly how the model comment and `management_api.yml` describe it.

The once-per-Deployment rule is enforced elsewhere and is unaffected: `getNewDeploymentForDevice` only considers deployments created strictly after the device's last finished device deployment (`FindNewerActiveDeployment` filters `created: {$gt: ...}`), and `deviceDeployment.Created` is copied from `deployment.Created`.

This PR splits the two rules apart, and documents the Reset Device Deployments history endpoint as the supported way to make a device eligible again for still-active Deployments.

- [x] Commits follow the conventional commit specification
- [x] Commits are signed off

🤖 Generated with [Claude Code](https://claude.com/claude-code)


02.Overview/05.Deployment/docs.md
```diff
@@ -13,6 +13,13 @@ In other words, a Deployment is the server-side model which represents a device
 The characteristics of a Deployment are:
 * Only accepted devices can be part of a Deployment
 * A device won't take part in the same Deployment twice
+    * To install a Release on a device that has already finished a Deployment for it, create a new Deployment
+    * Alternatively, resetting the device's Deployments history with the [Reset Device Deployments history
+      endpoint](https://docs.mender.io/api/#management-api-deployments-reset-device-deployments-history)
+      marks the device's finished Device Deployments as deleted, making the device eligible again for all
+      Deployments that are still active
+* A device won't install an Artifact it already has installed; the device Deployment finishes with the
+  *already-installed* status instead
```
nt-alan:
```suggestion
    * Exceptionally this can be overridden by resetting the device's Deployments history with the [Reset Device Deployments history endpoint](https://docs.mender.io/api/#management-api-deployments-reset-device-deployments-history). This marks the device's finished Device Deployments as deleted, making the device eligible again for a Deployment it was already a part of.
```

nt-alan (approved):

nt-alan (approved):

PR #2900, opened 2026-09-22
docs(orchestrator): document how Artifacts are selected

danielskinstad:
Ticket: MEN-9792


07.Orchestrate-updates/07.Artifacts handling/04.Artifact selection/docs.md
```diff
@@ -0,0 +1,56 @@
...
+
+Which Artifacts are candidates depends on how the Component type is set in the Manifest:
+
+* `artifact_path`: the Artifact at that path is the only candidate. If the file does not exist and `artifact_name` is also set, Mender Orchestrator falls back to `artifact_name`.
+* `artifact_name`: the candidates are the Artifacts in that Release on the Mender Server that are compatible with the `component_type` from the Topology. Artifacts without `artifact_provides` are ignored.
+
+Mender Orchestrator then queries the Component's Interface for its `Provides` and narrows the candidates down:
+
+1. Artifacts already installed on the Component are excluded. An Artifact counts as installed if applying its `artifact_provides` and `clears_artifact_provides` to the Component's `Provides` would change nothing. `artifact_name` is left out of this comparison unless it is the only key on either side.
+2. The remaining Artifacts are filtered by `artifact_depends`: every key the Artifact depends on must be present in the Component's `Provides` with one of the listed values. See [Compatibility checks](../03.Compatibility%20checks/docs.md) for how `device_type` is handled.
```
michalkopczan, line 25: Should we add and explanation that the whole Provides of a given component is compared, which is different than for regular mender update, where we compare only the relevant update module's Provides? E.g. if we install docker-compose artifact on a Mender device, the artifact provides/depends are compared to the docker-compose's provides/depends. When updating a component however, there's no UMs on the component, so the whole Provides is compared.

This results in something that may not be clear to customers, i.e. that we cannot deploy an artifact to a Component where the artifact modifie only a part of the Component.
I think that this may be not really clear to people, and actually this may be something that we will need to change in the future, as we know already that customers want to use orchestrator to treat UMs as separate components. So if using e.g. the docker-compose UM as a Component, the Interface would need to Provide the docker-compose's data exclusively in order for the comparison to be successful. Otherwise, the orchestrator would see the Provides as "docker-compose's provides + single-file provides + ...", and assume that given artifact is not currently installed on the Component, even if the docker-compose's Provides itself matches.

07.Orchestrate-updates/07.Artifacts handling/04.Artifact selection/docs.md
```diff
@@ -0,0 +1,56 @@
...
+3. Of the remaining Artifacts, the one with the smallest size is selected.
+
+Selection runs for every Component before any Artifact is downloaded or installed.
+
+## Example
+
+Consider a Topology with three Components of type `rtos`, updated through the `rtos-interface` Interface, and a Manifest that contains:
+
+```yaml
+component_types:
```
michalkopczan, line 35: I think it would be good to include the full topology actually, including all three components, the Interface and interface_args.

07.Orchestrate-updates/07.Artifacts handling/04.Artifact selection/docs.md
danielskinstad: > E.g. if we install docker-compose artifact on a Mender device, the artifact provides/depends are compared to the docker-compose's provides/depends.

There are no per-UM provides, it's one list on the device: depends are checked against all of it, and installing removes whatever matches clears_artifact_provides before writing the new provides.

For a component, the orchestrator counts an artifact as installed if installing it wouldn't change the component's provides. mender-update has no such check (the server decides "already installed" by artifact name). So a docker-compose artifact is still seen as installed next to single-file provides, unless its clears_artifact_provides matches those too.

I added a sentence to step 1 to make the partial case explicit, but I don't think it needs a separate explanation beyond that.

michalkopczan: EDIT: disregard for now, I think I may have misuderstood the ProvidesMatchNewProvides function. WIll edit this comment soon.

> so an Artifact that updates only part of a Component counts as installed when its own keys match

That's not true AFAIK. There's a simple check between the artifact's Provides and the Component's Provides. clears_provides is removed from Artifact's Provides, and the resulting set is compared 1:1 with the Component's Provides (see ProvidesMatchNewProvides in context/context.cpp).

So if the Artifact's Provides is only a part of the Component's Provides, even if the part from the Artifact is the same in the Component, Orchestrator will say that the artifact IS NOT currently installed on the Component.

michalkopczan (approved):

